PR #3, opened 2017-12-24
removed a couple of verification TODO comments

strattonbrazil:
I couldn't find any docs on this, but verified on a Roku.  

```
a = 1.23456789D+12 
b = 1.23456789D-12 
c = 1.23456789D12 
d = 1.23456789E+12 
e = 1.23456789E-12 
f = 1.23456789E12 

?type(a)
?type(b)
?type(c)
?type(d)
?type(e)
?type(f)
```

Types and values:

```
Double
Double
Double
Float
Float
Float

```

### DIFF
--- a/src/Lexer.ts
+++ b/src/Lexer.ts
@@ -237,7 +237,6 @@ function number() {
         advance();
 
         // exponents are optionally signed
-        // TODO: Confirm this in BrightScript documentation!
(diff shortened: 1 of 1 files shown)


sjbarag:
This is great! Thanks *so much* for confirming on a device, Josh - you’re the best!

PR #6, opened 2017-12-25
Started defining a formal grammar for brightscript

Slushy:
**Note: I'm primarily sending out this PR as a talking point to see if you all would be interested in having and iterating on a grammar file. I personally see this as a larger benefit in the medium-long run when we start trying to parse more complicated syntax trees, but I think it could also benefit us these earlier stages as well.** @sjbarag @strattonbrazil 

With that said, I began to lay out some basic brightscript grammar which should be a good base for further iteration and could also possibly help guide the parser's development direction. But depending on how you guys feel, if we are OK with this approach then this should be a good enough base to merge in. But regardless i'm definitely open to hearing what you guys think and answer any questions. For what I _do_ have here I tried to make sure everything I defined was according to brightscript docs, but please let me know of any questions or if you spot anything out of place.

**I've added grammar to support the following:**
- a full program
- _basic_ assignment and conditional expressions following operational precedence (does not include post-fix operations)
  - x = 10
  - y *= 50.5
  - x <= 2 ^ (20 mod 4)
  - z = x or y : x = y and z

-  a block if/else if/ else statement
```
  if x < 20 then
    y = 20
  else if x < 30 then
    y = 30
  else
    y = 0
  end if
```
- method _declaration_
```
  function add (a as integer, b as integer) as integer
  ... statements ...
  end function

  sub print(val)
  ... statements ...
  end sub 
```

### DIFF
--- a/src/grammar/BrightscriptGrammer.ebnf
+++ b/src/grammar/BrightscriptGrammer.ebnf
@@ -1,0 +1,196 @@
+(*
+    This is an attempt at defining a formal grammar of the brightscript language that we can iterate on and
+    define as we continue to implement the parser. This is written in EBNF syntax (but using regex
+    to instead represent required/optional symbols or groupings).
+    
+    Note: Any words that are fully capitalized represent a lexeme (see full enum in lexeme.ts) whose text
+    representation may vary or where deems appropriate (e.g. the 'Integer' lexeme can represent any valid integer, 
+    so we reference the 'Integer' lexeme here as 'INTEGER', but the 'True' lexeme only represents the value 'true' 
+    so we instead just reference the text value here as 'true' to be explicit). 
+
+    QRC (Quick Regex Cheatsheet)
+        *  - zero or more
+        +  - one or more
+        ?  - zero or one
+*)
+
+
+(* Start from the top, a brightscript script will have a number of "statements" before reaching the end of the file *)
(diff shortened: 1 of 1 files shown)


sjbarag:
This looks great @Slushy !  *very* nice work here.  You and @strattonbrazil bring up a good point, in that this enables us to use a parser generator library if we wanted.  I've been hesitant to use one because I traditionally have a lot of trouble getting them to work (and they tend to have opaque error messages), but I'm not opposed to giving one a try.  I'm also tempted to just hand-write a parser using this as a base, because the grammar seems unlikely to change more than once per year. I'd assume those changes will be backwards-compatible, or else Roku channel devs would have to do some nasty conditional compilation to support multiple firmware versions.

At the very least, having an EBNF file will make hand-writing a parser easier.  Who knows, maybe I'll just use [moustache](https://github.com/janl/mustache.js/) to generate a bunch of functions from this class!

Regardless, thanks again for all the effort you put into this. It made a wonderful little Christmas gift ❤️ 🎄

sjbarag:
For clarity & posterity: I'm merging this into `master`.  Its location may change in the future as we figure out the correct app architecture 😄

sjbarag:
Noooooo I forgot to click "Approve", I'm sorry!

PR #7, opened 2017-12-31
(test) Switch from mocha+chai to jest for testing

sjbarag:
Facebook's [jest](http://facebook.github.io/jest/) unit testing framework has really nice snapshotting support and is pretty popular within the `node` community (> 3,000,000 downloads in the past month, though I'm not sure how many of those are from Facebook 🤷‍♂️).  Snapshotting will allow us to not hand-write expected output for our parser while still allowing us to be notified when things break.  Mocha and Chai have some plugins that add snapshotting, but being able to point to a single website and say "we're  using this for tests" is rather nice.

The downside is that unit tests now take 2-3 times as long to run, because Jest attempts to target projects with lots and lots of tests by parallelizing test execution where possible.  This winds up being slower than Mocha's naive approach of just running the tests in-order, due to the small number of tests.  I imagine this slow-down will eventually diminish to a negligible difference.


sjbarag:
Possibly the biggest downside to this is a 3x increase in build times on Travis, from 2 minutes to 6ish minutes:

travis-build-times

Slushy:
> Possibly the biggest downside to this is a 3x increase in build times on Travis, from 2 minutes to 6ish minutes

This is good to know. I don't see this being that big of a deal unless this repo experiences some sort of fast moving pr flood.

sjbarag:
> > Possibly the biggest downside to this is a 3x increase in build times on Travis, from 2 minutes to 6ish minutes
>
> This is good to know. I don't see this being that big of a deal unless this repo experiences some sort of fast moving pr flood.

Looks like travis speeds up some builds too, so this isn't actually as big of a problem as I thought!

PR #8, opened 2018-01-01
(parse) Add a parser phase

sjbarag:
An interpreter isn't useful until it executes an AST, and it's the parser's job to build that AST.  This PR adds a parser that supports:

1. Literal values (including `true`, `false`, and `invalid`)
2. Unary operators (`not` and negation)
3. The exponential operator (`^`)
4. Addition, subtraction, multiplication, division, modulo, and integer division
5. Comparisons

The `orbs` binary now prints an AST for every line entered, e.g.:

```
$ orbs
ORBS> (2 + 3) < 9^-1
(<
  (group
    (+
      2
      3
    )
  )
  (^
    9
    (-
      1
    )
  )
)
ORBS> 
```

There's a bunch of tests for this phase as well, all utilizing Jest's [snapshot testing](http://facebook.github.io/jest/docs/en/snapshot-testing.html) paradigm to save me from having to build  a bunch of ASTs for comparison by hand.  I almost certainly missed some things, so feel free to open an issue or a PR if you find something 😄.  Or just complain here!

Happy new year!


sjbarag:
> *sjbarag* wants to merge 15 commits into `master` from `start-parser`
> `start-parser`

i'm really good at naming i swear

src/parser/index.ts
```diff
@@ -0,0 +1,178 @@
+import * as Expr from "./Expression";
+type Expression = Expr.Expression;
+import { Lexeme } from "../Lexeme";
+import { Token } from "../Token"
+import * as ParseError from "./ParseError";
+
+let current: number;
+let tokens: ReadonlyArray<Token>;
+
+export function parse(toParse: ReadonlyArray<Token>) {
```
Slushy: I definitely see this class just becoming a large function chain, but the fact that you have actually defined the functions in the order they are actually called makes it significantly easier to understand and follow. I'm sort of just thinking ahead, but will we ever want to separate out related parsing logic (e.g. expressions, for/while/for each loops, if/else, functions)? Or is the plan for now to sort of keep this class the home for all parsing logic? It is a bit hard to gauge on how hairy this class could become.

sjbarag: That's a great point!  I've got no idea how rough this will become, but I suspect it'll just grow as a single file for quite a long time.  For comparison, [TypeScript's parser.ts](https://github.com/Microsoft/TypeScript/blob/0477f91dbad209809363323042f34926d3a6e175/src/compiler/parser.ts#L7498) weighs in at 7,498 lines `0_o`

PR #14, opened 2018-01-05
(exec) Add execution stage

sjbarag:
A programming language isn't very useful if it just produces an abstract syntax tree.  Let's evaluate that tree!  BRS is now a functioning (but _very_ simple) interpreter that supports:

1. Arithmetic:
    * Addition
    * Subtraction
    * Multiplication
    * Division
    * Integer division
    * Modulo operations
    * Exponentiation
    * Parenthetic grouping
2. Boolean logic:
    * And/or/not of booleans
3. Comparisons
    * Less-than, less-than-or-equal, greater-than, greater-than-or-equal, equal, and non-equal

The REPL works pretty well, but because there's no `print` statement yet executing a file doesn't actually produce anything.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -43,7 +43,7 @@ function run(contents: string) {
         return;
     }
 
-    const printer = new Executioner();
+    const executioner = new Executioner();
 
-    console.log(printer.exec(expr!));
+    executioner.exec(expr!);
 }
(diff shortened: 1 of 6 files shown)


sjbarag:
This still needs a ton more tests.  And we're going to be able to do end-to-end testing very soon!

PR #25, opened 2018-02-16
chore(naming): Rename "Executioner" -> "Interpreter"

sjbarag:
While the "Executioner" name was fun at first, it became a pain to say. More importantly, it's a super non-standard name, which makes it harder for a newcomer to understand how this whole project is structured.

This project now has no Judge Judy or Executioner.


sjbarag:
Note to self: rename the `executioner` GitHub label when this PR gets merged.

PR #30, opened 2018-02-25
feat: Add while loop support

sjbarag:
While loops are necessary for basically every language, because they allow the execution duration of a program to not be bound to the source length of the program.

Also BrightScript on Rou makes heavy use of `while` loops, so we need them too.

### DIFF
--- a/src/Lexeme.ts
+++ b/src/Lexeme.ts
@@ -58,6 +58,9 @@ export enum Lexeme {
     EndWhile,
     Eval,
     Exit,
+    // ExitFor isn't listed as a reserved word, but it seems like it'd match 'exitwhile'.
+    // TODO: Verify its reservedness on-device.
+    ExitFor,
     ExitWhile,
     False,
     For,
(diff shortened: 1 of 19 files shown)


src/Lexeme.ts
```diff
@@ -58,6 +58,9 @@ export enum Lexeme {
     EndWhile,
     Eval,
     Exit,
+    // ExitFor isn't listed as a reserved word, but it seems like it'd match 'exitwhile'.
+    // TODO: Verify its reservedness on-device.
```
sjbarag: Experimentally, `exitfor` isn't a reserved word, and `exit for` is the only way to exit a `for` loop:

```brightscript
for i=0 to 10
    print i
    if i = 2 then exit for
end for

exitfor = "not reserved"
print exitfor             ' prints "not reserved"
```

But `exitwhile` is indeed a reserved word, and is substitutable for `exit while`:

```brightscript
a = 0
while a < 10
    a = a + 1
    print a
    if a = 3 then exitwhile
end while

exitwhile = "not reserved?"    ' ERRNO: 175 - Exit While is not inside a While.
print exitwhile
```

PR #35, opened 2018-04-04
feat(parse,interp,e2e): Add basic global native functions

sjbarag:
This gets the groundwork set up for the rapid implementation of global functions &mdash; including the ability to call them at runtime.  There's a few caveats there for now though:

1. Only `LCase`, `UCase`, (string lowercase and string uppercase) and `RebootSystem` (which **does nothing** in this implementation [1]) are implemented so far.
2. I haven't figured out how I want to implement function argument type checking, e.g. causing an error to occur when `LCase(185)` is called.
3. Adding new globally-scoped functions is probably a bit tougher than it needs to be right now.  That's pretty minor though.

[1] I guess I could shell out to `shutdown -r` but that seems particularly mean and useless.


sjbarag:
Oh right, I forgot to mention:

The same `Callable` system will work for user-defined functions as well!  We still need to parse and interpret them, but the overall pattern will still apply :smile:

PR #36, opened 2018-04-05
 implemented Asc() and Chr() global string functions

strattonbrazil:
There are probably several edge cases I'm missing to match Roku's implementation.


strattonbrazil:
I tested these functions on the Roku.  I noticed Chr() returns an empty string for negative numbers, so I updated the implementation.  I also added some UTF-16 symbols based on my Roku tests.  

Roku test code:
```
    ?asc("a")
    ?asc("\n")
    ?asc("ぇ")
    ?"-----"
    ?chr(12359)
    ?chr(-1)
    ?chr(0)
    ?chr(17)
```

Output:
```
 97
 92
 12359

-----
ぇ


?

```

src/stdlib/String.ts
```diff
@@ -24,5 +24,40 @@ class LCaseCallable extends Callable {
...
 }
 
+class AscCallable extends Callable {
+    arity = {
+        required: 1,
+        optional: 0
+    }
+
+    call(interpreter: Interpreter, str: BrsString): BrsType {
+        if (str.value.length == 0)
```
sjbarag: We might be able to take advantage of some `node`-isms to reduce this a bunch!  Consider something like:

```typescript
return new Int32(
  str.value.charCodeAt(0) || 0
);
```

As it turns out, `"".charCodeAt(0)` returns `NaN`, which lets us fall back to `0` as the base case :smiley:

But it's totally possible that I'm misunderstanding Roku's implementation here!

src/stdlib/String.ts
```diff
@@ -24,5 +24,40 @@ class LCaseCallable extends Callable {
...
+}
+
+class ChrCallable extends Callable {
+    arity = {
+        required: 1,
+        optional: 0
+    }
+
+    call(interpreter: Interpreter, value: BrsType): BrsType {
+        if (value instanceof Int32) {
```
sjbarag: Any particular reason to use an `instanceof` check vs. comparing to `value.kind === ValueKind.Int32`?  The TS compiler should give you the same type support, but at runtime we only have to do a numeric comparison.

It's not a blocker by any means &ndash; I'm just curious (and secretly hoping I haven't somehow broken those `value.kind` comparisons :sweat_smile:)

src/stdlib/String.ts
```diff
@@ -24,5 +24,40 @@ class LCaseCallable extends Callable {
     }
 }
 
+class AscCallable extends Callable {
+    arity = {
+        required: 1,
+        optional: 0
+    }
+
+    call(interpreter: Interpreter, str: BrsString): BrsType {
```
sjbarag: Just as an FYI if you go to bash on these stdlib functions more, calling `Asc(invalid)` or `Asc(false)` probably crashes the intepreter due to the lack of runtime type checking.  It's totally my fault!  Function argument type checking is a work-in-progress, so don't panic :vulcan_salute:

src/stdlib/String.ts
strattonbrazil: Will switch.

src/stdlib/String.ts
strattonbrazil: Unfamiliarity with the type system.  Will change.

PR #42, opened 2018-04-19
feat(interp,stdlib): Add call-time type checking

### LINKED ISSUE #39: Add runtime type checking
It's currently possible to crash the entire interpreter by calling a function with the wrong argument types, e.g.:

* attempting to lowercase a non-string: `LCase(false)`
* attempting to uppercase a non-string: `UCase(invalid)`

Rather than crashing with an esoteric JavaScript stack trace, we should provide a helpful error message that tells users which arguments are incorrectly typed and in which function.

sjbarag:
My use of BrightScript on Roku devices has a rich history of throwing vague `Type mismatch` errors and... well that's about it.  This interpreter needs centralized type checking for function calls, to ensure `brs` don't ungracefully crash when attempting to call `UCase(5)`.  The goal here is to provide more helpful error messages than the reference interpreter though, e.g.:

```
brs> UCase(5)
[line 1] Type mismatch in 'UCase': argument 's' must be of type String, but received Integer.
[Line 1] Type mismatch(es) detected.
brs> 
```

Perhaps I should add in argument indexes too?  Something like "Type mismatch in 'FunctionName': argument n, 'name', must be of type Foo, but received Bar." could be nice.

That can come in a future PR though I imagine :smile: 

fixes #39

### DIFF
--- a/src/Error.ts
+++ b/src/Error.ts
@@ -6,7 +6,7 @@ export function make(message: string, line: number, file?: string) {
 }
 
 export function runtime(message: string, line: number, file?: string) {
-    let location = file ? `${file}: ${line}` : `line ${line}`;
+    let location = file ? `${file}: ${line}` : `Line ${line}`;
     let error = new Error(`[${location}] ${message}`);
     if (process.env.NODE_ENV !== "test") {
         console.error(error.message);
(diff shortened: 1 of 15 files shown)


sjbarag:
@strattonbrazil and @Slushy &ndash; if either of you get some time I'd love to get your opinions on this!

I think I forgot end-to-end tests so let me go do that real fast :sweat_smile:

sjbarag:
@strattonbrazil and @Slushy -- last call for reviews!  I'm going to merge this tonight if I don't hear from you :grinning:

PR #54, opened 2018-05-23
feat(lex,parse,interp,e2e): Implement print lists

sjbarag:
Roku's BrightScript [supports printing multiple things with one `print` statement](https://sdkdocs.roku.com/display/sdkdoc/Program+Statements#ProgramStatements-PRINTitemlist), so we should too!  This was surprisingly more difficult than I expected, but a pretty big part of that was making it possible to spy on output without messing with `process.stdout.write`.

Merging this PR will result in the following file being valid according to `brs`:

```brightscript
' All examples inspired by Roku documentation:
' https://sdkdocs.roku.com/display/sdkdoc/Program+Statements#ProgramStatements-PRINTitemlist

' no separators
print "lorem " 1 "psum"

' arbitrary mixes of separators
print 9; " is equal to"; 3^2
print "column a", "column b", "column c", "column d"

' arbitrary tabbing
print tab(3) "I started at col 3"; tab(25) "I started at col 25"

' print position
print "0123" pos(0)

' calculate tabs based on position, though you'd probably want to just define "    " as a variable
print "lorem" tab(pos(0) + 4) "ipsum" tab(pos(0) + 4) "dolor" tab(pos(0) + 4) "sit" tab(pos(0) + 4) "amet"

' suppress trailing newline
print "no newline";
```

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -5,13 +5,17 @@ import { Token } from "./Token";
 import * as Lexer from "./lexer";
 import * as Parser from "./parser";
 import { AstPrinter } from "./interpreter/AstPrinter";
-import { Interpreter } from "./interpreter";
+import { Interpreter, OutputStreams } from "./interpreter";
 import { stringify } from "./Stringify";
 import * as BrsError from "./Error";
 
-const interpreter = new Interpreter();
+/** The `stdout`/`stderr` pair from the process that invoked `brs`. */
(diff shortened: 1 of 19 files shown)


sjbarag:
This still needs a good bit of documentation in a lot of places; I'll try to cover that in the next few days!

PR #64, opened 2018-10-02
Added some String Global Functions

peteley:
I did not comment on the issue and I did not finish all of the String functions, but I wanted to help out this project.

I have added Right, Mid, Len, and Instr. I have added tests around some basic functionality with those functions. I hope this helps and I welcome all feedback.

I also added `.idea/` to `.gitignore` since I use WebStorm for development.

### DIFF
--- a/src/stdlib/String.ts
+++ b/src/stdlib/String.ts
@@ -68,6 +68,25 @@ export const Left = new Callable(
 );
 
 /**
+ * Returns the last n characters in a string
+ */
+export const Right = new Callable(
+    {
+      name: 'Right',
+      args: [{name: 's', type: ValueKind.String}, {name: 'n', type: ValueKind.Int32}],
+      returns: ValueKind.String
+    },
+    (interpreter: Interpreter, s: BrsString, n: Int32) => {
+        let end = s.value.length - 1;
+        let start = end - (n.getValue() - 1);
+
+        if (n.getValue() <= 0) return new BrsString("");
+        else if (start < 0) return new BrsString(s.value);
+        return new BrsString(s.value.substr(start, end));
+    }
+);
+
+/**
  * Returns the index (1 based) of a string inside another string. Zero if there is no match.
  */
 export const Instr = new Callable(
--- a/src/stdlib/String.ts
+++ b/src/stdlib/String.ts
@@ -1,12 +1,12 @@
-import { BrsType, Callable, ValueKind, BrsString, Int32 } from "../brsTypes";
-import { Interpreter } from "../interpreter";
+import { BrsType, Callable, ValueKind, BrsString, Int32 } from '../brsTypes';
+import { Interpreter } from '../interpreter';
 
 /** Converts the string to all uppercase. */
 export const UCase = new Callable(
     {
-        name: "UCase",
-        args: [{ name: "s", type: ValueKind.String }],
-        returns: ValueKind.String,
+      name: 'UCase',
+      args: [{name: 's', type: ValueKind.String}],
+      returns: ValueKind.String
     },
     (interpreter: Interpreter, s: BrsString) => new BrsString(s.value.toUpperCase())
 );
@@ -14,13 +14,12 @@ export const UCase = new Callable(
 /** Converts the string to all lowercase. */
 export const LCase = new Callable(
     {
-        name: "LCase",
-        args: [{ name: "s", type: ValueKind.String }],
-        returns: ValueKind.String,
+      name: 'LCase',
+      args: [{name: 's', type: ValueKind.String}],
+      returns: ValueKind.String
     },
     (interpreter: Interpreter, s: BrsString) => new BrsString(s.value.toLowerCase())
 );
-
 
 /**
  * Returns the Unicode ("ASCII") value for the first character of the specified string.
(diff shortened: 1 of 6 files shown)


src/stdlib/String.ts
```diff
@@ -1,36 +1,35 @@
...
     (interpreter: Interpreter, s: BrsString) => new BrsString(s.value.toUpperCase())
 );
 
 /** Converts the string to all lowercase. */
 export const LCase = new Callable(
     {
-        name: "LCase",
-        args: [{ name: "s", type: ValueKind.String }],
-        returns: ValueKind.String,
+      name: 'LCase',
```
sjbarag: Looks like the IDEA products default to two-space indents and single quotes, but the convention in these files has typically been four-space indents and double quotes.  That's *entirely* on me, because I never set up a `.editorconfig` or anything like it!

Would it be too much trouble to reformat after I set one up tonight?  I'd normally merge this as-is and reformat later, but I'd much rather you show up on these lines in `git blame` than me! :smile:

src/stdlib/String.ts
```diff
@@ -61,9 +60,67 @@ export const Chr = new Callable(
...
+ */
+export const Mid = new Callable(
+    {
+      name: 'Mid',
+      args: [{name: 's', type: ValueKind.String}, {name: 'p', type: ValueKind.Int32}, {name: 'n', type: ValueKind.Int32}],
+      returns: ValueKind.String
+    },
+    (interpreter: Interpreter, s: BrsString, p: Int32, n: Int32): BrsString => {
+      let start = p.getValue() - 1;
+      return (n) ? new BrsString(s.value.substring(start, start + n.getValue())) : new BrsString(s.value.substring(start));
```
sjbarag: :100: :100: :100: 
Nice and clean!

src/stdlib/String.ts
```diff
@@ -61,9 +60,67 @@ export const Chr = new Callable(
...
+    {
+      name: 'Right',
+      args: [{name: 's', type: ValueKind.String}, {name: 'n', type: ValueKind.Int32}],
+      returns: ValueKind.String
+    },
+    (interpreter: Interpreter, s: BrsString, n: Int32) => {
+        let end = s.value.length - 1;
+        let start = end - (n.getValue() - 1);
+
+        if (n.getValue() <= 0) return new BrsString("");
```
sjbarag: Looks like the [BrightScript docs](https://sdkdocs.roku.com/display/sdkdoc/Global+String+Functions#GlobalStringFunctions-Right(sasString,nasInteger)asString) don't specify the behavior here &ndash; is that how it works in the reference implementation on a Roku?  I can't remember at the moment :sweat_smile:

test/stdlib/String.test.js
```diff
@@ -84,4 +84,76 @@ describe("global string functions", () => {
...
+            ).toEqual(new BrsString("pple"));
+        });
+
+        it("get the original string back with n larger than string length", () => {
+            expect(
+                Right.call(interpreter, new BrsString("boy"), new Int32(5))
+            ).toEqual(new BrsString("boy"));
+        });
+
+        it("get back empty string when character length is 0 or negative", () => {
```
sjbarag: Wow, thanks for being so thorough in these tests!  These look phenomenal.

test/stdlib/String.test.js
```diff
@@ -84,4 +84,76 @@ describe("global string functions", () => {
...
+            ).toEqual(new Int32(0));
+        });
+
+        it("returns the index of the first found string", () => {
+            expect(
+                Instr.call(interpreter, new Int32(1), new BrsString("apple"), new BrsString("a"))
+            ).toEqual(new Int32(1));
+        });
+
+        it("returns the index of the first found after the starting index passed in", () => {
```
sjbarag: I *love* that the `it` description reads like a full sentence describing the behavior here.  Consider using that for the other test cases you added here &mdash; these are super easy to read!

sjbarag:
> I did not comment on the issue and I did not finish all of the String functions, but I wanted to help out this project.

Don't worry about it!  I wrote that to keep multiple people from building the same thing, but that's definitely not happened here.  Thanks for contributing! :heart:

sjbarag:
Just as an FYI, I landed #65, which adds `tslint` support to this build.  You'll probably want to rebase on top of `master` to ensure your changes still build.  I'm sorry for the churn here!  This won't happen again :cry:

src/stdlib/String.ts
peteley: Sorry about that, it is defaulted at 2. I do not mind converting back. (I will also change the default for working on this project)

src/stdlib/String.ts
peteley: I will check on my Roku, but I believe this is correct. I will add this to the e2e test as well and run in both places (brs and my Roku).

test/stdlib/String.test.js
peteley: I agree about making tests easy to read. I will work on the others. (excuse: It was late and my eyes were a bit blurry 😄)

peteley:
Thanks for the nice words about my effort. I will work on this tonight and tomorrow to get it cleaned up.

I would like to thank you for making this project a thing. It is well written and very easy to understand and jump into to start contributing.

Cheers,
Pete

Issue #50, opened 2018-04-26
Implement global String conversion functions


sjbarag:
See [the BrightScript documentation](https://sdkdocs.roku.com/display/sdkdoc/Global+String+Functions) for details on what each one does, but this should cover:

* `Str(value as Float) as String`
* `StrI(value as Integer) as String`
* `StrI(value as Integer, radix as Integer) as String`
* `Val(s as String) as Float`
* `Val(str as String, radix as Integer) as Integer`

PR #68, opened 2018-11-13
Str stri val implementation

strattonbrazil:
This contains a fix to better represent floats in a double-only world.  It also implements the global string functions Str(), StrI(), and Val().


sjbarag:
Handles most of #50 , but I've still got some work to do to get optional parameters working for functions in the `stdlib` tree.

src/stdlib/String.ts
```diff
@@ -122,3 +122,50 @@ export const Mid = new Callable(
...
+ * Return a float from a string
+ */
+export const Val = new Callable(
+    {
+        name: "Val",
+        args: [{name: "s", type: ValueKind.String}],
+        returns: ValueKind.Float
+    },
+    (interpreter: Interpreter, s: BrsString): Float => {
+        return new Float(+s.value);
```
sjbarag: This is fine, but consider using [Float.fromString](https://github.com/sjbarag/brs/blob/e5f4e01420a2e8145227152f314f06a8025026d8/src/brsTypes/Float.ts#L25-L34) or `new Float(Number(s.value))`  for future use-cases like this.  It's a bit more explicit than the `+` operator.

... though now that I think about it, should `Float.fromString` be changed to remove the `Number.parseFloat` call?  Seems unnecessary in light of the changes to the `Float` constructor in this PR. :thinking:

strattonbrazil: I tried removing the parseFloat call in in Float, but that broke other tests, so I held off on that.  I did remove the '+' operator to be more explicit.

strattonbrazil:
Added e2e test, which I didn't realize existed.  Thanks for pointing them out.

src/stdlib/String.ts
sjbarag: Thanks for letting me know!

src/interpreter/index.ts
```diff
@@ -77,7 +77,10 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
             { name: "Right",        func: StdLib.Right },
             { name: "Instr",        func: StdLib.Instr },
             { name: "Len",          func: StdLib.Len },
-            { name: "Mid",          func: StdLib.Mid }
+            { name: "Mid",          func: StdLib.Mid },
+            { name: "Str",          func: StdLib.Str },
+            { name: "StrI",         func: StdLib.StrI },
+            { name: "Val",          func: StdLib.Val }
```
sjbarag: One of these days I'll figure out a pattern for this that I don't hate.  Until then, this is :dollar:!

PR #69, opened 2018-11-14
fix(stdlib): Allow optional parameters and signature overloads in stdlib functions

sjbarag:
Signature overrides weren't previously supported at all; now they are!  Optional parameters weren't previously supported in stdlib functions (e.g. `StrI`'s `radix` parameter).  Those are now supported as well!

This includes a demo of that in the `StrI` function.


sjbarag:
Hmm, considering that `Rnd` can accept `0` and return a `float` or can accept an `integer` and return an `integer`, this probably isn't _quite_ the right solution here.  Might be better to support *multiple signatures* instead of a single signature with multiple argument types.

But if there's going to be multiple signatures, perhaps there's multiple implementations instead of trying to cram them together.  Does that mean this interpreter would just support multiple / dynamic dispatch of functions?

strattonbrazil:
Is this a signature problem with Rnd() though?  Many functions will return
a dynamic value like an object or invalid.  Returning an integer or float
doesn't seem that big of a deal.  In the case of Sgn, can't it just accept
floats, and ints passed in are casted?  That one seems easy as the return
type is always an integer, no?

-Josh

src/brsTypes/Callable.ts
```diff
@@ -92,7 +145,68 @@ export class Callable implements Brs.BrsValue {
...
+            return "[anonymous function]";
+
+        }
+    }
+
+    getName(): string {
+        return this.name || "";
+    }
+
+    getFirstSatisfiedSignature(args: Brs.BrsType[]): SignatureAndImplementation | undefined {
```
strattonbrazil: Do we expect many ambiguous cases here?

PR #70, opened 2018-11-15
implement math functions Exp, Log, Sqrt, and Rnd (issue 47)

strattonbrazil:
This implements the Exp, Log, Sqrt, and Rnd global math functions.  The first three are fairly straight forward mapping to javascript, but Rnd is a little weird.  

**Rnd Range**
Javascript's Math.random() returns a value in the range [0,1) like most languages, while the Roku docs say the Rnd(0) implementation returns the following range (0,1).  I don't know if that's true, but in practice this shouldn't be an issue as nothing should rely on 0.0 ever being explicitly returned.  

**Rnd Testing**
The first random test just verifies that the random float range is between zero and one.  This doesn't do anything to verify the distribution of random.  

The second random test is actually mocking the Math.random() function to make sure to possible values coming out correctly pick values in the range provided including the first and last possible values.


test/stdlib/Math.test.js
```diff
@@ -0,0 +1,66 @@
...
+    describe("Rnd", () => {
+        it("calculates a random float between zero and one", () => {
+            let result = Rnd.call(interpreter, new Int32(0));
+            expect(result.kind).toBe(ValueKind.Float);
+            expect(result.value).toBeGreaterThanOrEqual(0.0);
+            expect(result.value).toBeLessThan(1.0);
+        });
+
+        it("calculates a random number between one and some number", () => {
+            function mockRandomToAlwaysBe(val) {
```
sjbarag: Watch out - looks like anything that uses `Math.random` after this function is called will still use the last-mocked value! I've got a commit that fixes that though, so I'll push that in a sec.

strattonbrazil: I figured I wouldn't use random anywhere else in this file and that jest supposedly resets global state per test file, but that's cleaner anyway.

test/e2e/StdLib.test.js
```diff
@@ -39,4 +39,16 @@ describe("end to end standard libary", () => {
             ]);
         });
     });
+
+    test("stdlib/math.brs", () => {
+        return execute(resourceFile("stdlib", "math.brs"), outputStreams).then(() => {
```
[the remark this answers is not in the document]
strattonbrazil: Doh!  That's what I get for adding e2e last minute.

PR #72, opened 2018-11-17
implements Abs, Cdbl, Cint, Csng, Fix, Int, Sgn (issue #46)

### LINKED ISSUE #46: Implement rounding/truncating/converting global math functions
Another subset of #43!  Covers:

* `Abs(x as Float) as Float`
* `Cdbl(x as Integer) as Float`
* `Cint(x as Float) as Integer`
* `Csng(x as Integer) as Float`
* `Fix(x as Float) as Integer`
* `Int(x as Float) as Integer`
* `Sgn(x as Float) as Integer`
* `Sgn(x as Integer) as Integer`

strattonbrazil:
This implements a bunch of global math functions.  When available, I tried to use test examples straight from the brightscript docs.  

A couple of things to note:
* the e2e tests are currently breaking because Sgn() needs to support int and float params -- should I use a dynamic type instead and cast it in the function?  
* in the implementation of Int() I call Math.floor() on the float to behave like integer casting.  I need to do this because the constructor for Int32 calls Math.round() on the input, but I'm not sure why.

Closes #46


sjbarag:
There's a nifty debug mode for tests that I haven't documented yet.  End-to-end tests will normally fail with "Error occurred." which isn't very helpful.  That's because `stderr` gets captured during the end-to-end tests to prevent intentional type-checking errors from spamming the test output stream.  It's a bit of a bummer because that means legitimate failures don't appear.

Running tests with the `NODE_ENV` environment variable set to anything other than `test` will display the full output.  Used for this PR, here's the output:

```
$ NODE_ENV=debug yarn test test/e2e/StdLib.test.js 
yarn run v1.10.1
$ jest test/e2e/StdLib.test.js
 FAIL  test/e2e/StdLib.test.js
  end to end standard libary
    ✓ stdlib/strings.brs (18ms)
    ✕ stdlib/math.brs (11ms)

  ● end to end standard libary › stdlib/math.brs

    Error
      Failed: Error occurred

  console.error lib/Error.js:14
    [Line 17] Type mismatch in 'Sgn': argument 'x' must be of type Float, but received Integer.

Test Suites: 1 failed, 1 total
Tests:       1 failed, 1 passed, 2 total
Snapshots:   0 total
Time:        0.625s, estimated 1s
Ran all test suites matching /test\/e2e\/StdLib.test.js/i.
error Command failed with exit code 1.
info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
```

sjbarag:
> the e2e tests are currently breaking because Sgn() needs to support int and float params -- should I use a dynamic type instead and cast it in the function?

Supporting both `Float` and `Integer` inputs will be supported once #73 lands in master, which should be sometime today or tomorrow!  You'll need to rebase on `master` and change the construction of these `Callable`s but it's pretty straightforward.  If I've got time, I can do the rebase and it push it to this branch.

> in the implementation of Int() I call Math.floor() on the float to behave like integer casting. I need to do this because the constructor for Int32 calls Math.round() on the input, but I'm not sure why.

I think that was an initial attempt to get Float -> Int conversions working, but it seems like that's supposed to be `Math.floor` anyway.  Feel free to change it if it doesn't make sense!

strattonbrazil:
That's great to know, Sean.  I did notice that error message not being very helpful.  Don't worry about rebasing this branch.  I'll take care of that once your change lands in master.  

I thought about changing the Math.floor call, but it broke a few tests I wasn't in the mood to fix.  Maybe we should file that as a future investigation.

src/stdlib/Math.ts
```diff
@@ -1,6 +1,86 @@
...
+      args: [{name: "x", type: ValueKind.Float}],
+      returns: ValueKind.Float
+    },
+    (interpreter: Interpreter, x: Float) => new Float(Math.abs(x.getValue()))
+);
+
+/*
+ * Returns the integer as a 32-bit float. 
+ * ** NOTE: the function name implies it makes a 64-bit float, but the docs say
+ *     it currently returns a 32-bit float, but may return a 64-bit float in the future.
```
sjbarag: > but may return a 64-bit float in the future

:man_facepalming:

src/stdlib/Math.ts
```diff
@@ -1,6 +1,112 @@
...
+            else if (val < 0.0) return new Int32(-1);
+            else return new Int32(0);
+        }
+    },
+    {
+        signature: {
+            args: [{name: "x", type: ValueKind.Int32}],
+            returns: ValueKind.Int32
+        },
+        impl: (interpreter: Interpreter, x: Int32) => {
```
sjbarag: Given that these are the same implementation, do you think it'd be worth pulling that out to a separate variable and using it in both signatures?  Something like this:

```typescript
function SgnImpl(interpreter: Interpreter, x: Int32 | Float) {
    let val = x.getValue();
    if (val > 0.0) return new Int32(1);
    else if (val < 0.0) return new Int32(-1);
    else return new Int32(0);
}

export const Sgn = new Callable(
    "Sgn",
    {
        signature: {
            // Accepts Float
        },
        impl: SgnImpl
    },
    {
        signature: {
            // Accepts Int32
        },
        impl: SgnImpl
    }
);
```

strattonbrazil: I actually did that locally and thought it wasn't significant enough to justify, but if you'd prefer I can add something like this.

PR #73, opened 2018-11-19
feat(stdlib,interp): Allow multiple overloads for stdlib functions

sjbarag:
Some standard library functions require full-fledged signature overloads!  Implementing that in `brs` basically requires [multiple dispatch](https://en.wikipedia.org/wiki/Multiple_dispatch) to be implemented.  But now multiple signatures for a single `stdlib` function is possible!  The correct signature and its associated implementation will be dynamically chosen at runtime for each function invocation.

A neat side-effect is that type-mismatch error messages are much nicer now!


sjbarag:
New error message format:

```
[Line 1] Provided arguments don't match any of Mid's signatures.
    function Mid(s as String, p as Integer) as String:
        * Mid requires at least 2 arguments, but received 1.
        * Argument 's' must be of type String, but received Integer.
    function Mid(s as String, p as Integer, n as Integer) as String:
        * Mid requires at least 3 arguments, but received 1.
        * Argument 's' must be of type String, but received Integer.
Error occurred
```

sjbarag:
Just gonna merge this &mdash; there was a ton of change in the `stdlib` tree to match the new `Callable` constructor, and a lot of logic moved around from `BrsFunction.ts` and `interpreter/index.ts` into `Callable.ts`, but that's a good thing I guess.

Otherwise there were basically no test changes other than a few checks that looked for specific error messages.  It doesn't seem like we've broken anything!

PR #74, opened 2018-11-22
finishing implementation of Val int conversion

### LINKED ISSUE #50: Implement global String conversion functions

strattonbrazil:
Most of issue #50 has been resolved, but the Val implementation didn't support int conversion.  This is much easier with the default parameter support.  The radix support was extremely simple, because it's handled in much the same way as javascript's parseInt().  For tests I included the examples from the Roku docs as well as the base-10 case.  

NOTE: I have a really cheesy way to detect if a string is a float just be checking for a decimal place.  Open to other ideas.  

Closes #50.

### DIFF
--- a/src/stdlib/String.ts
+++ b/src/stdlib/String.ts
@@ -1,6 +1,7 @@
 import { BrsType, Callable, ValueKind, BrsString, Int32, Float } from "../brsTypes";
 import * as Expr from "../parser/Expression";
 import { Interpreter } from "../interpreter";
+import { BrsNumber } from "../brsTypes/BrsNumber";
 
 /** Converts the string to all uppercase. */
 export const UCase = new Callable(
@@ -205,7 +206,7 @@ export const StrI = new Callable(
                     type: ValueKind.Int32
                 },
                 {
-                    name: "radis",
+                    name: "radix",
                     type: ValueKind.Int32,
                     defaultValue: new Expr.Literal(new Int32(10))
                 }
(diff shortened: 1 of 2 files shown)


src/stdlib/String.ts
```diff
@@ -233,11 +234,29 @@ export const Val = new Callable(
...
+                    type: ValueKind.Int32,
+                    defaultValue: new Expr.Literal(new Int32(10))
+                }
+            ],
+            returns: ValueKind.Dynamic
         },
-        impl: (interpreter: Interpreter, s: BrsString): Float => {
-            return new Float(Number(s.value));
+        impl: (interpreter: Interpreter, s: BrsString, brsRadix: Int32): BrsNumber => {
+            function isBrsStrFloat(str: BrsString): boolean {
```
sjbarag: > NOTE: I have a really cheesy way to detect if a string is a float just be checking for a decimal place. Open to other ideas.

You could probably use [Number.isInteger](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isInteger) here in combination with [Number.isNan](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Number/isNaN)?

PR #75, opened 2018-11-22
implement Substitute function (issue #51)

### LINKED ISSUE #44: Implement global string functions
We've already got four of these, but it'd be great if we had a few more!  Specifically, there's a bunch of [global string functions]() in BrightScript:

* `Instr(start as Integer, text as String, substring as String) as Integer`
* `Left(s as String, n as Integer) as String`
* `Len(s as String) as Integer`
* `Mid(s as String, p as Integer, [n as Integer]) as String`
* `Right(s as String, n as Integer) as String`
* `Str(value as Float) as String`
* `StrI(value as Integer) as String`
* `StrI(value as Integer, radix as Integer) as String`
* `String(n as Integer, str as String ) as String`
* `StringI(n as Integer, ch as Integer) as String`
* `Val(s as String) as Float`
* `Val(str as String, radix as Integer) as Integer`
* `Substitute(str as String, arg0 as String, arg1 = "" as String, arg2 = "" as String, arg3 = "" as String) as String`

### LINKED ISSUE #51: Implement global miscellaneous string functions
I kind of wish these fit into a better bucket, but they don't.  See [the BrightScript documentation](https://sdkdocs.roku.com/display/sdkdoc/Global+String+Functions) for details on what each one does, but this should cover:

* `Instr(start as Integer, text as String, substring as String) as Integer`
* `Len(s as String) as Integer`
* `String(n as Integer, str as String ) as String`
* `StringI(n as Integer, ch as Integer) as String`
* `Substitute(str as String, arg0 as String, arg1 = "" as String, arg2 = "" as String, arg3 = "" as String) as String`

strattonbrazil:
Substitute looks like it's the last string function to implement to close #44 and close #51.

### DIFF
--- a/src/stdlib/String.ts
+++ b/src/stdlib/String.ts
@@ -263,14 +263,11 @@ export const Substitute = new Callable(
             returns: ValueKind.String
         },
         impl: (interpreter: Interpreter, str: BrsString, arg0: BrsString, arg1: BrsString, arg2: BrsString, arg3: BrsString): BrsString => {
-            let tmpStr = str.value;
-            let replacements:{ [index:string]: BrsString } = { "0" : arg0, "1" : arg1, "2" : arg2, "3" : arg3 };
-            for (const index in replacements) {
-                let token = "\\{" + index + "\\}";
-                tmpStr = tmpStr.replace(new RegExp(token, "g"), replacements[index].value);
-            }
-
-            return new BrsString(tmpStr);
+            let completelyReplaced = [arg0, arg1, arg2, arg3].reduce(
+                (replaced, replacement, index) => replaced.replace(new RegExp(`\\{${index}\\}`, "g"), replacement.value),
+                str.value
+            );
+            return new BrsString(completelyReplaced);
         }
     }
 );
--- a/src/stdlib/String.ts
+++ b/src/stdlib/String.ts
@@ -264,8 +264,8 @@ export const Substitute = new Callable(
         },
         impl: (interpreter: Interpreter, str: BrsString, arg0: BrsString, arg1: BrsString, arg2: BrsString, arg3: BrsString): BrsString => {
             let tmpStr = str.value;
-            let replacements = [ arg0, arg1, arg2, arg3 ];
-            for (let index in replacements) {
+            let replacements:{ [index:string]: BrsString } = { "0" : arg0, "1" : arg1, "2" : arg2, "3" : arg3 };
+            for (const index in replacements) {
                 let token = "\\{" + index + "\\}";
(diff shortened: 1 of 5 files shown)


src/stdlib/String.ts
```diff
@@ -227,6 +227,54 @@ export const StrI = new Callable(
...
+                    type: ValueKind.String,
+                    defaultValue: new Expr.Literal(new BrsString(""))
+                },
+            ],
+            returns: ValueKind.String
+        },
+        impl: (interpreter: Interpreter, str: BrsString, arg0: BrsString, arg1: BrsString, arg2: BrsString, arg3: BrsString): BrsString => {
+            let tmpStr = str.value;
+            let replacements = [ arg0, arg1, arg2, arg3 ];
+            for (let index in replacements) {
```
strattonbrazil: Using for-in on the array is breaking the linter.  I kind of like this syntax, but am willing to change it.  Any suggestions on a clean way to do this?  

    for (let i = 0; i < replacements.length; i++)

src/stdlib/String.ts
```diff
@@ -264,8 +264,8 @@ export const Substitute = new Callable(
         },
         impl: (interpreter: Interpreter, str: BrsString, arg0: BrsString, arg1: BrsString, arg2: BrsString, arg3: BrsString): BrsString => {
             let tmpStr = str.value;
-            let replacements = [ arg0, arg1, arg2, arg3 ];
-            for (let index in replacements) {
```
sjbarag: The `for ... in` lint check is enabled to avoid accidentally encountering the [weirdness of using it on arrays](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...in#Array_iteration_and_for...in).  Manually creating the `replacements` object feels like a bit of a hack, because it seems like `for in` is what you wanted here. There's a few options:

1. Use `Array.prototype.forEach` to get both the index and value at the same time:
    ```typescript
    replacements.forEach( (replacement, index) => {
            let token = `{${index}}`;
            tmpStr = tmpStr.replace(new RegExp(token, "g"), replacement.value); 
    });
    ```
2. (My preference) Remove `tmpStr` entirely and use `Array.prototype.reduce`:
    ```typescript
        let completelyReplaced = [arg0, arg1, arg2, arg3].reduce(
            (replaced, replacement, index) => replaced.replace(new RegExp(`{${index}}`, "g"), replacement.value),
            str.value
        );
    ```
2. Disable TSLint on that line (please be stingy with this though :smile: )
    ```typescript
    // tslint:disable-next-line:no-for-in-array
    for (let index in replacements) {
        // ...
    }
    ```

PR #84, opened 2018-11-29
chore(arch): Move lexer and friends into lexer/

sjbarag:
While having lexer/index.ts made sense from a node conventions standpoint, having a proper Lexer.ts makes this all much easier to work with (e.g. no need to type `lexer index` into a fuzzy-finding file opener; just type `lexer`).  This also allows for cleaner importing and exporting, sot hat our unit tests can test via the public export paths like any consumer would.  That should help prevent accidental breaking changes in the future.

No functionality changes here though, other than some things getting exported from the `brs` module's `main` file now.


sjbarag:
I'd like to do something similar for the `parser` and `interpreter` trees as well.  Hopefully one day soon!

PR #89, opened 2018-12-01
implements file reads/writes using memfs library (#88)

### LINKED ISSUE #88: implement file reading and writing global utilities
Implement the global utilities related to file reading a writing (https://sdkdocs.roku.com/display/sdkdoc/Global+Utility+Functions).  

* ReadAsciiFile(filepath as String) as String
* WriteAsciiFile(filepath as String, text as String) as Boolean

strattonbrazil:
This adds the ReadAsciiFile() and WriteAsciiFile() functions to stdlib.  This implementation uses the _memfs_ library so all file operations will be in memory.  This will make working with files safer during brs development.  In the future we can support reading from the hard drive by copying files into the memfs volumes at startup of the interpreter.  

Right now only a single volume, _tmp_, is implemented and is attached to the interpreter object.  In the future we can add additional ones such as _common_ and _package_.  

I've written unit tests for this, but am having trouble with the end-to-end tests for some reason.  

closes #88

### DIFF
--- a/src/stdlib/File.ts
+++ b/src/stdlib/File.ts
@@ -1,17 +1,8 @@
 import { Callable, ValueKind, BrsString, BrsBoolean } from "../brsTypes";
 import { Interpreter } from "../interpreter";
 import { Volume } from "memfs/lib/volume";
+import { URL } from "url";
 
-/* Returns the brs-path uri scheme or an empty string if not found */
-export function parseVolumeName(path: string): string {
-    const volumeRegex = new RegExp("^([a-z]+):");
-    const volumePrefix = path.match(volumeRegex);
-    if (volumePrefix !== null) {
-        return volumePrefix[1];
-    } else {
-        return "";
-    }
-}
 
 /*
  * Returns a memfs volume based on the brs path uri.  For example, passing in
@@ -20,22 +11,25 @@ export function parseVolumeName(path: string): string {
  * Returns invalid in no appopriate volume is found for the path
  */
 export function getVolumeByPath(interpreter: Interpreter, path: string): Volume | null {
-    const volumePrefix = parseVolumeName(path);
-    if (volumePrefix === "tmp") {
-        return interpreter.temporaryVolume;
-    } else {
+    try {
+        const protocol = new URL(path).protocol;
+        switch (protocol) {
+            case "tmp:":
+                return interpreter.temporaryVolume;
+            default:
+                return null;
+        }
+    } catch (err) {
         return null;
     }
 }
 
 /*
- * Removed the scheme from the uri.
- *    ex. "tmp:///test/test1.txt" -> "///test/test1.txt"
- * 
- * NOTE: memfs as well as other file systems correctly handle extraneous slashes
+ * Returns a memfs file path from a brs file uri 
+ *   ex. "tmp:///test/test1.txt" -> "/test/test1.txt"
  */
-function stripScheme(brsUri: string): string {
-    return brsUri.replace(parseVolumeName(brsUri) + ":", "");
+export function getMemfsPath(fileUri: string) {
+    return new URL(fileUri).pathname;
 }
 
 /** Reads ascii file from file system. */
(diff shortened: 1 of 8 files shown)


src/stdlib/File.ts
```diff
@@ -0,0 +1,87 @@
...
+}
+
+/*
+ * Returns a memfs volume based on the brs path uri.  For example, passing in
+ * "tmp:///test.txt" will return the memfs temporary volume on the interpreter. 
+ * 
+ * Returns invalid in no appopriate volume is found for the path
+ */
+export function getVolumeByPath(interpreter: Interpreter, path: string): Volume | null {
+    const volumePrefix = parseVolumeName(path);
```
sjbarag: Consider using `node`'s built-in `URL` class instead of writing this parser yourself - it'll save us some trouble and we won't have to test the implementation :smile: Something like this would probably work:

```
try {
    let url = new URL(path);
    if (url.scheme === "tmp:") return interpreter.temporaryVolume;
catch (err) {
    return null;
}
```

src/stdlib/File.ts
```diff
@@ -0,0 +1,87 @@
...
+    {
+        signature: {
+            args: [
+                {name: "filepath", type: ValueKind.String}
+            ],
+            returns: ValueKind.String
+        },
+        impl: (interpreter: Interpreter, filepath: BrsString, text: BrsString) => {
+            const volume = getVolumeByPath(interpreter, filepath.value);
+            if (volume === null) {
```
sjbarag: `tslint` is configured to allow `==` comparisons against `null`, because it matches both `null` and `undefined`.  Might be worth using that here.

src/stdlib/File.ts
```diff
@@ -0,0 +1,87 @@
+import { Callable, ValueKind, BrsString, BrsBoolean } from "../brsTypes";
+import { Interpreter } from "../interpreter";
+import { Volume } from "memfs/lib/volume";
```
sjbarag: Try to avoid importing things from other modules using their file path - it's entirely likely that they'd do a reorganization internally and not do a major version bump because the public interface has changed.  Consider following the patterns `memfs` recommends: https://github.com/streamich/memfs/blob/master/docs/reference.md

src/stdlib/File.ts
strattonbrazil: Thanks. Switched to this, which also handles unknown volumes:

```
    try {
        const protocol = new URL(path).protocol;
        if (protocol === "tmp:") return interpreter.temporaryVolume;
    } catch (err) {
        return null;
    }
    return null;
```

src/stdlib/File.ts
strattonbrazil: What's the advantage?

src/stdlib/File.ts
strattonbrazil: The docs might not match the memfs library I ended up using.  The latest one had some missing dependencies.  

That makes sense, but what is the typescript type _getVolumeByPath_ would return?  Volume and vol are constructors, but not a type from what I can tell.  

```
// Volume is a constructor function for creating new volumes:
import {Volume} from 'memfs';

function getVolumeByPath(..): Volume {
}
```
This raises the following error:

> Return type of exported function has or is using private name 'Volume'.

I can simply remove the return by from _getVolumeByPath_, but I'm hoping we can keep that typing.

sjbarag: You'd probably have to return `typeof vol` to get the compile-time type associated with that constructor.  It's a bit of a bummer, but I have a feeling `memfs`'s declarations use old-style constructor functions instead of classes.

src/stdlib/File.ts
sjbarag: Nothing major, but anything that accidentally returns `undefined` is also handled.

src/stdlib/File.ts
sjbarag: (updating here for posterity): @strattonbrazil and I talked out-of-band and agreed it's fine to reference by file path for now since there's a `yarn.lock` in place – especially when there's some discrepancies between the `memfs` docs and the implementation we're using.

PR #90, opened 2018-12-03
feat(parse,interp): Implement setters for arrays and associative arrays

sjbarag:
Immutability is cool and all, but having write-once datatypes is pretty silly.  RBI allows mutation of existing arrays and associative arrays without creating new ones and overwriting the old ones, so we should too!  Now the following should work:

```brightscript
a = [1, 2, 3]
a[0] = "one"

b = { foo: "bar" }
b.lorem = [ "ipsum", "dolor", "sit", "amet" ]
```


src/interpreter/index.ts
```diff
@@ -575,7 +575,7 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
         let source = this.evaluate(expression.obj);
         if (!isIterable(source)) {
             throw BrsError.typeMismatch({
-                message: "Attemptin to retrieve property from non-iterable value",
```
strattonbrazil: I prefer the casualness of "attemptin".

src/interpreter/index.ts
```diff
@@ -776,7 +776,61 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
     visitM(expression: Expr.M) {
         return BrsInvalid.Instance;
     }
-    visitSet(expression: Expr.Set) {
+
+    visitDottedSet(expression: Stmt.DottedSet) {
```
strattonbrazil: Is this expression one line or can it be an entire code block?

test/brsTypes/Array.test.js
```diff
@@ -39,4 +39,46 @@ describe("Array", () => {
...
+            let arr = new BrsArray([a, b, c]);
+
+            arr.set(new Int32(0), new BrsString("replacement for a"));
+            arr.set(new Int32(2), new BrsString("replacement for c"));
+
+            expect(arr.get(new Int32(0))).toEqual(new BrsString("replacement for a"));
+            expect(arr.get(new Int32(2))).toEqual(new BrsString("replacement for c"));
+        });
+
+        it("sets values at out-of-bounds indexes", () => {
```
strattonbrazil: I didn't realize brightscript arrays could do this.  I assume any new indices in between would be undefined?  If so, might be good to add a check for some index like 2xx to verify it's undefined.

src/interpreter/index.ts
sjbarag: Whoops, forgot to rename the variable in that last commit. Should have been `statement`.

Code blocks are considered statements on their own (see `Statement.Block`. The new value from a `Set` is an expression — IIRC `Expr.Call` and `Expr.IndexedGet` are about the most complex expressions besides anonymous functions. `foo["Lorem-ipsum"] = bar.baz()` is supported. Same with anonymous functions 😉

test/brsTypes/Array.test.js
sjbarag: I could have sworn it was supported but I never actually  checked RBIs behavior. I can do that today – adding bounds checking shouldn’t be too hard if necessary.

strattonbrazil: I was just unaware.  Per the docs: 

> Sets an entry at a given index to the passed value.  If index is beyond the bounds of the array, the array is expanded to accommodate it.

sjbarag:
Interesting, apparently commas are supported within bracketed Gets as well (via https://sdkdocs.roku.com/plugins/servlet/mobile#Expressions,Variables,andTypes-ArrayOperator ). I’ll have to open an issue for that.

>>>
Arrays in BrightScript are one dimensional. Multi-dimensional arrays are implemented as arrays of arrays. The "[ ]" operator will automatically map a list of indexes separated by commas to the appropriate sequence of indexing. For example, the following two expressions to fetch "item" are the same:

Example
```
dim array[5,5,5]
item = array[1][2][3]
item = array[1,2,3]
```

test/brsTypes/Array.test.js
sjbarag: Neato! Got a link for that handy?

strattonbrazil: https://sdkdocs.roku.com/display/sdkdoc/ifArraySet#ifArraySet-SetEntry(indexAsInteger,tvalueAsDynamic)AsVoid

PR #100, opened 2018-12-04
Implement StrToI() function (Issue #94)

### LINKED ISSUE #94: implement global function StrToI
StrToI is a global "utility" function related to string parsing we should implement to match the spec (https://sdkdocs.roku.com/display/sdkdoc/Global+Utility+Functions#GlobalUtilityFunctions-StrToI(strasString)asDynamic).  

* StrToI(str as String) as Dynamic

jweeber:
This implements brightscript's `StrToI()` function (#94). This function returns an integer from a string or 0 if it can't be parsed.

`StrToI()` documentation: https://sdkdocs.roku.com/display/sdkdoc/Global+Utility+Functions#GlobalUtilityFunctions-StrToI(strasString)asDynamic

closes #94


src/stdlib/String.ts
```diff
@@ -273,7 +273,7 @@ export const Substitute = new Callable(
 );
 
 /**
- * Return a float from a string
+ * Return a float or integer from a string
```
sjbarag: Oooh good catch!  Thanks for fixing that 💯

PR #101, opened 2018-12-04
feat(interp,env): Implement `m` pointer

sjbarag:
The Reference BrightScript Implementation includes a parallel to JavaScript's `this` variable:

> A BrightScript object is an roAssociativeArray which contains function pointers.  When a member function is called "from" an AssociativeArray, the special variable "m" is set to point to that AssociativeArray. "m" is accessible inside the called function to access other data in the AssociativeArray object.

(https://sdkdocs.roku.com/display/sdkdoc/Component+Architecture#ComponentArchitecture-%22m%22TheBrightScript%22thispointer%22)

This BrightScript implementation needs to have support for it as well.

Depends on #95

### DIFF
--- a/src/interpreter/index.ts
+++ b/src/interpreter/index.ts
@@ -514,7 +514,7 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
 
         if (satisfiedSignature) {
             try {
-                let mPointer = new AssociativeArray([]);
+                let mPointer = this._environment.getM();
 
                 if (expression.callee instanceof Expr.DottedGet || expression.callee instanceof Expr.IndexedGet) {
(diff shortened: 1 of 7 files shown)


src/interpreter/index.ts
```diff
@@ -515,7 +514,23 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
 
         if (satisfiedSignature) {
             try {
-                return callee.call(this, ...args);
+                let mPointer = this._environment.getM();
+
+                if (expression.callee instanceof Expr.DottedGet || expression.callee instanceof Expr.IndexedGet) {
```
sjbarag: This is where the "magic" of the `m` pointer happens, for anyone reading this PR

PR #103, opened 2018-12-04
fix(parse): Allow `as object` argument type suffix

### LINKED ISSUE #102: Using `as object` in a function signature causes parse failure
The Reference BrightScript Implementation allows `as object` to be used as a type suffix for function arguments and function returns, but it currently causes a parse failure here:

Failing file:
```brightscript
sub Main()
    test(invalid)
end sub

sub test(foo as object)
    print "if you can read this, the bug has been fixed"
end sub

Main()
```

Failed output:
```
$ brs ./object-mwe.brs
[Line 5] Function parameter [object Object] is of invalid type 'object'
[Line 7] Found unexpected token 'end sub'
Error occurred
```

sjbarag:
`as object` is basically the same as `as dynamic`, except that `as object` automatically boxes primitive numbers before passing them into the function.  Since we don't have support for boxed numbers yet, we can ignore that for now.

closes #102


src/brsTypes/BrsType.ts
```diff
@@ -60,6 +62,7 @@ export namespace ValueKind {
             case "dynamic": return ValueKind.Dynamic;
             case "void": return ValueKind.Void;
             case "<uninitialized>": return ValueKind.Uninitialized;
+            case "object": return ValueKind.Object;
```
strattonbrazil: Should this be the ValueKind we'll use for objects created with createObject()?

sjbarag: Hmm, probably!  It'd certainly align with my current understanding of `as object` automatically boxing primitive values (e.g. `int` -> `roInt`).

PR #104, opened 2018-12-05
chore(interp): Don't require explicit stdlib registration

sjbarag:
We had stdlib function names duplicated a whole bunch, and it was really easy to forget to register a stdlib function.  Now it's automatic!  Just export it from `src/stdlib/` and it'll get registered.

### DIFF
--- a/src/interpreter/index.ts
+++ b/src/interpreter/index.ts
@@ -14,7 +14,8 @@ import {
     isIterable,
     SignatureAndMismatches,
     MismatchReason,
-    isComparable
+    isComparable,
+    Callable
 } from "../brsTypes";
 
 import * as Expr from "../parser/Expression";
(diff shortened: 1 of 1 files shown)


sjbarag:
(whoops wrong button)

src/interpreter/index.ts
```diff
@@ -56,54 +57,12 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
...
-            { name: "WriteAsciiFile", func: StdLib.WriteAsciiFile },
-            { name: "StrToI",       func: StdLib.StrToI }
-        ].forEach(({name, func}) =>
-            this._environment.define(Scope.Global, name, func)
-        );
+        Object.keys(StdLib)
+            .map(name => (StdLib as any)[name])
+            .filter(func => func instanceof Callable)
+            .forEach((func: Callable) =>
+                this._environment.define(Scope.Global, func.name || "", func)
```
strattonbrazil: Why would we default to a blank string?  Would that just never get picked up?

sjbarag: True!  I should probably make that warn, brb.

PR #105, opened 2018-12-06
Implement CreateObject()

jweeber:
This implements brightscript's `createObject()` and a new `BrsObjects` map that will contain a list of brightscript components that can be created.

### DIFF
--- a/src/stdlib/CreateObject.ts
+++ b/src/stdlib/CreateObject.ts
@@ -1,0 +1,19 @@
+import { Callable, ValueKind, BrsInvalid, BrsString } from "../brsTypes";
+import { BrsObjects } from "../brsTypes/components/BrsObjects";
+import { Interpreter } from "../interpreter";
+import { createObjectBindingPattern } from "typescript";
+import { AssociativeArray } from "../brsTypes/components/AssociativeArray";
+
+export const CreateObject = new Callable(
+    "CreateObject",
+    {
+        signature: {
+            args: [{ name: "objName", type: ValueKind.String }],
+            returns: ValueKind.Dynamic
+        },
+        impl: (interpreter: Interpreter, objName: BrsString) => {
+            let ctor = BrsObjects.get(objName.value.toLowerCase());
+            return ctor ? ctor() : BrsInvalid.Instance;
+        }
+    }
+);
--- a/src/stdlib/CreateObject.ts
+++ b/src/stdlib/CreateObject.ts
@@ -1,8 +1,6 @@
 import { Callable, ValueKind, BrsInvalid, BrsString } from "../brsTypes";
 import { BrsObjects } from "../brsTypes/components/BrsObjects";
 import { Interpreter } from "../interpreter";
-import { createObjectBindingPattern } from "typescript";
-import { AssociativeArray } from "../brsTypes/components/AssociativeArray";
 
 /** Creates a new instance of a given brightscript component (e.g. roAssociativeArray) */
 export const CreateObject = new Callable(
--- a/src/stdlib/CreateObject.ts
+++ b/src/stdlib/CreateObject.ts
@@ -4,6 +4,7 @@ import { Interpreter } from "../interpreter";
 import { createObjectBindingPattern } from "typescript";
 import { AssociativeArray } from "../brsTypes/components/AssociativeArray";
 
+/** Creates a new instance of a given brightscript component (e.g. roAssociativeArray) */
 export const CreateObject = new Callable(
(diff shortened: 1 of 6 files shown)


src/stdlib/CreateObject.ts
```diff
@@ -0,0 +1,20 @@
+import { Callable, ValueKind, BrsInvalid, BrsString } from "../brsTypes";
+import { BrsObjects } from "../brsTypes/components/BrsObjects";
+import { Interpreter } from "../interpreter";
+import { createObjectBindingPattern } from "typescript";
+import { AssociativeArray } from "../brsTypes/components/AssociativeArray";
```
strattonbrazil: Is this still needed?

src/stdlib/CreateObject.ts
```diff
@@ -0,0 +1,20 @@
...
+export const CreateObject = new Callable(
+    "CreateObject",
+    {
+        signature: {
+            args: [{ name: "objName", type: ValueKind.String }],
+            returns: ValueKind.Dynamic
+        },
+        impl: (interpreter: Interpreter, objName: BrsString) => {
+            let ctor = BrsObjects.get(objName.value.toLowerCase());
+            return ctor ? ctor() : BrsInvalid.Instance;
```
strattonbrazil: Nice and neat.

test/stdlib/CreateObject.test.ts
```diff
@@ -0,0 +1,16 @@
+const { AssociativeArray, BrsObjects, CreateObject, BrsInvalid } = require("../../lib/brsTypes");
+const BrsError = require("../../lib/Error");
+
+describe("CreateObject", () => {
+    beforeEach(() => BrsError.reset());
```
strattonbrazil: Is this needed if we're not checking errors?  I honestly don't know as other tests don't have it.

test/stdlib/CreateObject.test.ts
```diff
@@ -0,0 +1,16 @@
...
+
+    it("creates a new instance of associative array", () => {
+        let obj = CreateObject("roAssociativeArray");
+        expect(obj).toEqual(new AssociativeArray([]));
```
strattonbrazil: How does equalTo work here?  Where is the comparison implemented?  Curious for future objects we add.

sjbarag: That comes [from jest](https://jestjs.io/docs/en/expect#toequalvalue), and performs a deep equality check between two JavaScript objects.

test/brsTypes/BrsObjects.test.js
```diff
@@ -0,0 +1,13 @@
+const { AssociativeArray, BrsObjects } = require("../../lib/brsTypes");
+const BrsError = require("../../lib/Error");
+
+describe("BrsObjects", () => {
+    beforeEach(() => BrsError.reset());
+
+    describe("new object instances", () => {
```
sjbarag: Neato!  These are great to have.  As we add more components this may become a pain to navigate, but we'll deal with that in the future 🤷‍♂️

test/stdlib/CreateObject.test.ts
sjbarag: You're right, it's not strictly necessary.  Doesn't hurt anything I suppose but it also doesn't add anything.

sjbarag:
I was going to ask about an end-to-end test for this, but I realized I'd rather handle that as part of testing methods on various BrightScript components.

src/interpreter/index.ts
```diff
@@ -100,7 +100,8 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
             { name: "Sgn",          func: StdLib.Sgn },
             { name: "ReadAsciiFile", func: StdLib.ReadAsciiFile },
             { name: "WriteAsciiFile", func: StdLib.WriteAsciiFile },
-            { name: "StrToI",       func: StdLib.StrToI }
+            { name: "StrToI",       func: StdLib.StrToI },
+            { name: "CreateObject", func: StdLib.CreateObject }
```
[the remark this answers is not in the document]
sjbarag: Whoops I hadn't merged that yet when I wrote my comment – just make sure you've got https://github.com/sjbarag/brs/commit/4b067b6e876fcd2f43878914bbafd44fd24fcb7e locally 😃

test/stdlib/CreateObject.test.ts
jweeber: Yep, copy 🍝 fail. Removed these.

PR #106, opened 2018-12-06
implements file I/O operations for issue #87

### LINKED ISSUE #87: implement file I/O copy/move/delete/list global utilities
Implement the global utilities related to file operations besides read/write (https://sdkdocs.roku.com/display/sdkdoc/Global+Utility+Functions).  

* ListDir(path as String) as Object
* CopyFile(source as String, destination as String) as Boolean
* MoveFile(source as String, destination as String) as Boolean
* DeleteFile(file as String) as Boolean
* DeleteDirectory(dir as String) as Boolean
* CreateDirectory(dir as String) as Boolean
* FormatDrive(drive as String , fs_type as String) as Boolean

FormatDrive() should be stubbed out as we shouldn't ever implement it.

strattonbrazil:
This implements the file functions not already covered with the exception of MatchFiles, which I'll move to another issue.  Most of these functions wrap existing fs functions that have similar behavior (ex. not allowing deletion of a non-empty directory).  

Functions defined include:
* ListDir
* CopyFile
* MoveFile
* DeleteFile
* DeleteDirectory
* CreateDirectory
* FormatDrive

Although, _FormatDrive_ is stubbed as not implemented.  

Closes #87

### DIFF
--- a/src/stdlib/File.ts
+++ b/src/stdlib/File.ts
@@ -189,7 +189,9 @@ export const FormatDrive = new Callable(
             returns: ValueKind.Boolean
         },
         impl: (interpreter: Interpreter, dir: BrsString) => {
-            console.warn("`FormatDrive` is not implemented in `brs`.");
+            if (process.env.NODE_ENV !== "test") {
+                console.error("`FormatDrive` is not implemented in `brs`.");
+            }
             return BrsBoolean.False;
         }
     }
--- a/src/stdlib/File.ts
+++ b/src/stdlib/File.ts
@@ -1,8 +1,7 @@
-import { Callable, ValueKind, BrsString, BrsBoolean } from "../brsTypes";
+import { Callable, ValueKind, BrsString, BrsBoolean, BrsArray } from "../brsTypes";
 import { Interpreter } from "../interpreter";
-import { Volume } from "memfs/lib/volume";
 import { URL } from "url";
-
+const MemoryFileSystem = require("memory-fs");
 
 /*
  * Returns a memfs volume based on the brs path uri.  For example, passing in
@@ -10,7 +9,7 @@ import { URL } from "url";
  * 
  * Returns invalid in no appopriate volume is found for the path
  */
-export function getVolumeByPath(interpreter: Interpreter, path: string): Volume | null {
+export function getVolumeByPath(interpreter: Interpreter, path: string): typeof MemoryFileSystem | null {
     try {
         const protocol = new URL(path).protocol;
         if (protocol === "tmp:") return interpreter.temporaryVolume;
@@ -27,6 +26,32 @@ export function getVolumeByPath(interpreter: Interpreter, path: string): Volume 
 export function getMemfsPath(fileUri: string) {
     return new URL(fileUri).pathname;
 }
+
+export const ListDir = new Callable(
+    "ListDir",
+    {
+        signature: {
+            args: [
+                {name: "path", type: ValueKind.String}
+            ],
+            returns: ValueKind.Array // TODO: change this to roList when available
+        },
+        impl: (interpreter: Interpreter, path: BrsString) => {
+            const volume = getVolumeByPath(interpreter, path.value);
+            if (volume === null) {
+                return new BrsArray([]);
+            }
+
+            const memfsPath = getMemfsPath(path.value);
+            try {
+                let subPaths = volume.readdirSync(memfsPath);
+                return new BrsArray(subPaths);
+            } catch (err) {
+                return new BrsArray([]);
+            }
+        }
+    }
+)
 
 /** Reads ascii file from file system. */
 export const ReadAsciiFile = new Callable(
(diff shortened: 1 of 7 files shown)


src/stdlib/File.ts
```diff
@@ -1,16 +1,15 @@
...
+import MemoryFileSystem from "memory-fs";
 
 /*
  * Returns a memfs volume based on the brs path uri.  For example, passing in
  * "tmp:///test.txt" will return the memfs temporary volume on the interpreter. 
  * 
  * Returns invalid in no appopriate volume is found for the path
  */
-export function getVolumeByPath(interpreter: Interpreter, path: string): Volume | null {
+export function getVolumeByPath(interpreter: Interpreter, path: string): MemoryFileSystem | null {
```
sjbarag: Obviously it wasn't changed here but I don't think we need to export `getVolumeByPath` – seems like it's only used within this file and for testing, right?  I ask only because it gets exported with the rest of `StdLib` even though it's not a `Callable`.  It might be worth moving this (and `getMemFsPath`) into a separate file so they don't get exported with the rest of the standard library, but that can happen later.

src/stdlib/File.ts
```diff
@@ -28,6 +27,201 @@ export function getMemfsPath(fileUri: string) {
...
+    {
+        signature: {
+            args: [
+                {name: "drive", type: ValueKind.String},
+                {name: "fs_type", type: ValueKind.String}
+            ],
+            returns: ValueKind.Boolean
+        },
+        impl: (interpreter: Interpreter, dir: BrsString) => {
+            console.warn("`FormatDrive` is not implemented in `brs`.");
```
sjbarag: This'll warn every time `formatDrive()` is called, right?  That's probably better than warning once like we do for `RebootSystem`.

src/stdlib/File.ts
```diff
@@ -28,6 +27,201 @@ export function getMemfsPath(fileUri: string) {
...
+
+/** Returns an array of paths in a directory */
+export const ListDir = new Callable(
+    "ListDir",
+    {
+        signature: {
+            args: [
+                {name: "path", type: ValueKind.String}
+            ],
+            returns: ValueKind.Array // TODO: change this to roList when available
```
sjbarag: re: your todo – this'll likely be `ValueKind.Object` for `roList`, which matches the docs pretty nicely IMO

src/stdlib/File.ts
```diff
@@ -28,6 +27,201 @@ export function getMemfsPath(fileUri: string) {
...
+        },
+        impl: (interpreter: Interpreter, path: BrsString) => {
+            const volume = getVolumeByPath(interpreter, path.value);
+            if (volume === null) {
+                return new BrsArray([]);
+            }
+
+            const memfsPath = getMemfsPath(path.value);
+            try {
+                let subPaths = volume.readdirSync(memfsPath).map((s) => new BrsString(s));
```
sjbarag: Nice use of `Array.prototype.map` here!

src/stdlib/File.ts
strattonbrazil: Ah, when I started writing this the functions weren't auto-imported so I thought it was fine for testing.  I see the auto-import already filters out non-Callables.  What's the concern of leaving this here for testing?  Is there a better way to expose this so it can still be testing?

src/stdlib/File.ts
strattonbrazil: Right.  The thing I don't like about this is there's a warning now whenever we run the tests, which is an eye-sore and will cause confusion later on.  I was wondering if we should wrap this in some env check to only warn in release builds.

strattonbrazil:
I'll write some e2e tests for this soon.

src/stdlib/File.ts
sjbarag: Oh right!  I forgot how the timing worked out there 😅 

I'm mostly concerned with exporting what is essentially a private method – I'm not sure you'd want anyone using this outside of `File.ts`, so exporting it makes the tacit statement that it's okay to do that.  Probably the only way to expose it for testing without exporting it out of `StdLib` is to pull these two functions into a separate file (`FileUtils.ts` or something) and test them by importing that file directly.  They could still be used here by importing `FileUtils`, but they wouldn't get exported from `StdLib` because `stdlib/index.ts` wouldn't have an `export * from "./FileUtils";` 😄

src/stdlib/File.ts
sjbarag: Good idea!  We could definitely do that – similar things already happen in [Error.ts](https://github.com/sjbarag/brs/blob/0719932c49e9e7fe9e709f97b91966320dc42921/src/Error.ts#L13-L15).  Also try replacing this with `console.error`, because the `stderr` output tends to get silenced for unit tests anyway (at least I thought?)

src/stdlib/File.ts
strattonbrazil: I'm not sure I see the different if stdlib is filtering out the non-Callables.  I did give it a brief attempt, but ran into some import issues due to my inexperience.  We can talk more about this offline.

src/stdlib/File.ts
strattonbrazil: I see console.warn in the output.  According to the docs _warn_ is an alias for _error_, so it also goes to _stderr_.  

https://nodejs.org/api/console.html#console_console_warn_data_args

PR #119, opened 2018-12-17
Add FormatJson, ParseJson stdlib functions

jwfearn:
Add [`FormatJson`](https://sdkdocs.roku.com/display/sdkdoc/Global+Utility+Functions#GlobalUtilityFunctions-FormatJson(jsonasObject,flags=0asInteger)asString) and [`ParseJson`](https://sdkdocs.roku.com/display/sdkdoc/Global+Utility+Functions#GlobalUtilityFunctions-ParseJson(jsonStringasString)asObject) standard library functions and tests. Currently the `flags` argument to `FormatJson` is ignored.

Tests use some BRS-specific custom expectations.


src/stdlib/Json.ts
```diff
@@ -0,0 +1,132 @@
...
+export const ParseJson = new Callable("ParseJson", {
+    signature: { returns: ValueKind.Dynamic, args: [
+        { name: "jsonString", type: ValueKind.String }
+    ]},
+    impl: (_: Interpreter, jsonString: BrsString) => {
+        try {
+            return brsValueOf(JSON.parse(jsonString.toString()));
+        } catch (err) {
+            // example RBI error:
+            // "BRIGHTSCRIPT: ERROR: ParseJSON: Unknown identifier 'x': pkg:/source/main.brs(25)"
```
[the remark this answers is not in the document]
jwfearn: My approach was to use the same prefix (namely `"BRIGHTSCRIPT: ERROR: ParseJSON: "`) and implementation-defined error details. I think we'd have to write our own JSON parser rather than using the builtin TS/JS one. I suppose would could add various special cases to match specific known errors. I vote we add a ticket for this work as a nice-to-have.

src/stdlib/Json.ts
```diff
@@ -0,0 +1,132 @@
...
+import { Interpreter } from "../interpreter";
+import { Literal } from "../parser/Expression";
+import {
+    BrsBoolean,
+    BrsInvalid,
+    BrsString,
+    BrsType,
+    BrsValue,
+    Callable,
+    Float as BrsFloat,
```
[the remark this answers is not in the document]
jwfearn: I will do this as the last step after any other changes. I find it a lot easier to work with the code when it has consistent naming.

test/stdlib/Json.test.js
```diff
@@ -0,0 +1,231 @@
...
+} = require("../../lib/brsTypes");
+
+// aliases
+const BrsAssociativeArray = AssociativeArray;
+const BrsFloat = Float;
+const BrsInteger = Int32;
+const BrsLongInteger = Int64;
+const BrsUninitialized = Uninitialized;
+
+expect.extend({
```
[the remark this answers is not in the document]
jwfearn: Why? Then we lose the names, which I thought made some sense.

src/stdlib/Json.ts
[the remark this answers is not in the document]
jwfearn: I'm not sure how `parseJson` and formatJson` would have access to the file and line number from the call site to add to the error message. Is that info available from the Interpreter?

src/stdlib/Json.ts
```diff
@@ -0,0 +1,132 @@
...
+        return fn(key, brsAa.get(key));
+    });
+}
+
+type ElementFn = (v: BrsJsonValue) => string;
+function elementsMap(brsArray: BrsArray, fn: ElementFn) {
+    return brsArray.getElements().map(fn);
+}
+
+function jsonOfItem(k: BrsString, v: BrsValue): string {
```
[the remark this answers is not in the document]
jwfearn: done

src/stdlib/Json.ts
```diff
@@ -0,0 +1,132 @@
...
+                };
+            })
+        );
+    default:
+        throw new Error(`brsValueOf not implemented for: ${x} <${t}>`);
+    }
+}
+
+type ItemFn = (k: BrsString, v: BrsValue) => string;
+function itemsMap(brsAa: BrsAssociativeArray, fn: ItemFn) {
```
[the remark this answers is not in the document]
jwfearn: these functions have been removed/inlined

src/stdlib/Json.ts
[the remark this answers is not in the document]
jwfearn: done

test/stdlib/Json.test.js
```diff
@@ -0,0 +1,231 @@
...
+            actual = ParseJson.call(interpreter, brsArrayStr);
+            expect(actual).toEqualBrsArray(brsArray);
+        });
+
+        it('converts to BRS associative array', () => {
+            actual = ParseJson.call(interpreter, brsAssociativeArrayStrAsc);
+            expect(actual).toEqualBrsAssociativeArray(brsAssociativeArrayDesc);
+        });
+    });
+});
```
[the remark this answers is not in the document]
jwfearn: done

src/stdlib/Json.ts
```diff
@@ -0,0 +1,132 @@
...
+
+function jsonOfItem(k: BrsString, v: BrsValue): string {
+    return `"${k.toString()}":${jsonOf(v)}`;
+}
+
+function jsonOf(x: BrsValue): string {
+    if (x instanceof BrsInvalid) {
+        return "null";
+    }
+    if (x instanceof BrsAssociativeArray) {
```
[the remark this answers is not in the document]
jwfearn: Nested object references are now rejected

src/stdlib/Json.ts
```diff
@@ -0,0 +1,127 @@
...
+    case ValueKind.Int64:
+        return x.toString();
+    case ValueKind.Object:
+        if (x instanceof AssociativeArray) {
+            try {
+                if (x.get(uid) !== BrsInvalid.Instance) {
+                    throw new Error("Nested object reference");
+                }
+                let elements = x.getElements();
+                x.set(uid, BrsBoolean.True);
```
sjbarag: This is a super neat approach!  I'm unfortunately not comfortable modifying elements in a formatting function.  Regardless of my comfort, attempting to set a random property won't work on a BrightScript array, which suffers from the same "Nested object reference" problem that associative arrays do, e.g.:

```brightscript
sub main()
    a = [1, 2, 3]
    a.push(a)
    print formatJson(a)
end sub
```

This seems like a good place to use a JavaScript [Set](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set) to keep track of the objects that have been visited so far.  Something like this should work:

```typescript
/**
  * Creates a JavaScript string containing the JSON representation of the BrightScript value `x`.
  * @param x the thing
  * @returns a string (oh look that rhymed!)
  */
function jsonOf(x: BrsType): string {
    let visitedElements = new Set();

    /** Function-local implementation, used to close over `visitedElements`.  Should only be called from within `__jsonOf`, recursively. */
    function __jsonOf(x: brsType): string {
        if (visitedElements.has(x)) {
            // duplicate element detected!
            // throw an error.
        }
        visitedElements.add(x);
        switch (x.kind) {
            // most of what you have already
        }
    }

    return __jsonOf(x);
}
```

src/stdlib/Json.ts
```diff
@@ -0,0 +1,127 @@
...
+    BrsString,
+    BrsType,
+    Callable,
+    Float,
+    Int32,
+    Int64,
+    ValueKind
+} from "../brsTypes";
+
+function brsValueOf(x: any): any {
```
sjbarag: This function returns a `BrsType`, not `any` – being strict with what functions return will help to prevent weird type-checking errors!  Looks like there's one change required to make that compile, which I'll comment on below.

src/stdlib/Json.ts
```diff
@@ -0,0 +1,127 @@
...
+        }
+        return new Float(x);
+    case "object":
+        if (Array.isArray(x)) {
+            return new BrsArray(x.map(brsValueOf));
+        }
+        return new AssociativeArray(
+            Object.getOwnPropertyNames(x).map((k: string) => {
+                return {
+                    name: brsValueOf(k),
```
sjbarag: An associative array's keys must be strings in brightscript, but there's no easy way to guarantee that `brsValueOf(someString)` will return a `BrsString` at compile-time.  Switching `brsValueOf` to return `BrsType` complains here because `BrsBoolean`, `BrsInvalid`, etc. aren't valid brightscript strings (naturally).  Since we know as humans that these need to be strings, and we know at compile time that `k` is a string, it's safe to just construct a `new BrsString(k)` here!  Plus, it's really straightforward 😄

src/stdlib/Json.ts
```diff
@@ -0,0 +1,127 @@
...
+                    value: brsValueOf(x[k])
+                };
+            })
+        );
+    default:
+        throw new Error(`brsValueOf not implemented for: ${x} <${t}>`);
+    }
+}
+
+function jsonOf(interpreter: Interpreter, x: BrsType, uid: BrsString): string {
```
sjbarag: Please add a JSDoc comment to `jsonOf` and `brsValueOf` to describe what they're used for and what significance the parameters and return values have.  I know they're not exported, but they've got some pretty significant chunks of logic and the meanings of everything aren't immediately obvious.

src/stdlib/Json.ts
```diff
@@ -0,0 +1,127 @@
...
+        }
+        break;
+    case ValueKind.Callable:
+    case ValueKind.Uninitialized:
+        break;
+    default:
+        const _: never = x;
+        break;
+    }
+    throw new Error(`jsonOf not implemented for: ${x}`);
```
sjbarag: This seems like the `default` case – any reason to not move it in there, and remove line 83?  Something like:

```typescript
switch (x.kind) {
    case ValueKind.Callable:
    case ValueKind.Uninitialized:
    default:
        throw new Error(`jsonOf not implemented for: ${x}`);
}
```

src/stdlib/Json.ts
```diff
@@ -0,0 +1,127 @@
...
+        }
+        if (x instanceof BrsArray) {
+            return `[${x.getElements().map((el: BrsType) => { return jsonOf(interpreter, el, uid); }).join(",")}]`;
+        }
+        break;
+    case ValueKind.Callable:
+    case ValueKind.Uninitialized:
+        break;
+    default:
+        const _: never = x;
```
sjbarag: If you get a minute, please let me know what this does!  Without this line I don't get any compiler errors for `x` being unused, because it's used in other branches.  But as far as I can tell it's unnecessary?

src/stdlib/Json.ts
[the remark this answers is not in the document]
sjbarag: > I'm not sure how `parseJson` and `formatJson` would have access to the file and line number from the call site to add to the error message. Is that info available from the Interpreter?

At the moment, call-site information is stored in tokens only.  An instance of `Call` (from [Expression.ts](https://github.com/sjbarag/brs/blob/1e17f358debeb8f7ee3ff8d88f1551ac8bba2c30/src/parser/Expression.ts#L55)) contains the closing `)` token from the function call, but it's not passed here.  Not having call site locations isn't a deal breaker right now, though it does limit the usefulness of these error messages.  Feel free to file an issue for it!


> My approach was to use the same prefix (namely `"BRIGHTSCRIPT: ERROR: ParseJSON: "`) and implementation-defined error details. I think we'd have to write our own JSON parser rather than using the builtin TS/JS one. I suppose would could add various special cases to match specific known errors. I vote we add a ticket for this work as a nice-to-have.

Good point!  Would you mind special-casing the empty string message in this PR?  I don't expect to need many others.  A BrightScript dev may not have come across the JavaScript message for empty JSON strings before, and I was initially a bit concerned about malformed JSON before I realized what the JavaScript message meant.  Of course, empty strings _are_ malformed JSON, but "Data is empty" is pretty clear.  The other messages seem fine to use the JavaScript version for though!

src/stdlib/Json.ts
```diff
@@ -0,0 +1,132 @@
...
+        return `"${x.toString()}"`;
+    }
+    if (!(x instanceof BrsUninitialized)) {
+        return x.toString();
+    }
+    throw new Error(`jsonValueOf not implemented for: ${x}`);
+}
+
+function logBrsErr(functionName: string, err: Error): void {
+    console.error(`BRIGHTSCRIPT: ERROR: ${functionName}: ${err.message}`);
```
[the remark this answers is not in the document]
sjbarag: Long term, you're probably right!  I see the current approach as a naïve/hack approximation of log levels.  It wouldn't be unreasonable to pass all logging through a facade (e.g. https://www.slf4j.org/) and configure tests to use a "NONE" logging level that no-ops calls to `log.debug()` and friends, so this doesn't feel unprecedented.

It hasn't bit anyone too hard yet – realistically this is only the third place in the codebase that's needed to message anything to users, and the other two were implementation specific warnings about a function being unimplemented (`RebootSystem` and `FormatDrive`, specifically).  This is the first case where a stdlib function prints something to match RBI, and the pain of adding in a quick `process.env.NODE_ENV` check hasn't yet outweighed the pain of stubbing out the global namespace via tests.

src/stdlib/Json.ts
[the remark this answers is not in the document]
jwfearn: done

src/stdlib/Json.ts
jwfearn: done

src/stdlib/Json.ts
jwfearn: This is the exhaustive check that you turned me on to: https://basarat.gitbooks.io/typescript/content/docs/types/discriminated-unions.html. If the switch statement above does not fully cover all types in the union, this statement will cause a compile error. To demonstrate, comment out line 82 and build. You'll see a compile error:
```
src/stdlib/Json.ts:85:15 - error TS2322: Type 'Uninitialized' is not assignable to type 'never'.
```

src/stdlib/Json.ts
```diff
@@ -0,0 +1,132 @@
...
+type ElementFn = (v: BrsJsonValue) => string;
+function elementsMap(brsArray: BrsArray, fn: ElementFn) {
+    return brsArray.getElements().map(fn);
+}
+
+function jsonOfItem(k: BrsString, v: BrsValue): string {
+    return `"${k.toString()}":${jsonOf(v)}`;
+}
+
+function jsonOf(x: BrsValue): string {
```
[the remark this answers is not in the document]
jwfearn: done

src/stdlib/Json.ts
jwfearn: Done, the Set approach is way cleaner. I had forgotten that JS keys can be any object not just strings. BrightScript is rotting my brain!

src/stdlib/Json.ts
jwfearn: done

src/stdlib/Json.ts
jwfearn: done

src/stdlib/Json.ts
jwfearn: We want the default branch to be unreachable so that it can contain the `never` assignment. See https://basarat.gitbooks.io/typescript/content/docs/types/discriminated-unions.html

src/stdlib/Json.ts
sjbarag: Ah, I see.  It's been a while since I've read through that page 😅

src/stdlib/Json.ts
```diff
@@ -0,0 +1,149 @@
...
+    Float,
+    Int32,
+    Int64,
+    ValueKind
+} from "../brsTypes";
+
+/**
+ * Converts a value to its representation as a BrsType. If no such
+ * representation is possible, throws an Error.
+ * @param {any} x Some value.
```
sjbarag: Nit for future reference: The type listed in `{curly braces}` isn't required when documenting TypeScript because the compiler (and the typescript language service which provides editor help) already knows what types these params are!  These are fine for now, but don't be surprised if they get removed in a later commit 😉

src/stdlib/Json.ts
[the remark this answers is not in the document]
sjbarag: Thanks!

test/stdlib/Json.test.js
```diff
@@ -0,0 +1,180 @@
...
+        process.env[k] = v;
+        return fn();
+    } finally {
+        process.env[k] = save;
+    }
+}
+
+function expectConsoleError(expected, fn) {
+    jest.spyOn(console, 'error').mockImplementationOnce((s) => {
+        expect(s).toMatch(expected);
```
sjbarag: Watch out! Because the assertion only happens during the stubbed call to `console.error`, any test that calls `expectConsoleError` will still pass!  Same with additional calls to `console.error`.  You'll likely want to have the assertion outside of the `mockImplementation` – I've got that working locally so I'll just push that in a second.

src/stdlib/Json.ts
jwfearn: Makes sense!

PR #121, opened 2018-12-21
Add BRS runtime error handling

jweeber:
While working on `createObject("roTimespan")`, I noticed when calling a method it wasn't logging or returning anything. With @sjbarag's help it was discovered that were just weren't showing runtime errors in the interpreter. Now we should show them so we know what's going on! Useful!

### DIFF
--- a/src/interpreter/index.ts
+++ b/src/interpreter/index.ts
@@ -81,6 +81,9 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
         try {
             this._environment = newEnv;
             return func(this);
+        } catch (err) {
+            console.error("Runtime error encountered in BRS implementation: ", err);
+            throw err;
         } finally {
             this._environment = originalEnvironment;
         }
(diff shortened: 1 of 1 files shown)


src/interpreter/index.ts
```diff
@@ -81,6 +81,9 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
         try {
             this._environment = newEnv;
             return func(this);
+        } catch (err) {
+            console.error("Runtime error encountered in BRS implementation: ", err);
```
sjbarag: Travis tests are currently failing and that's entirely my fault!  I forgot I'm throwing errors to handle `return` statements in functions, `exit for` and `exit while` to break out of loops, and… maybe one more thing?  Anyway, this should fix it:

```suggestion
            if (err.kind == null) {
                console.error("Runtime error encountered in BRS implementation: ", err);
            }
```

(I'm testing github's "suggested changes" feature so if this looks weird or doesn't work or something blame github 🤷‍♂️ https://help.github.com/articles/incorporating-feedback-in-your-pull-request/#applying-a-suggested-change)

PR #126, opened 2019-01-10
Implement createObject("roTimespan")

jweeber:
This implements brightscript's `roTimespan` component as well as the methods available for that component. 

https://sdkdocs.roku.com/display/sdkdoc/ifTimespan

Note that this will also add two new dependencies, `luxon` (allows us to parse all valid ISO8601 date formats) and `lolex` (allows us to mock a clock for Timespan tests).


.npmrc
```diff
@@ -0,0 +1 @@
+registry = "https://registry.yarnpkg.com/"
```
sjbarag: I've been meaning to do this!

src/brsTypes/components/Timespan.ts
```diff
@@ -0,0 +1,103 @@
...
+            },
+            impl: (_: Interpreter, date: BrsString) => {
+                let dateAsSeconds;
+                let now = Date.now();
+                let dateToParse = luxon.DateTime.fromISO(date.value, { zone: "utc" });
+
+                if (dateToParse.isValid) {
+                    dateAsSeconds = (Date.parse(dateToParse.toISO()) - now) / 1000;
+                } else {
+                    dateAsSeconds = 2077252342;
```
sjbarag: I'm so curious what's going to happen at Mon, 29 Oct 2035 06:32:22 +00:00 – I'd maybe understand the 2038 problem, but 2035 is off just enough where I feel like it's gotta be intentional.

jweeber:
Thanks @sjbarag! I'll look into creating an e2e test.

PR #140, opened 2019-01-21
feat: Allow programmatic access to errors

### LINKED ISSUE #136: How to get parse errors programmatically
I'm using the `Lexer` and `Parser` classes to try and validate a brightscript file. How can I get a list of parse errors? 

I'm doing the following in TypeScript: 

```typescript
import { Lexer, Parse } from 'brs';
let tokens = Lexer.scan(text);
let statements = Parser.parse(tokens);
```

Problem is, it just seems to eat any errors it encounters. For example, based on [this line](https://github.com/sjbarag/brs/blob/f35b21191fa81fc655f57da287f6ba0e525aa88e/src/parser/Parser.ts#L114) in `Parser.ts`, there should be a parse error thrown whenever you declare a sub with an `as` clause. 

Here's a very simple brightscript file that should have at least the above mentioned parse error

```brightscript
sub SayHello() as string 
    print "hello world"
end sub
```

I don't get any exceptions thrown...the statements array contains a single entry, representing the print line. If I remove the `as string`, then the statements array has one entry at the top level represneting the `SayHello` function (as expected).

What I am hoping to get is a list of parse errors, and their line numbers (which in this case would be 1 parse error, something to the effect of 

```typescript
{
    error: `"'Sub' functions are always void returns, and can't have 'as' clauses""`, 
    line: 1,
    column: 14,
    file: "theFile.brs"
}
```

Is this functionality available? If not, do you think it's do-able?

sjbarag:
BREAKING-CHANGE: When the lexer, parser, preprocessor, and interpreter were first written, I didn't expect them to be used outside of this project, and thus they never offered programmatic access to the errors they encounter!  #136 asked how to do that, but it unfortunately wasn't possible at the time.

This PR incorporates some of @twitchbronbron 's suggestions on how to expose errors!  It introduces *a few breaking changes* to the exported interface:

1. The `brs.Lexer` class is now at `brs.lexer.Lexer`
2. `brs.Lexeme` is now at `brs.lexer.Lexeme`
3. `brs.Parser` is now a class in `brs.parser.Parser`
4. `brs.Preprocessor` is now in `brs.preprocessor.Preprocessor`
5. Probably a few more that I've since forgotten

fixes #136


src/lexer/Lexer.ts
```diff
@@ -13,474 +15,504 @@ import {
...
+interface ScanResults {
+    /** The tokens produced by the Lexer. */
+    tokens: ReadonlyArray<Token>,
+    /** The errors encountered by the Lexer. */
+    errors: ReadonlyArray<BrsError>
+}
+
+export class Lexer {
+    /** Allows consumers to observe errors as they're detected. */
+    readonly events = new EventEmitter();
```
TwitchBronBron: Rather than needing to call `myLexer.events.on` where it doesn't tell you what events are exposed, you should consider creating your own "on" method that wraps the events internally, so that consumers of this class get better intellisense as to what actually is acceptable. Same concept applies for the Parser too. Something like this:

```typescript
class Lexer{
    private eventEmitter = new EventEmitter();

    public on(eventName: 'error', callback: (err: BrsError) => void);
    public on(eventName: string, callback: (data: any) => void) {
        this.eventEmitter.on(eventName, callback);
        //create a disconnect function that the caller can use to stop listening to this event
        let disconnect = ()=>{
            this.eventEmitter.removeListener(eventName, callback);
        };
        return disconnect;
    }
}
```

src/Error.ts
```diff
@@ -1,19 +1,11 @@
...
-    foundError = true;
-    return runtime(message, line, file);
-}
-
-export function runtime(message: string, line: number, file?: string) {
-    let location = file ? `${file}: ${line}` : `Line ${line}`;
-    let error = new Error(`[${location}] ${message}`);
-    if (process.env.NODE_ENV !== "test") {
-        console.error(error.message);
+export class BrsError extends Error {
```
TwitchBronBron: Would you be open to including `lineNumber` and `fileName` as properties on BRSError instead of adding them to the message text? That way other tools can use that information to show the error in the right spot without needing to parse the message as a string, and your code that writes the errors to the console can construct the whole message based on the properties.

TwitchBronBron: Also, if you have `columnNumber` (or `beginColumnNumber` and `endColumnNumber`) available when you're creating these errors, could you add that too?

src/parser/Parser.ts
```diff
@@ -24,797 +26,872 @@ type BlockTerminator =
...
+    /** Allows consumers to observe errors as they're detected. */
+    readonly events = new EventEmitter();
 
-    let statements: Statement[] = [];
+    /**
+     * Parses an array of `Token`s into an abstract syntax tree that can be executed with the `Interpreter`.
+     * @param toParse the array of tokens to parse
+     * @returns an array of `Statement` objects that together form the abstract syntax tree of the program
+     */
+    parse(toParse: ReadonlyArray<Token>): ParseResults {
```
TwitchBronBron: You have a static `scan()` convenience method on the lexer, but you didn't include a static `parse()` convenience method for the parser. Was this intentional?

src/parser/Parser.ts
```diff
@@ -24,797 +26,872 @@ type BlockTerminator =
     Lexeme.EndSub |
     Lexeme.EndFunction;
 
-let current: number;
-let tokens: ReadonlyArray<Token>;
+/** The results of a Parser's parsing pass. */
+interface ParseResults {
+    /** The statements produced by the parser. */
+    statements: ReadonlyArray<Stmt.Statement>,
```
TwitchBronBron: It appears that you're actually returning a regular array in the parse method. Making this a `ReadonlyArray` causes me to have to use the same `ReadonlyArray` type in my code, which doesn't seem to provide much benefit. I think it's probably safe enough to just make this an array, and it's my own dumb fault if I screw up my `statements` array.

src/lexer/Lexer.ts
sjbarag: I spent about an hour trying to find a typesafe way to do this, trying to avoid writing custom wrappers. Writing those wrappers would have taken ten minutes. 😐

I’ll add one before I merge (and will likely make `.events` private).

src/parser/Parser.ts
sjbarag: To be honest ReadonlyArray was more about keeping me from screwing things up within the runtime. I’ll look into making the lexer, parser, and friends accept readonly arrays (to keep me from mutating within them)?but return plain-old arrays.

Thanks for the feedback!

sjbarag:
Ignore that circleci failure - I'm still testing that out on a separate branch and can't disable it without changing that last commit :man_shrugging:

PR #141, opened 2019-01-26
test pr please ignore

sjbarag:


sjbarag:
Yeah I think I like circleci - it's really fast for linux builds.  Not sure what to do about windows builds (maybe appveyor or azure pipelines, but azure devops is major overkill right now)  - that's a problem for another day.

PR #186, opened 2019-03-01
fix(lex,parse,e2e): Allow 'then' as a property

### LINKED ISSUE #182: Then should be allowed as function name
If you have something like the following:
```
sendAppConfigNetworkRequest().then(sub(node)
    ....
end sub)
```

You get the message `Expected property name after '.'` as it doesn't know that then is actually the name of the function.

sjbarag:
The `then` keyword isn't _actually_ a keyword, in that it can be used as a valid identifier.  But it's still reserved as far as I can tell?

fixes #182


sjbarag:
(inb4 I find this PR again in 9 weeks and realize that I've done something wrong :man_shrugging:)

PR #189, opened 2019-03-06
feat(parse): Allow assignment operators in set statements

### LINKED ISSUE #173: Assignment operator parse errors for object properties
Assignment operators do not work when used against an object property. Here is a test file that compiles and runs on Roku, but produces parse errors in `brs`.

image

<details><summary>Source</summary>
<p>

```brightscript
function Main()
   m.age = 1

   m.age += 1
   m.age -= 1
   m.age *= 1
   m.age /= 1
   m.age \= 1
   m.age <<= 1
   m.age >>= 1

   m["age"] += 1
   m["age"] -= 1
   m["age"] *= 1
   m["age"] /= 1
   m["age"] \= 1
   m["age"] <<= 1
   m["age"] >>= 1

   print m.age
end function
```

</p>
</details>

sjbarag:
Due to the fact that the `=` token is ambiguous without context (i.e. "is it an assignment or comparison?"), assignments get parsed as binary operations and transformed into assignments after the binary operation's been completely parsed.  Dotted and indexed "set" statements (e.g. `foo.bar = "baz"`) required a bit of additional parsing logic when they were added.  Since assignment operators like `+=` are unambiguously used in statements only, there was yet more parsing logic required to handle assignment operators when the left-hand side is a dotted or indexed set.

This should take care of it though!

fixes #173

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -346,7 +346,7 @@ export class Parser {
             if (match(Lexeme.Return)) { return returnStatement(); }
 
             // TODO: support multi-statements
-            return expressionStatement(...additionalterminators);
+            return setStatement(...additionalterminators);
         }
 
         function whileStatement(): Stmt.While {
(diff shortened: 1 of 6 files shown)


sjbarag:
CC: @triwav since you asked about this on slack the other day :smiley:

PR #198, opened 2019-03-13
Use `name.text` for argument error messages.

### LINKED ISSUE #196: [object Object] appears in error message
`Argument '[object Object]' has no default value, but comes after arguments with default values`

I don't have a test-case for this, but the error message should be enough.  Use `grep` @sjbarag, gosh.

TwitchBronBron:
When the function arguments had their `name` property converted to an object containing location, these spots were missed, causing the error messages to contain `'[object Object]'` instead of the actual argument name

Fixes #196

### DIFF
--- a/src/interpreter/index.ts
+++ b/src/interpreter/index.ts
@@ -767,7 +767,7 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
                 let messageParts = [];
 
                 let args = sig.args.map(a => {
-                    let requiredArg = `${a.name} as ${ValueKind.toString(a.type.kind)}`;
+                    let requiredArg = `${a.name.text} as ${ValueKind.toString(a.type.kind)}`;
                     if (a.defaultValue) {
                         return `[${requiredArg}]`;
                     } else {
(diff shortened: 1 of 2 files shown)


src/interpreter/index.ts
```diff
@@ -767,7 +767,7 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
                 let messageParts = [];
 
                 let args = sig.args.map(a => {
-                    let requiredArg = `${a.name} as ${ValueKind.toString(a.type.kind)}`;
+                    let requiredArg = `${a.name.text} as ${ValueKind.toString(a.type.kind)}`;
```
TwitchBronBron: I couldn't figure out how to test this one, but pretty sure it needed changed as well.

sjbarag:
Thanks for the PR bud! I’m not going to have access to my laptop for a few days but I’ll give this a look on Monday 🙂

src/interpreter/index.ts
sjbarag: IIRC this comes into play when calling something like `UCase(5)` - any type mismatch causes the function signature to be printed out before the mismatched arguments.  But it definitely needed to be fixed.  Great catch!

PR #203, opened 2019-03-18
Parser support type designators

### LINKED ISSUE #162: Add support for type designators
Identifiers are _mostly_ correct, but we're lacking support for type designators – special characters that appear at the end of a variable name to give it a type.  It's sort of like `perl`'s [datatypes](https://perldoc.perl.org/perldata.html) but at the end, and doesn't do anything to the data contained in the variable other than enforce type checks at runtime.

They currently cause parse failures, oh no!

Docs: https://sdkdocs.roku.com/display/sdkdoc/Expressions%2C+Variables%2C+and+Types#Expressions,Variables,andTypes-Identifiers

Thanks for leaving me a TODO without an issue, past-self: https://github.com/sjbarag/brs/blob/8dcb36e9f700489f07e9afa7f35e869680f1fa53/src/lexer/Lexer.ts#L438-L439

TwitchBronBron:
Adds parser support for type designators (i.e. a trailing `$` `%` `#` `!` after a variable identifier). This will still need to be implemented in the runtime at a later time. 

All function, variable, and parameter identifier tokens are permitted to end with a type designator. Then, in the parser, we enforce that function identifiers are NOT allowed to end with a type designator. 

Fixes #162

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -174,6 +174,17 @@ export class Parser {
             } else {
                 name = consume(`Expected ${functionType.text} name after '${functionType.text}'`, Lexeme.Identifier) as Identifier;
                 leftParen = consume(`Expected '(' after ${functionType.text} name`, Lexeme.LeftParen);
+
+                //prevent functions from ending with type designators
+                let lastChar = name.text[name.text.length - 1];
+                if (["$", "%", "!", "#"].indexOf(lastChar) > -1) {
+                    //Since function type is first word, force the first char to upper case to match style of other error messages
+                    let properFunctionName = functionType.text[0].toUpperCase() + functionType.text.substring(1);
+                    //this error should not prevent parsing, but should still show up in the list of errors
+                    try {
+                        addError(new ParseError(name, `${properFunctionName} name cannot end with type designator '${lastChar}'`));
+                    } catch (e) { }
+                }
             }
 
             let args: Argument[] = [];
@@ -244,7 +255,7 @@ export class Parser {
             } else {
                 // only consume trailing newlines in the statement context; expressions
                 // expect to handle their own trailing whitespace
-                while(match(Lexeme.Newline));
+                while (match(Lexeme.Newline));
                 return new Stmt.Function(name!, func);
             }
         }
@@ -373,7 +384,7 @@ export class Parser {
         function exitWhile(): Stmt.ExitWhile {
             let keyword = advance();
             consume("Expected newline after 'exit while'", Lexeme.Newline);
-            while (match(Lexeme.Newline)) {}
+            while (match(Lexeme.Newline)) { }
             return new Stmt.ExitWhile({ exitWhile: keyword });
         }
 
@@ -392,7 +403,7 @@ export class Parser {
                 // BrightScript for/to/step loops default to a step of 1 if no `step` is provided
                 increment = new Expr.Literal(new Int32(1), peek().location);
             }
-            while(match(Lexeme.Newline));
+            while (match(Lexeme.Newline));
 
             let body = block(Lexeme.EndFor, Lexeme.Next);
             if (!body) {
@@ -401,7 +412,7 @@ export class Parser {
                 );
             }
             let endFor = advance();
-            while(match(Lexeme.Newline));
+            while (match(Lexeme.Newline));
 
             // WARNING: BrightScript doesn't delete the loop initial value after a for/to loop! It just
             // stays around in scope with whatever value it was when the loop exited.
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -99,10 +99,17 @@ export class Parser {
 
         let errors: ParseError[] = [];
 
-        const addError = (err: ParseError) => {
+        /**
+         * Add an error to the parse results. 
+         * @param token - the token where the error occurred
+         * @param message - the message for this error
+         * @returns an error object that can be thrown if the calling code needs to abort parsing
+         */
+        const addError = (token:Token, message:string) => {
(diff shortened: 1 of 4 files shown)


src/parser/Parser.ts
```diff
@@ -174,6 +174,17 @@ export class Parser {
             } else {
                 name = consume(`Expected ${functionType.text} name after '${functionType.text}'`, Lexeme.Identifier) as Identifier;
                 leftParen = consume(`Expected '(' after ${functionType.text} name`, Lexeme.LeftParen);
+
+                //prevent functions from ending with type designators
+                let lastChar = name.text[name.text.length - 1];
+                if (["$", "%", "!", "#"].indexOf(lastChar) > -1) {
+                    //Since function type is first word, force the first char to upper case to match style of other error messages
+                    let properFunctionName = functionType.text[0].toUpperCase() + functionType.text.substring(1);
```
sjbarag: Would you mind reformatting the error message so the function name isn't the first word?  Something like `Function named  '${name.text}' cannot end with type designator '${lastChar}'` would let us keep the function name capitalized the way it's written in the source text &mdash; even if it's `rEaLlYsilLy$()`.  Then we wouldn't need this line!

src/parser/Parser.ts
```diff
@@ -401,7 +412,7 @@ export class Parser {
                 );
             }
             let endFor = advance();
-            while(match(Lexeme.Newline));
+            while (match(Lexeme.Newline));
```
sjbarag: One of these days I'll set up `tsfmt` or something like that.

test/parser/statement/Function.test.js
```diff
@@ -178,6 +178,69 @@ describe("parser", () => {
...
+                    return 1
+                end function
+
+                function DoubleFunc#()
+                    return 1
+                end function
+            `);
+            const { statements, errors } = parser.parse(tokens);
+            expect(errors.length).toEqual(4);
+            expect(errors[0].location).toMatchObject({
```
sjbarag: Rather than building these locations manually, consider using `expect(errors).toMatchSnapshot()`!  Please correct me if I'm wrong, but it seems like the spirit of these tests is to make sure four errors are produced, not necessarily the locations themselves.  Since we probably want the tests to fail if the messages change too, using a snapshot might save some space and make future updates to these tests easier!

src/parser/Parser.ts
```diff
@@ -174,6 +174,17 @@ export class Parser {
...
+
+                //prevent functions from ending with type designators
+                let lastChar = name.text[name.text.length - 1];
+                if (["$", "%", "!", "#"].indexOf(lastChar) > -1) {
+                    //Since function type is first word, force the first char to upper case to match style of other error messages
+                    let properFunctionName = functionType.text[0].toUpperCase() + functionType.text.substring(1);
+                    //this error should not prevent parsing, but should still show up in the list of errors
+                    try {
+                        addError(new ParseError(name, `${properFunctionName} name cannot end with type designator '${lastChar}'`));
+                    } catch (e) { }
```
sjbarag: Oh dang, I think this is the first place where we're emitting an error but still producing an AST node.  Looks really neat!  I'll try to think of a better way to handle this case than a `try/catch` around `addError`.  Or perhaps it just doesn't throw anymore?  We'll see!

src/parser/Parser.ts
TwitchBronBron: Dang, I missed that one. I'm a very aggressive auto-formatter, so the last step in my commits is to undo all the auto-formats. lol Do you want me to undo this?

sjbarag: Eh, doesn't bother me much either way right now.  It's a sensible change so I don't mind leaving it!  It was more a note to myself :man_shrugging:

test/parser/statement/Function.test.js
TwitchBronBron: So I _did_ want to verify the errors were in the proper locations (i.e. are these the errors I was actually looking for). However, since my initial test passed, I can assume the snapshot is also valid, so I think you're right, just use another snapshot.

src/parser/Parser.ts
TwitchBronBron: Done.

TwitchBronBron: Although, not sure if you noticed, this wasn't the function name. it was the function TYPE (literally the words `Sub` or `Function`)

src/parser/Parser.ts
TwitchBronBron: Ok, so this comment got me thinking about a way to remove the try catch. I just pushed some new code where I removed the throw inside of `addError`, and now, the code _calling_ `addError` is responsible for throwing. 

I also moved the creation of `ParserError` into of `addError`. I think this cuts down on the verbosity of throwing errors because you don't need to new up a new `ParseError` every time.

These changes help readability because an explicit `throw` shows that the code will short-circuit at the call site of `addError`. 

Let me know what you think. I can back it out if you don't like it.

TwitchBronBron:
I think I made all the changes discussed, and I tweaked the `addError` concept a little bit. ready for your review again.

src/parser/Parser.ts
sjbarag: Oh heck, I got thrown off by the `properFunctionName` variable, thinking it was the function's name but formatted as a proper noun!  Sorry about that!

src/parser/Parser.ts
sjbarag: Thanks for taking a stab at that - I definitely like the explicit `throw` you've added!  At some level I liked `new`ing a `ParseError` for that same level of explicitness (explicitacity?), but since every one wound up being a `ParseError` it seems better this way.  Good call!

PR #206, opened 2019-03-18
Add better parser recovery for mismatch end function and end sub

### LINKED ISSUE #205: Add better parser recovery for mismatch end function and end sub
It's fairly common to change a sub to a function, and vice versa. Sometimes you forget to change the end keyword. When this happens, the parser doesn't recover very well, and instead consumes the rest of the tokens without parsing them, and adds an error to the end of the file. Like this:
image

For this specific scenario, we should allow any function or sub to terminate with the opposite keyword, and throw an error after the function/sub is constructed, thus allow the rest of the program to be parsed successfully.

TwitchBronBron:
This PR enables functions and subs to be terminated by both `end sub` and `end function`. A parse error is still added, but this enables the parser to recover and continue with the rest of the file.

Fixes #205


test/parser/statement/Function.test.js
```diff
@@ -12,6 +12,33 @@ describe("parser", () => {
...
+                    print "Hello world"
+                end sub
+                
+                sub DoSomething()
+                
+                end sub 
+            `);
+            let { statements, errors } = parser.parse(tokens);
+            expect(errors.length).toBe(1);
+            expect(errors[0].location).toMatchObject({
```
sjbarag: Like in #203, consider using a second snapshot here!  The location seems slightly less important than the error's message, and we can check both at the same time with a second snapshot :camera_flash:

TwitchBronBron:
Resolved. Ready for another review.

test/parser/statement/Function.test.js
```diff
@@ -181,6 +198,23 @@ describe("parser", () => {
...
+                
+                sub DoSomething()
+                
+                end sub 
+            `);
+            let { statements, errors } = parser.parse(tokens);
+            expect(errors.length).toBe(1);
+            expect(statements).toBeDefined();
+            expect(statements).not.toBeNull();
+            expect({ errors, statements }).toMatchSnapshot();
```
sjbarag: I don't know why I've never thought to package these up this way - I'd looked into using multiple snapshots (which ostensibly works?) but this is super nice :heart:

sjbarag:
Oh heck, I think merging your other PRs caused a merge conflict.  I'd resolve it, but the last time I made a commit in someone else's PR and did a squash-and-merge, Github decided to give me all the credit :cry: I don't want that to happen here - would you mind resolving the conflict when you get a minute?

TwitchBronBron:
Ok, I merged...ready for your review again.

PR #207, opened 2019-03-18
Fix for each space

### LINKED ISSUE #201: more than 1 space between else if causes parse error
Having extra whitespace between the `else` and `if` causes parse errors, but runs successfully on a Roku.
image

### LINKED ISSUE #202: more than 1 space between for each causes parse error
If there is more than one whitespace character between the `for` and `each`, that results in a parser error. However, this successfully compiles and runs on Roku.

image

TwitchBronBron:
Fix issues in the parser with multiple spaces between two-word keywords (like `for each`, `exit while`, etc). 

I also verified that tabs are treated the same as spaces, so I added support for tabs between these keywords as well. This program is an example (replace `\t\t` with actual tab chars):

```
sub SayHi(message as string)
    for \t\t each num in [1,2,3]
		print num
	end for
end sub
```

Fixes #202 
Fixes #201

### DIFF
--- a/src/lexer/Lexer.ts
+++ b/src/lexer/Lexer.ts
@@ -486,13 +486,21 @@ export class Lexer {
             let text = source.slice(start, current).toLowerCase();
 
             // some identifiers can be split into two words, so check the "next" word and see what we get
-            if ((text === "end" || text === "else" || text === "exit" || text === "for") && peek() === " ") {
+            if ((text === "end" || text === "else" || text === "exit" || text === "for") && (peek() === " " || peek() === "\t")) {
                 let endOfFirstWord = current;
 
-                advance(); // skip past the space
+                // skip past any whitespace
+                let whitespace = "";
+                while (peek() === " " || peek() === "\t") {
+                    //keep the whitespace so we can replace it later
+                    whitespace += peek();
+                    advance();
+                }
                 while (isAlphaNumeric(peek())) { advance(); } // read the next word
 
                 let twoWords = source.slice(start, current);
+                //replace all of the whitespace with a single space character so we can properly match keyword token types
+                twoWords = twoWords.replace(whitespace, " ");
                 let maybeTokenType = KeyWords[twoWords.toLowerCase()];
                 if (maybeTokenType) {
                     addToken(maybeTokenType);
--- a/test/lexer/Lexer.test.js
+++ b/test/lexer/Lexer.test.js
@@ -342,8 +342,8 @@ describe("lexer", () => {
         });
     });
 
-    describe('for each loop', () => {
-        it('supports various types of spacing between the two keywords', () => {
+    describe('two word keywords', () => {
+        it('supports various spacing between for each', () => {
             var k = 2;
             let { tokens } = Lexer.scan('for each for  each for    each for\teach for\t each for \teach for \t each');
             console.log(tokens);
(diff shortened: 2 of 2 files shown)


test/lexer/Lexer.test.js
```diff
@@ -341,4 +341,38 @@ describe("lexer", () => {
             ]);
         });
     });
+
+    describe('two word keywords', () => {
+        it('supports various spacing between for each', () => {
+            var k = 2;
+            let { tokens } = Lexer.scan('for each for  each for    each for\teach for\t each for \teach for \t each');
+            console.log(tokens);
```
sjbarag: We probably don't need this anymore :wink:

test/lexer/Lexer.test.js
```diff
@@ -341,4 +341,38 @@ describe("lexer", () => {
             ]);
         });
     });
+
+    describe('two word keywords', () => {
+        it('supports various spacing between for each', () => {
+            var k = 2;
```
sjbarag: Doesn't look like this (or the other `var k = 2` on line 362) are being used.  Would you mind removing them?

src/lexer/Lexer.ts
```diff
@@ -486,13 +486,21 @@ export class Lexer {
...
+                while (peek() === " " || peek() === "\t") {
+                    //keep the whitespace so we can replace it later
+                    whitespace += peek();
+                    advance();
+                }
                 while (isAlphaNumeric(peek())) { advance(); } // read the next word
 
                 let twoWords = source.slice(start, current);
+                //replace all of the whitespace with a single space character so we can properly match keyword token types
+                twoWords = twoWords.replace(whitespace, " ");
```
sjbarag: Oh interesting!  I hadn't thought of doing it this way.  Neat idea!

test/lexer/Lexer.test.js
TwitchBronBron: oops. Yeah, my bad. The typescript debugger doesn't always break on the line when you have `let {tokens}`, so I had to break the line before.

src/lexer/Lexer.ts
TwitchBronBron: Thanks. I considered using a regular expression replacement....but you haven't used regex anywhere else in the project, and they are slow relative to while loops, so I figured this would be ever so slightly faster. When opening a project with 200-300 files, that might save some milliseconds!

test/lexer/Lexer.test.js
sjbarag: No worries, friend!

test/lexer/Lexer.test.js
TwitchBronBron: removed!

test/lexer/Lexer.test.js
TwitchBronBron: Removed!

TwitchBronBron:
Should be good now.

test/lexer/Lexer.test.js
sjbarag: Hmm, perhaps my browser's borked - I control-shift-R'd and it's still there! :thinking:

TwitchBronBron: Yeah, it might be messed up. Switch to the "Files changed" tab and see if the line still exists over there. I don't see it over there...but the comment above is definitely still showing it.

TwitchBronBron: Oh, I didn't remove the console.log. Der.

TwitchBronBron:
I believe I resolved everything. Ready for another review.

PR #211, opened 2019-03-19
Parser support support all keywords as property names

### LINKED ISSUE #194: property named "next" causes parse errors
There is a parse error when using a property called `next` on an object. I just ran this on a roku, and it works.

image

### LINKED ISSUE #208: `end` causes parse errors when used as a property name
The word `end` can be used as an object property name. However, `brs` runs into parse errors in this case. This is another one of those "it's a keyword _sometimes_" issues.

image

### LINKED ISSUE #226: Allow exit as AA key
using `exit` as a key in AA like
```
this = {
     exit: sub()
         ...
    end sub
}
```
or 
`this.exit = ...`
results in `Expected identifier or string as associative array key, but received 'exit'`. Adding quotes works around this but this is valid syntax so shouldn't throw an error.

TwitchBronBron:
The parser doesn't currently support the `end` word as property of an object. This PR fixes that.

This PR also adds back in the ability to throw parse errors when using keywords as identifiers in places that are not allowed. For example, the `end` keyword is not allowed as a standalone variable, but it IS allowed as a property name. So, we should see an error when used as a block variable. 

Since I deleted one of the token types, a lot of the snaps have shifted their `kind` values down by one. 

fixes #208
fixes #194  
fixes #226


src/parser/Statement.ts
```diff
@@ -257,6 +257,27 @@ export class Return implements Statement {
...
 
+export class End implements Statement {
+    constructor(
+        readonly tokens: {
+            end: Token
+        }
+    ) { }
+
+    accept<R>(visitor: Visitor<R>): BrsType {
+        //TODO implement this in the runtime. It should immediately terminate program execution, without error
```
TwitchBronBron: @sjbarag you probably want to implement this at some point...

src/lexer/ReservedWords.ts
```diff
@@ -98,3 +97,10 @@ export const KeyWords: {[key: string]: L} = {
     true: L.True,
     while: L.While,
 };
+
+/**
+ * These reserved words can be used as object properties (but NOT as standalone vars)
+ */
+export const ReservedWordsUsableAsProperty = new Set([
```
TwitchBronBron: This is where I would expect us to add all of the other "keyword-ish" keywords. If you add them here, they will automatically cause recoverable parse errors. So we should add all of those keywords in here at some point. I can do that now if you know the list offhand, or we can just pus it off for another day.

test/parser/statement/Misc.test.js
```diff
@@ -0,0 +1,52 @@
...
+                sub Main()
+                    end = true
+                end sub
+            `);
+            let { statements, errors } = parser.parse(tokens);
+            expect(errors.length).toEqual(1);
+            expect(errors[0].location.start.line).toEqual(3)
+            //TODO - re-enable this test once the `end` identifier location bug is fixed
+            // //specifically check for the error location, because the identifier location was wrong in the past
+            // expect(errors[0].location).toEqual({
```
TwitchBronBron: My change introduced an issue in the lexer that is incorrectly computing the position of the `end` token with my new change. It's off by 1 character (i.e. says char 8 but should be char 7). I tried to figure it out, but couldn't. 

The error is close enough to the correct position, and subsequent tokens get the proper location, so I don't think this is a show stopper. However, it would be great if you could take a look at this once the PR gets merged, and let me know what is wrong.

sjbarag: Interesting!  I can't seem to reproduce that with simple files:

```brightscript
sub main()
    end
end sub
```

Produces this token for `"end"`:

```json
{
  "kind": 28,
  "text": "end",
  "isReserved": false,
  "location": {
    "start": {
      "line": 2,
      "column": 4
    },
    "end": {
      "line": 2,
      "column": 7
    },
    "file": "/path/to/the/file.brs"
  }
}
```

Looks okay to me?

sjbarag: @TwitchBronBron do you happen to have a file that'd repro the off-by-one you mentioned above?  I'd like to get a feel for what kinds of situations are affected before I merge this.

test/parser/statement/Misc.test.js
```diff
@@ -0,0 +1,52 @@
...
+            let { tokens } = brs.lexer.Lexer.scan(`
+                sub Main()
+                    end = true
+                end sub
+            `);
+            let { statements, errors } = parser.parse(tokens);
+            expect(errors.length).toEqual(1);
+            expect(errors[0].location.start.line).toEqual(3)
+            //TODO - re-enable this test once the `end` identifier location bug is fixed
+            // //specifically check for the error location, because the identifier location was wrong in the past
```
TwitchBronBron: @sjbarag uncomment this code to see the off-by-one position error.

test/parser/statement/Misc.test.js
TwitchBronBron: The error shows up when you use an assignment with end. Like this: 
```'master branch
sub main()
   end = 12
end sub
```
image

I added a comment in this PR to a commented out test that shows the issue.

sjbarag: Finally got some time to figure this out!  Turns out there was a small bug in column tracking involving the `end` string: we attempt to read the next word to see if it's a multiword `end if`, `end while` etc. and reset the value of `current` (the index into the string representation of the file we're lexing), but not the value of `column` (the index into the specific line in that file).


This should fix things:

```diff
diff --git a/src/lexer/Lexer.ts b/src/lexer/Lexer.ts
index 5799f81..817ccb7 100644
--- a/src/lexer/Lexer.ts
+++ b/src/lexer/Lexer.ts
@@ -503,7 +503,10 @@ export class Lexer {
 
             // some identifiers can be split into two words, so check the "next" word and see what we get
             if ((text === "end" || text === "else" || text === "exit" || text === "for") && (peek() === " " || peek() === "\t")) {
-                let endOfFirstWord = current;
+                let endOfFirstWord = {
+                    position: current,
+                    column: column
+                };
 
                 // skip past any whitespace
                 let whitespace = "";
@@ -523,7 +526,8 @@ export class Lexer {
                     return;
                 } else {
                     // reset if the last word and the current word didn't form a multi-word Lexeme
-                    current = endOfFirstWord;
+                    current = endOfFirstWord.position;
+                    column = endOfFirstWord.column;
                 }
             }
```

TwitchBronBron: Thanks. That did the trick!

TwitchBronBron:
@sjbarag  I just pushed the changes to fix the lexer off-by-one issue, and merged from master so everything is up-to-date. Ready for your review.

sjbarag:
Hiya!  I think I've found a much simpler way to handle this that'll require no changes to the Lexer at all :smile:  Basically, we can keep a list of Lexemes that are valid as property names, then check for those in addition to standard Identifiers while parsing!  The diff wound up being very small:

```diff
diff --git a/src/lexer/ReservedWords.ts b/src/lexer/ReservedWords.ts
index 32dfbdf..16fe86f 100644
--- a/src/lexer/ReservedWords.ts
+++ b/src/lexer/ReservedWords.ts
@@ -49,6 +49,13 @@ export const ReservedWords = new Set([
     "while"
 ]);
 
+export const AllowedProperties = [
+    L.And,
+    L.Or,
+    L.Next,
+    L.End
+];
+
 /**
  * The set of keywords in the refrence BrightScript runtime. Any of these that *are not* reserved
  * words can be used within a BrightScript file for other purposes, e.g. `tab`.
diff --git a/src/parser/Parser.ts b/src/parser/Parser.ts
index fdb33ca..62da310 100644
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -4,7 +4,7 @@ import * as Expr from "./Expression";
 type Expression = Expr.Expression;
 import * as Stmt from "./Statement";
 type Statement = Stmt.Statement;
-import { Lexeme, Token, Identifier, ReservedWords } from "../lexer";
+import { Lexeme, Token, Identifier, ReservedWords, AllowedProperties } from "../lexer";
 import { ParseError } from "./ParseError";
 
 import {
@@ -857,9 +857,12 @@ export class Parser {
                 } else if (match(Lexeme.Dot)) {
                     while (match(Lexeme.Newline));
 
-                    let name = consume("Expected property name after '.'", Lexeme.Identifier) as Identifier;
+                    let name = consume("Expected property name after '.'", Lexeme.Identifier, ...AllowedProperties);
 
-                    expr = new Expr.DottedGet(expr, name);
+                    // force it into an identifier so the AST makes some sense
+                    name.kind = Lexeme.Identifier;
+
+                    expr = new Expr.DottedGet(expr, name as Identifier);
                 } else {
                     break;
                 }
@@ -945,7 +948,7 @@ export class Parser {
 
                     function key() {
                         let k;
-                        if (check(Lexeme.Identifier)) {
+                        if (check(Lexeme.Identifier, ...AllowedProperties)) {
                             k = new BrsString(advance().text!);
                         } else if (check(Lexeme.String)) {
                             k = advance().literal! as BrsString;
```

and works on this simple test file:

```brightscript
sub main()
    aa = {
        previous: "http://example.com/1",
        next: "http://example.com/3",
        end: "http://example.com/30"
    }

    print "Previous: " aa.previous
    print "Next: " aa.next
    print "End: " aa.end
end sub
```

This does run the risk of making `consume` a bit slower, since we have to compare against a potentially large array, but it wouldn't be too difficult to build a `Set` out of those Lexemes within `consume`.  That seems like a problem for future-me :wink: 

@TwitchBronBron what do you think?  Am I missing anything (besides a whole bunch of lexemes that are valid property names)?

sjbarag:
:thinking: Now that I think about it, that `AllowedProperties` array is only used in the parser anyway so it can probably live in `src/parser/` somewhere!

src/lexer/Lexer.ts
```diff
@@ -511,7 +511,10 @@ export class Lexer {
 
             // some identifiers can be split into two words, so check the "next" word and see what we get
             if ((text === "end" || text === "else" || text === "exit" || text === "for") && (peek() === " " || peek() === "\t")) {
-                let endOfFirstWord = current;
+                let endOfFirstWord = {
```
TwitchBronBron: This section is where we fixed the off-by-one location issue with two-word tokens

src/parser/Parser.ts
```diff
@@ -148,7 +263,7 @@ export class Parser {
                 // BrightScript is like python, in that variables can be declared without a `var`,
                 // `let`, (...) keyword. As such, we must check the token *after* an identifier to figure
                 // out what to do with it.
-                if ( check(Lexeme.Identifier) && checkNext(...assignmentOperators)) {
+                if (check(Lexeme.Identifier, ...allowedIdentifiers) && checkNext(...assignmentOperators)) {
```
TwitchBronBron: this allows certain keywords to be used as identifiers.

src/parser/Parser.ts
```diff
@@ -311,6 +426,11 @@ export class Parser {
 
         function assignment(...additionalterminators: Lexeme[]): Stmt.Assignment {
             let name = advance() as Identifier;
+            //add error if name is a reserved word that cannot be used as an identifier
+            if (disallowedIdentifiers.indexOf(name.text.toLowerCase()) > -1) {
```
TwitchBronBron: throw an error anytime a disallowed reserved word is used as an identifier.

src/parser/Parser.ts
```diff
@@ -945,7 +1082,7 @@ export class Parser {
 
                     function key() {
                         let k;
-                        if (check(Lexeme.Identifier)) {
+                        if (check(Lexeme.Identifier, ...allowedProperties)) {
```
TwitchBronBron: This is the new section that you suggested for supporting many reserved words as property names

test/parser/expression/AssociativeArrayLiterals.test.js
```diff
@@ -181,7 +181,7 @@ describe("parser associative array literals", () => {
             token(Lexeme.Integer, "2", new Int32(2)),
             token(Lexeme.Comma, ","),
             token(Lexeme.Newline, "\n"),
-            token(Lexeme.String, "requires-hyphens", new BrsString("requires-hypens")),
+            token(Lexeme.String, "requires-hyphens", new BrsString("requires-hyphens")),
```
TwitchBronBron: Fix typo...

PR #214, opened 2019-03-22
Better single line recovery

### LINKED ISSUE #157: Better recovery after errors
If there is invalid syntax above a for loop, the `end for` ends up with a parser error. Is there a way to improve the way the parser recovers so that the error is contained to the `1=1` line?

```brightscript
function Main1()
  1=1
  for each num in [1,2,3]
    print num
  end for
end function
```

image

TwitchBronBron:
Sometimes a parse error causes the `block` function to consume all of the remaining tokens. This PR fixes that issue by discarding the line, which should drastically improve parse error recovery.

Fixes #157

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -706,14 +706,20 @@ export class Parser {
 
             const statements: Statement[] = [];
             while (!check(...terminators) && !isAtEnd()) {
-                const dec = declaration();
+                //grab the location of the current token
+                let loopCurrent = current;
+                let dec = declaration();
+
                 if (dec) {
                     statements.push(dec);
                 } else {
-                    // //couldn't find a declaration. Try discarding the line entirely for now
-                    // while (!check(Lexeme.Newline) && !isAtEnd()) {
-                    //     advance();
-                    // }
+                    //something went wrong. reset to the top of the loop
+                    current = loopCurrent;
+
+                    //scrap the entire line
+                    consumeUntil(Lexeme.Colon, Lexeme.Newline, Lexeme.Eof);
+                    //trash the newline character so we start the next iteraion on the next line
+                    advance();
                 }
             }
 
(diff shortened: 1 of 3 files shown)


sjbarag:
Heya!  Looks like this reports a few more errors than expected.  Consider this test file:

```brightscript
sub main()
    aa = { foo: "bar",  baz: "qux" +}
    derp()
end sub

sub derp()
    print "hi!"
end sub
```

With your changes, three errors are reported:

```
$ brs error.brs
error.brs(2,36-37): Found unexpected token '}'
error.brs(2,21-22): Expected newline or ':' after expression statement
error.brs(2,35-36): Expected newline or ':' after expression statement
```

But in current-master, only the first is reported:

```
$ brs error.brs
error.brs(2,36-37): Found unexpected token '}'
```

That seems like a pretty big regression - anything we can do to reduce the extraneous errors?

src/parser/Parser.ts
```diff
@@ -706,9 +706,20 @@ export class Parser {
...
-                const dec = declaration();
+                //grab the location of the current token
+                let loopCurrent = current;
+                let dec = declaration();
+
                 if (dec) {
                     statements.push(dec);
+                } else {
+                    //something went wrong. reset to the top of the loop
+                    current = loopCurrent;
```
sjbarag: Looks like this is part of the issue - when we encounter an error, it appears we rewind to the start of the line and only consume until the first `:`, causing a partial associative array to remain.

TwitchBronBron:
In typescript, one invalid character can cause a bunch of bogus error messages too. I created a test typescript file and added the percent sign to the end of the `name` property of an object. While that's technically the only issue, it DOES cause a ripple effect of other parse errors. 

image

With respect to this PR, I wouldn't consider it a regression. Nothing stops working, and we still get the error we used to get. We just end up with a few extra errors as well. The current master branch, in many cases, scraps either the rest of the block or sometimes the rest of the file. I feel like this PR moves us in the right direction, and these extra errors are worth that forward progress. 

For example, using the master branch code, typing a not-fully-formed function name causes a similar set of errors to the ones you mention above, but in this case, line 2 is VALID but gets marked as invalid.
image


However, with this branch, we are now able to recover from that by discarding the `der` line
image

I know we have more work to do when handling parse errors inside of array and associative array literals, but I think this PR is still a step forward. 

Thoughts?

sjbarag:
Thanks for the writeup!  It seems Ruby does something similar: https://repl.it/repls/VariableSilentGravity

You're right - this is definitely an improvement over the status quo!

test/parser/expression/Call.test.js
```diff
@@ -25,10 +25,45 @@ describe("parser call expressions", () => {
...
+        let lineNumbers = errors.map(x => x.location.start.line);
+        for (let lineNumber of lineNumbers) {
+            expect(lineNumber).toEqual(3);
+        }
+        expect(statements).toMatchSnapshot();
+    });
+
+    it('does not invalidate the next statement on a multi-statement line', () => {
+        const { tokens } = brs.lexer.Lexer.scan(`
+            sub DoThingOne()
```
sjbarag: Like in #215, I typically try to avoid making these tests depend on the lexer's implementation.  I think I'm okay with leaving this as-is because of how verbose this test would be otherwise (since locations _are_ in fact important here).

PR #215, opened 2019-03-22
Support colons as separators for AA properties

### LINKED ISSUE #195: Colon as separator between anonymous object properties
Aparently brightscript allows you to use a colon to separate properties in objects. This is technically valid syntax. I tested each of these against a live roku, and they all printed the values out properly: 

image

So is this (colon and then newline): 
image

TwitchBronBron:
The parser does not properly parse objects with properties separate by colons: `obj = {x:0 : y: 1}`. This PR resolves that issue.

Fixes #195

### DIFF
--- a/test/parser/expression/AssociativeArrayLiterals.test.js
+++ b/test/parser/expression/AssociativeArrayLiterals.test.js
@@ -135,6 +135,17 @@ describe("parser associative array literals", () => {
         });
     });
 
+    it('allows separating properties with colons', ()=>{
+        const { tokens } = brs.lexer.Lexer.scan(`
+            sub Main()
+                person = {name: "Bob":::::age:50}
+            end sub 
+        `);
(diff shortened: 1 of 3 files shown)


test/parser/expression/AssociativeArrayLiterals.test.js
```diff
@@ -135,6 +135,17 @@ describe("parser associative array literals", () => {
         });
     });
 
+    it('allows separating properties with colons', ()=>{
+        const { tokens } = brs.lexer.Lexer.scan(`
```
sjbarag: In cases where location doesn't matter, please use the `identifier` and `token` convenience functions (see the other tests in this file for examples) to keep these tests isolated from the lexer's output.

TwitchBronBron: Sure. I updated the test to use an array of tokens like you asked, so you should be good to

PR #218, opened 2019-03-22
Tslint fixes

### LINKED ISSUE #209: Enforce code formatting lint rules for entire project
After doing a few PRs for this project, I've noticed there are lots of little differences in formatting between the vscode auto-format and what is stored in this project. It would be nice to pick a standard, and enforce this with CI on every PR (through tslint). 

Getting tslint to enforce spacing is a bit quirky sometimes, but I found a few rules from npm that make this do-able in my `brightscript-language` [tslint.json file](https://github.com/TwitchBronBron/brightscript-language/blob/9c4a6fdf36451300152b972f7cc816ce5f006e4d/tslint.json)

I'd be more than happy to put together a PR with the lint enforcements, and also with all the tslint changes if you want?

TwitchBronBron:
Did a full project formatting sweep, and updated tslint rules to enforce those new formatting settings. 

Fixes #209


TwitchBronBron:
@sjbarag I just merged from master, so there are no longer conflicts.

TwitchBronBron:
I'm not a fan of prettier. It's _too_ opinionated. I dislike several of its style choices that they don't have options to fix. However, it _does_ solve my desire for code consistency. Also, after looking through your branch, there weren't many formatting changes that I felt strong enough about to make an argument against using prettier. 

My biggest concern is the use of the git commit hook. I'm not a fan of having things just magically happen behind the scenes...as I find this causes problems at the most inopportune times. Like what happens when the commit fails? Or there's a merge conflict. Now my local code could potentially have more changes that I didn't anticipate, causing confusion to anyone who doesn't explicitly know you hacked into the commit hooks. 

Here's how I would handle it if this were my project:

1. Add CI check to enforce styles on PR. 
2. Add npm task that allows developers to auto-format the entire project (allowing THEM to decide the best time to run the formatting). 
3. Add a line in the "How to contribute to this project" section of the readme, saying "Run the format task before submitting your pull requests. Unformatted code will be rejected". 

Then, let developers decide to run the formatting task on their own. If they don't, CI will catch it and you can yell at them. If they do formatting on their own, everything runs smoothly. 

Thoughts?

sjbarag:
I've spent some time on other projects that have [prettier](https://prettier.io/) configured for automatic formatting and have fallen in love with it (at least when configured with a git commit hook to reformat committed files).  I'm gonna resolve this in favor of #243 , but thanks for opening it in the first place!  It drove me to look for a better way than what I had before :1st_place_medal:

PR #219, opened 2019-03-26
feat(stdlib): Implement createObject("roRegex")

alimnios72:
This feature implements all methods of the [ifRegex](https://sdkdocs.roku.com/display/sdkdoc/ifRegex) interface. 
Although javascript's `RegExp` doesn't fully support PCRE, as Brightscript does, it gives `brs` the ability to support most of the common regex use cases.

All flags in [roRegex](https://sdkdocs.roku.com/display/sdkdoc/roRegex) but "x" should be supported.

### DIFF
--- a/src/brsTypes/components/Regex.ts
+++ b/src/brsTypes/components/Regex.ts
@@ -108,11 +108,15 @@ export class Regex extends BrsComponent implements BrsValue {
         "split",
         {
             signature: {
-                args: [],
-                returns: ValueKind.Boolean
+                args: [
+                    new StdlibArgument("str", ValueKind.String)
+                ],
+                returns: ValueKind.Object
             },
             impl: (interpreter: Interpreter, str: BrsString) => {
-                return BrsBoolean.False;
+                let items = this.jsRegex[Symbol.split](str.value);
+                let brsItems = items.map( item => new BrsString(item));
+                return new BrsArray(brsItems);
             }
         }
     );
--- a/src/brsTypes/components/Regex.ts
+++ b/src/brsTypes/components/Regex.ts
@@ -1,8 +1,9 @@
-import { BrsValue, ValueKind, BrsString, BrsBoolean } from "../BrsType";
+import { BrsBoolean, BrsInvalid, BrsString, BrsValue, ValueKind } from "../BrsType";
 import { BrsComponent } from "./BrsComponent";
 import { BrsType } from "..";
 import { Callable, StdlibArgument } from "../Callable";
 import { Interpreter } from "../../interpreter";
+import { BrsArray } from "./BrsArray";
 
 export class Regex extends BrsComponent implements BrsValue {
     readonly kind = ValueKind.Object;
@@ -31,7 +32,7 @@ export class Regex extends BrsComponent implements BrsValue {
         {
             signature: {
                 args: [
-                    new StdlibArgument("str", ValueKind.String),
+                    new StdlibArgument("str", ValueKind.String)
                 ],
                 returns: ValueKind.Boolean
             },
@@ -45,11 +46,22 @@ export class Regex extends BrsComponent implements BrsValue {
         "match",
         {
             signature: {
-                args: [],
-                returns: ValueKind.Boolean
+                args: [
+                    new StdlibArgument("str", ValueKind.String)
+                ],
+                returns: ValueKind.Object
             },
             impl: (interpreter: Interpreter, str: BrsString) => {
-                return BrsBoolean.False;
+                const result = this.jsRegex.exec(str.value);
+                let arr = [];
+                if (result !== null) {
+                    for (let i = 0; i < result.length; i++) {
+                        let item = result[i] ? new BrsString(result[i]) : BrsInvalid.Instance;
+                        arr.push(item);
+                    }
+                }
+
+                return new BrsArray(arr);
             }
         }
     );
--- a/src/brsTypes/components/Regex.ts
+++ b/src/brsTypes/components/Regex.ts
@@ -7,11 +7,12 @@ import { BrsArray } from "./BrsArray";
 
 export class Regex extends BrsComponent implements BrsValue {
     readonly kind = ValueKind.Object;
+    readonly supportedFlags = "ims";
     private jsRegex : RegExp;
 
     constructor(expression: BrsString, flags = new BrsString("")) {
         super("roRegex");
-        this.jsRegex = new RegExp(expression.value, flags.value);
+        this.jsRegex = new RegExp(expression.value, this.parseFlags(flags.value));
 
         this.registerMethods([
             this.isMatch,
@@ -25,6 +26,25 @@ export class Regex extends BrsComponent implements BrsValue {
 
     toString(parent?: BrsType): string {
         return "<Component: roRegex>";
+    }
+
+    private parseFlags(inputFlags: string): string {
+        let parsedFlags = "";
+        if (inputFlags.length === 0) {
+            return parsedFlags;
+        }
+
+        for (const flag of inputFlags) {
+            if (flag === "x") {
+                console.warn("'x' flag is not implemented yet, ignoring flag.");
+            } else if(this.supportedFlags.indexOf(flag) === -1) {
+                throw new Error(`${flag} is not supported.`);
+            } else {
+                parsedFlags += flag;
+            }
+        }
+
+        return parsedFlags;
     }
 
     private parseReplacementPattern(pattern: string): string {
--- a/src/brsTypes/components/Regex.ts
+++ b/src/brsTypes/components/Regex.ts
@@ -27,6 +27,10 @@ export class Regex extends BrsComponent implements BrsValue {
         return "<Component: roRegex>";
     }
 
+    private parseReplacementPattern(pattern: string): string {
+        return pattern.replace(/\\/g, "\$");
+    }
+
     private isMatch = new Callable(
         "ismatch",
         {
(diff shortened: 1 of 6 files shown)


src/brsTypes/components/Regex.ts
```diff
@@ -0,0 +1,153 @@
...
+            signature: {
+                args: [
+                    new StdlibArgument("str", ValueKind.String),
+                    new StdlibArgument("str", ValueKind.String)
+                ],
+                returns: ValueKind.String
+            },
+            impl: (interpreter: Interpreter, str: BrsString, replacement: BrsString) => {
+                const source = this.jsRegex.source;
+                const flags = this.jsRegex.flags + "g";
```
strattonbrazil: If someone passes "g" into roRegex I think a second one will be applied here, which will raise a SyntaxError.  

    new RegExp("some thing", "gg"); // throws error

I'm not sure how Roku's implementation, but in both the constructor above and here we might want to check for duplicates.

src/brsTypes/components/Regex.ts
```diff
@@ -0,0 +1,153 @@
...
+        {
+            signature: {
+                args: [
+                    new StdlibArgument("str", ValueKind.String)
+                ],
+                returns: ValueKind.Object
+            },
+            impl: (interpreter: Interpreter, str: BrsString) => {
+                const source = this.jsRegex.source;
+                const flags = this.jsRegex.flags + "g";
```
strattonbrazil: Same as above.

src/brsTypes/components/Regex.ts
```diff
@@ -0,0 +1,153 @@
...
+import { Interpreter } from "../../interpreter";
+import { BrsArray } from "./BrsArray";
+
+export class Regex extends BrsComponent implements BrsValue {
+    readonly kind = ValueKind.Object;
+    private jsRegex : RegExp;
+
+    constructor(expression: BrsString, flags = new BrsString("")) {
+        super("roRegex");
+        this.jsRegex = new RegExp(expression.value, flags.value);
```
strattonbrazil: > All flags in roRegex but "x" should be supported.

Maybe we can display a not-implemented warning here if that is passed in.

sjbarag:
> @alimnios72: Although javascript's RegExp doesn't fully support PCRE, as Brightscript does, it gives brs the ability to support most of the common regex use cases.

Looking at `roRegex`, it seems unclear how much of PCRE is even supported on device!

> In general, most Perl compatible regular expressions are supported.

I suspect we'll get a lot of "you're missing X weird corner case" bug reports once people start using `roRegex` heavily, but I'm okay with that for now.  Other than putting the reference implementation through a compliance test of PCRE (which we'd have to find and rewrite into BrightScript), I'm not sure of a good way to find which bits of PCRE are actually supported anyway.  And it turns out there's a PCRE2 library these days so I don't even know which set of compliance tests we'd want to run! 🤕💥

src/brsTypes/components/Regex.ts
```diff
@@ -0,0 +1,153 @@
...
+        "ismatch",
+        {
+            signature: {
+                args: [
+                    new StdlibArgument("str", ValueKind.String)
+                ],
+                returns: ValueKind.Boolean
+            },
+            impl: (interpreter: Interpreter, str: BrsString) => {
+                return this.jsRegex.test(str.value) ? BrsBoolean.True : BrsBoolean.False;
```
sjbarag: Real tiny nit: Would you mind replacing the ternary operator with `BrsBoolean.from(this.jsRegex.test(Str.value))`?  That's already set up to do this kind of mapping :)

src/brsTypes/components/Regex.ts
```diff
@@ -0,0 +1,153 @@
...
+                args: [
+                    new StdlibArgument("str", ValueKind.String)
+                ],
+                returns: ValueKind.Object
+            },
+            impl: (interpreter: Interpreter, str: BrsString) => {
+                const result = this.jsRegex.exec(str.value);
+                let arr = [];
+                if (result !== null) {
+                    for (let i = 0; i < result.length; i++) {
```
sjbarag: Consider using `Array.prototype.map` here!  Something like this should get the same thing in slightly less code:

```javascript
arr = result.map(found => found ? new BrsString(found) : BrsInvalid.Instance);
```

src/brsTypes/components/Regex.ts
```diff
@@ -0,0 +1,153 @@
...
+        }
+    );
+
+    private replace = new Callable(
+        "replace",
+        {
+            signature: {
+                args: [
+                    new StdlibArgument("str", ValueKind.String),
+                    new StdlibArgument("str", ValueKind.String)
```
sjbarag: Would you mind updating this argument to be called `"replacement"` like it is in the reference docs?  These names get reported to users when a type mismatch is encountered, so it'd be nice to keep it clear 😄

src/brsTypes/components/Regex.ts
sjbarag: Great catch, @strattonbrazil!  I agree.  Maybe something like `this.jsRegex.flags.indexOf("g") === -1 ? this.jsRegex.flags + "g" : this.jsRegex.flags`?  Or something less verbose 🤷‍♂️

src/brsTypes/components/Regex.ts
```diff
@@ -0,0 +1,153 @@
...
+        }
+    );
+
+    private replaceAll = new Callable(
+        "replaceall",
+        {
+            signature: {
+                args: [
+                    new StdlibArgument("str", ValueKind.String),
+                    new StdlibArgument("str", ValueKind.String)
```
sjbarag: Same as match – looks like the variable name doesn't match the docs!

src/brsTypes/components/Regex.ts
```diff
@@ -0,0 +1,153 @@
...
+                ],
+                returns: ValueKind.Object
+            },
+            impl: (interpreter: Interpreter, str: BrsString) => {
+                const source = this.jsRegex.source;
+                const flags = this.jsRegex.flags + "g";
+                this.jsRegex = new RegExp(source, flags);
+                let arr = [];
+                let matches:any;
```
sjbarag: Looks like we're using `matches` as an array, correct?  Looks like its type might be `: string[] | null` – would you mind updating this?

test/brsTypes/components/Regex.test.js
```diff
@@ -0,0 +1,192 @@
...
+            expect(str1.kind).toBe(ValueKind.Boolean);
+            expect(str1).toBe(BrsBoolean.True);
+
+            let str2 = isMatch.call(interpreter, new BrsString("HeLlO_XXX_WoRlD"));
+            expect(str2.kind).toBe(ValueKind.Boolean);
+            expect(str2).toBe(BrsBoolean.False);
+        });
+
+        it("matches a string in ISO8601 format", () => {
+            let regx = new Regex(new BrsString("P(\\d+Y)?(\\d+M)?(\\d+D)?T(\\d+H)?(\\d+M)?(\\d+(.\\d+)?S)?"));
```
sjbarag: **woah** these are some crazy regexes.  Thanks for adding some real-world cases!

src/brsTypes/components/Regex.ts
```diff
@@ -0,0 +1,153 @@
...
+            this.split,
+            this.matchAll
+        ]);
+    }
+
+    toString(parent?: BrsType): string {
+        return "<Component: roRegex>";
+    }
+
+    private parseReplacementPattern(pattern: string): string {
```
sjbarag: Would you mind adding a quick JSDoc string for this function?

src/brsTypes/components/Regex.ts
```diff
@@ -0,0 +1,153 @@
...
+        return pattern.replace(/\\/g, "\$");
+    }
+
+    private isMatch = new Callable(
```
sjbarag: I've been trying to keep some of the reference documentation included here to make working with these a little easier (e.g. in https://github.com/sjbarag/brs/blame/e2168603cda83b746a7101f2bb7f782b0178ce5f/src/brsTypes/components/AssociativeArray.ts#L159 where @jweeber added a simple sentence for each member function).

If you get some time, would you mind adding something small to each of the member functions here?

src/brsTypes/components/Regex.ts
alimnios72: Awesome, I like this better.

src/brsTypes/components/Regex.ts
alimnios72: My javascript is getting rusty 🙁

src/brsTypes/components/Regex.ts
alimnios72: 🤦‍♂️

test/brsTypes/components/Regex.test.js
alimnios72: 😸

src/brsTypes/components/Regex.ts
alimnios72: Sure

src/brsTypes/components/Regex.ts
alimnios72: I'm solving this with a different approach: avoiding the constructor to take unwanted flags. Since `g` is not a valid brighstcript flag it will be never duplicated.

test/e2e/resources/components/roRegex.brs
```diff
@@ -0,0 +1,21 @@
...
+
+    regex = createObject("roRegex", "\d+", "i")
+
+    print "First match" regex.match("123 456 789")
+    matches = regex.matchall("123 456 789")
+    print "All matches: "
+    print matches[0]
+    print matches[1]
+    print matches[2]
+end sub
```
sjbarag: Great test file!  I don't see its output getting compared to anything though so it might not be getting executed 😅 We'll need something like this:

https://github.com/sjbarag/brs/blob/e2168603cda83b746a7101f2bb7f782b0178ce5f/test/e2e/BrsComponents.test.js#L53-L62

alimnios72: My bad, didn't see this. The E2E is being executed now 😄

PR #221, opened 2019-04-05
feat(cli): Add --stop-after CLI argument

sjbarag:
Adds a new CLI argument that stops execution after the lexer, preprocessor, or parser.  I've built something like this directly into src/index.ts several times, but it's always been a one-off hack.  It's been consistently useful though, so it's time to support it long-term.

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -56,6 +56,11 @@ export async function execute(filenames: string[], options: Partial<ExecutionOpt
             return Promise.reject({
                 message: "Error occurred during lexing"
             });
+        } else if (options.stopAfter === "lexer") {
+            console.log(JSON.stringify(scanResults.tokens, null, 2));
+            return Promise.reject({
+                message: "Execution stopped early because of `--stop-after` argument."
+            });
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -56,20 +56,34 @@ export async function execute(filenames: string[], options: Partial<ExecutionOpt
...
+            });
         }
 
         let preprocessResults = preprocessor.preprocess(scanResults.tokens, manifest);
         if (preprocessResults.errors.length > 0) {
             return Promise.reject({
                 message: "Error occurred during pre-processing"
             });
+        } else if (options.stopAfter === "preprocessor") {
+            console.log(JSON.stringify(preprocessResults.processedTokens, null ,2));
```
alimnios72: Small nit:
```suggestion
            console.log(JSON.stringify(preprocessResults.processedTokens, null, 2));
```

sjbarag:
This is extremely out of date so I'm going to close it for now, but if we need this in the future the PR will still exist.

PR #225, opened 2019-04-30
fix(parse): Allow `sub` declarations with return-type specifications

### LINKED ISSUE #220: Allow sub with return type
While against the main motivation of of using a sub IMO, it is legal in brightscript to do something like `sub getVideoUrl(videoFilename as String) as String`. Currently this will spit out `'Sub' functions are always void returns, and can't have 'as' clauses` which surprisingly is not true. Since this is valid brightscript, it should be allowed.

sjbarag:
While `sub` is conventionally used only for functions with `void` return types, the Reference BrightScript Implementation (RBI) allows non-void return types to be declared on `sub`s, e.g.:

```brightscript
sub main()
    print getFive()
end sub

sub getFive() as integer
'             ~~~~~~~~~~
'             This is valid in RBI!
    return five
end sub
```

fixes #220

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -203,11 +203,8 @@ export class Parser {
             rightParen = advance();
 
             let maybeAs = peek();
-            if (check(Lexeme.Identifier) && maybeAs.text && maybeAs.text.toLowerCase() === "as") {
+            if (check(Lexeme.Identifier) && maybeAs.text.toLowerCase() === "as") {
                 advance();
-                if (isSub) {
-                    throw addError(previous(), "'Sub' functions are always void returns, and can't have 'as' clauses");
-                }
 
                 let typeToken = advance();
                 let typeString = typeToken.text || "";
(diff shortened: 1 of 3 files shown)


src/parser/Parser.ts
```diff
@@ -203,11 +203,8 @@ export class Parser {
             rightParen = advance();
 
             let maybeAs = peek();
-            if (check(Lexeme.Identifier) && maybeAs.text && maybeAs.text.toLowerCase() === "as") {
+            if (check(Lexeme.Identifier) && maybeAs.text.toLowerCase() === "as") {
```
sjbarag: All tokens have text now so no need for this!

src/parser/Parser.ts
```diff
@@ -203,11 +203,8 @@ export class Parser {
             rightParen = advance();
 
             let maybeAs = peek();
-            if (check(Lexeme.Identifier) && maybeAs.text && maybeAs.text.toLowerCase() === "as") {
```
sjbarag: No need to check if `maybeAs.text` is truthy before calling functions on it, since it's always defined and non-null! There's probably a lot of this around that'll get cleaned up eventually.

PR #233, opened 2019-05-08
On windows, spawn needs to call node in order to execute js files.

TwitchBronBron:
The cli tests fail when running on windows. I think it has to do with a js file not being executable on its own. This PR fixes that.

### DIFF
--- a/test/cli/cli.test.js
+++ b/test/cli/cli.test.js
@@ -2,32 +2,34 @@ const child_process = require("child_process");
 const path = require("path");
 const pify = require("pify");
 
-const execFile = pify(child_process.execFile);
+const exec = pify(child_process.exec);
 
 describe("cli", () => {
     it("accepts a root directory", () => {
         let rootDir = path.join(__dirname, "resources");
 
-        return execFile(
-            'node', [
-                path.join(process.cwd(), "bin", "cli.js"),
-                "--root", rootDir,
-                path.join(rootDir, "requires-manifest.brs")
-            ]
-        ).then((stdout, stderr) => {
+        let command = [
+            "node",
+            path.join(process.cwd(), "bin", "cli.js"),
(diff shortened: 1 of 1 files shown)


test/cli/cli.test.js
```diff
@@ -9,8 +9,8 @@ describe("cli", () => {
         let rootDir = path.join(__dirname, "resources");
 
         return execFile(
```
sjbarag: We might want to switch these to a plain `exec()` if we're calling `node` as the command - looks like `execFile` expects the first argument to be the file to execute!

sjbarag:
```suggestion
         return exec(
```

test/cli/cli.test.js
```diff
@@ -21,8 +21,11 @@ describe("cli", () => {
 
     it("defaults --root to process.cwd()", () => {
         return execFile(
```
sjbarag:
```suggestion
         return exec(
```

test/cli/cli.test.js
sjbarag: I can't add suggestions to the `const execFile = ` line :cry:

test/cli/cli.test.js
```diff
@@ -2,29 +2,34 @@ const child_process = require("child_process");
...
 
-const execFile = pify(child_process.execFile);
+const exec = pify(child_process.exec);
 
 describe("cli", () => {
     it("accepts a root directory", () => {
         let rootDir = path.join(__dirname, "resources");
 
-        return execFile(
+        let command = [
```
TwitchBronBron: exec takes a single string with the full command, so I abstracted the command out into its own line so it's more readable

TwitchBronBron:
@sjbarag Ready for your review

PR #234, opened 2019-05-09
Consume trailing colons for if statements

### LINKED ISSUE #230: Colons after `if` statement condition cause parse failures
Similar to #195 , a colon following an `if`/`else if` statement's condition is valid in RBI!  It looks shockingly like python, though I suspect that's not the intention 😉  Another case of "colons are treated like newlines most of the time!"  Minimal working example:

```brightscript
sub main()
    a = 1
    if a > 0:
        print "positive!"
    else if a < 0:
        print "negative!"
    else:
        print "tHaT NuMbEr iS ZeRo"
    end if
end sub
```

TwitchBronBron:
Updates parser to support trailing colons after `if` and `elseif` and `else` statements. 

Fixes #230


src/parser/Parser.ts
```diff
@@ -618,6 +618,9 @@ export class Parser {
                 then = advance();
             }
 
+            //consume any trailing colons
+            while (check(Lexeme.Colon)) { advance(); }
```
sjbarag: This might need to get moved into the `if(match(Lexeme.Newline))` on line 624 - I suspect using `if (condition):` forces the parser into "multi-line `if`" mode.  I still have to confirm against RBI but I'd bet this isn't valid syntax:

```brightscript
function t()
    if 1 < 2: return true

    return false
end function
```

sjbarag: Confirmed that a `:` after the condition causes the compiler to assume it's a multi-line if statement :/

TwitchBronBron:
Good catch! I'll work on these and let you know when it's ready for another review.

src/parser/Parser.ts
```diff
@@ -615,21 +617,38 @@ export class Parser {
 
             if (checkThen()) {
                 // `then` is optional after `if ...condition...`, so only advance to the next token if `then` is present
-                then = advance();
+                thenToken = advance();
```
TwitchBronBron: Renamed this variable to match the other variables in this function

src/parser/Parser.ts
```diff
@@ -830,7 +865,7 @@ export class Parser {
             }
 
             let toReturn = expression();
-            while (match(Lexeme.Newline));
+            while (match(Lexeme.Newline, Lexeme.Colon));
```
TwitchBronBron: This fixes an issue with using return in a single-line if statement (`if true: return true: end if`) and also, as a byproduct, enables support for multiple returns in a single statement (`return true : return true`)

src/parser/Parser.ts
```diff
@@ -661,16 +684,28 @@ export class Parser {
...
                 } else {
                     match(Lexeme.Newline);
-                    endIf = consume(
+                    endIfToken = consume(
                         `Expected 'end if' to close 'if' statement started on line ${startingLine}`,
                         Lexeme.EndIf
                     );
+                    
+                    //ensure that single-line `if` statements have a colon right before 'end if'
+                    if (ifToken.location.start.line === endIfToken.location.start.line) {
```
TwitchBronBron: I'm detecting a missing colon by checking that the `if` token is on the same line as the `end if`. It feels a little weird comparing line numbers...but I couldn't find another way around it. Open to suggestions, but it works...

TwitchBronBron:
@sjbarag ready for your review

src/parser/Parser.ts
sjbarag: Interesting!  I suspect moving this check and error message up to line 695 and checking for either a newline or a `:` would get the job done nice and cleanly, without having to compare line numbers directly :smiley: 

Probably something like this?
```typescript
if (match(Lexeme.Newline, Lexeme.Colon)) {
    while (match(Lexeme.Newline, Lexeme.Colon));
} else if (check(Lexeme.EndIf)) {
    addError(peek(), "Expected ':' or newline to preceed 'end if'");
}
```

TwitchBronBron: I just tried your proposal, and it didn't work. It failed a few of the tests. I think because this code is after the `elseif` block captures. The point of this code is to detect when the `endif` is on the SAME line as the `if`...and in the situation where the are newlines in any of the `elseif` sections, your proposed solution wouldn't detect that. 

As an alternative, I could inspect every token between the `if` and the `endif`, and throw an error if no colon and no newline were detected. **However**, that's more expensive than a simple line check...not sure it's worth the effort.

sjbarag: I see!  Interestingly, I don't seem to be able to get into this branch at all - putting an `endif`-terminated `if` statement on one line with no `:` preceding the `end if` didn't report "Expected ':' or newline to precede 'end if'":

```brightscript
sub main()
    if true: print "foo": else: print "bar" end if
end sub

sub main2()
    if true: print "foo" end if
end sub
```

produces
```
$ brs scratch/newline.brs 
scratch/newline.brs(2,44-50): Found unexpected token 'end if'
scratch/newline.brs(3,0-7): Found unexpected token 'end sub'
scratch/newline.brs(6,25-31): Found unexpected token 'end if'
scratch/newline.brs(7,0-7): Found unexpected token 'end sub'
scratch/newline.brs(6,4-6): Expected 'end if', 'else if', or 'else' to terminate 'then' block
scratch/newline.brs(6,25-31): Found unexpected token 'end if'
scratch/newline.brs(8,0-1): (At end of file) Expected 'end sub' to terminate sub block
Error occurred parsing
```

Perhaps I'm missing something!

TwitchBronBron: So apparently I was wrong about the elseif blocks being processed. Here's a statement that produces the `Expected ':' to preceed 'end if'` error message.

```brightscript
function main()
    if 1 < 2 : return true end if
end function
```

TwitchBronBron:
@sjbarag this is ready for your review again

src/parser/Parser.ts
sjbarag: I see!  This works for now, but I have a gut feeling that the `block` function is going to learn some of these tricks pretty soon :wink:

PR #246, opened 2019-05-29
Empty print statement

### LINKED ISSUE #179: Empty print should be allowed
`print` by itself is a valid line of Brightscript that can be used to add a new line. Doing this throws an error ```Found unexpected token '
'```

TwitchBronBron:
Fixes #179

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,8 @@
     "clean": "rimraf ./lib ./types",
     "test": "jest",
     "lint": "tslint --project .",
-    "prettier:write": "prettier --write '{bin,src,test}/**/*.{js,ts}'",
-    "prettier": "prettier --check '{bin,src,test}/**/*.{js,ts}'",
+    "prettier:write": "prettier --write \"{bin,src,test}/**/*.{js,ts}\"",
+    "prettier": "prettier --check \"{bin,src,test}/**/*.{js,ts}\"",
     "prepublishOnly": "npm-run-all --serial clean build lint prettier test"
   },
   "files": [
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -1059,7 +1059,7 @@ export class Parser {
 
             //print statements can be empty, so look for empty print conditions
             if (isAtEnd() || check(Lexeme.Newline, Lexeme.Colon)) {
-                var emptyStringLiteral = new Expr.Literal(new Int32(1), printKeyword.location);
+                var emptyStringLiteral = new Expr.Literal(new BrsString(""), printKeyword.location);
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -14,8 +14,8 @@
     "clean": "rimraf ./lib ./types",
     "test": "jest",
     "lint": "tslint --project .",
-    "prettier:write": "prettier --write '{bin,src,test}/**/*.{js,ts}'",
-    "prettier": "prettier --check '{bin,src,test}/**/*.{js,ts}'",
+    "prettier:write": "prettier --write \"{bin,src,test}/**/*.{js,ts}\"",
```
sjbarag: Any idea what happened with this change?  Doesn't seem like it's necessary 😃

src/parser/Parser.ts
```diff
@@ -1056,7 +1056,14 @@ export class Parser {
                 | Expr.Expression
                 | Stmt.PrintSeparator.Tab
                 | Stmt.PrintSeparator.Space)[] = [];
-            values.push(expression());
+
+            //print statements can be empty, so look for empty print conditions
+            if (isAtEnd() || check(Lexeme.Newline, Lexeme.Colon)) {
+                var emptyStringLiteral = new Expr.Literal(new Int32(1), printKeyword.location);
```
sjbarag: I'm a little confused – won't this print `1` every time?  It doesn't seem like an empty string 😃

sjbarag: Dangit github updated while I was reviewing

src/parser/Parser.ts
```diff
@@ -1056,7 +1056,14 @@ export class Parser {
                 | Expr.Expression
                 | Stmt.PrintSeparator.Tab
                 | Stmt.PrintSeparator.Space)[] = [];
-            values.push(expression());
+
+            //print statements can be empty, so look for empty print conditions
+            if (isAtEnd() || check(Lexeme.Newline, Lexeme.Colon)) {
+                var emptyStringLiteral = new Expr.Literal(new BrsString(""), printKeyword.location);
```
sjbarag: Mind switching this to a `let` real quick?  I could have sworn I had `tslint` configured to disallow `var` 🤔

package.json
sjbarag: Ah, mentioned in another PR: https://github.com/sjbarag/brs/pull/247#discussion_r288729479

Nice catch!

src/parser/Parser.ts
TwitchBronBron: @sjbarag done.

PR #247, opened 2019-05-29
Stop statement syntax support

### LINKED ISSUE #192: Stop keyword causes parse errors
There are some issues with the `stop` keyword now, probably introduced in #164. 

```brightscript
sub Main()
    'this SHOULD be an error
    stop = 123
    
    'this SHOULD be an error
    stop()

    'this should NOT be an error (but...why would you do this?)
    print stop

    'this should NOT be an error
    stop

    'this should NOT be an error
    if true then stop

    obj = {
        'this should NOT be an error (working correctly right now)
        stop: true
    }
end sub

'this SHOULD be an error
sub Stop()
end sub

'this SHOULD be an error
sub DoSomething(stop as boolean)
    
end sub

### PR DESCRIPTION
**Pull request by TwitchBronBron**
Fixes #192

### DIFF
--- a/package.json
+++ b/package.json
@@ -14,8 +14,8 @@
     "clean": "rimraf ./lib ./types",
     "test": "jest",
     "lint": "tslint --project .",
-    "prettier:write": "prettier --write '{bin,src,test}/**/*.{js,ts}'",
-    "prettier": "prettier --check '{bin,src,test}/**/*.{js,ts}'",
+    "prettier:write": "prettier --write \"{bin,src,test}/**/*.{js,ts}\"",
+    "prettier": "prettier --check \"{bin,src,test}/**/*.{js,ts}\"",
     "prepublishOnly": "npm-run-all --serial clean build lint prettier test"
   },
   "files": [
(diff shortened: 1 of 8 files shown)

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -14,8 +14,8 @@
     "clean": "rimraf ./lib ./types",
     "test": "jest",
     "lint": "tslint --project .",
-    "prettier:write": "prettier --write '{bin,src,test}/**/*.{js,ts}'",
-    "prettier": "prettier --check '{bin,src,test}/**/*.{js,ts}'",
+    "prettier:write": "prettier --write \"{bin,src,test}/**/*.{js,ts}\"",
```
TwitchBronBron: The single quotes break running this command on windows, so I had to change to double quotes

src/parser/Statement.ts
```diff
@@ -293,6 +293,27 @@ export class End implements Statement {
...
+        }
+    ) {}
+
+    accept<R>(visitor: Visitor<R>): BrsType {
+        //TODO implement this in the runtime. It should pause code execution until a `c` command is issued from the console
+        throw new Error("Not implemented");
+    }
+
+    get location() {
+        return {
```
sjbarag: I suppose we could have gotten away with `return this.tokens.stop.location` but 🤷‍♂

PR #248, opened 2019-05-29
Added parser support for `goto` and labels

### LINKED ISSUE #158: Add support for goto
BrightScript supports `goto` statements, C-style!  They currently cause parse failures because an identifier followed by a `:` isn't supported as a statement.

From the spec: https://sdkdocs.roku.com/display/sdkdoc/Program+Statements#ProgramStatements-GOTOlabel

TwitchBronBron:
Fixes #158


src/parser/Parser.ts
```diff
@@ -1102,6 +1128,36 @@ export class Parser {
...
+         * Parses a `goto` statement
+         * @returns an AST representation of an `goto` statement.
+         */
+        function labelStatement() {
+            let tokens = {
+                identifier: advance(),
+                colon: advance(),
+            };
+
+            consume("Labels must be declared on their own line", Lexeme.Newline, Lexeme.Eof);
```
sjbarag: Since this is here, we might be able to drop the `checkNextNext()` function entirely!

sjbarag: Since we're already checking for newlines or EOF following a label, we might be able to drop the `checkNextNext()` function entirely!

src/parser/Parser.ts
```diff
@@ -547,6 +547,24 @@ export class Parser {
...
+            if (current > 0 && checkPrevious(Lexeme.Newline) === false) {
+                return false;
+            }
+
+            //labels are specifically detected when there's an identifier, a colon, then a newline or eof.
+            //If this check gets more greedy, it causes other issues.
+            return (
+                check(Lexeme.Identifier) &&
+                checkNext(Lexeme.Colon) &&
+                checkNextNext(Lexeme.Newline, Lexeme.Eof)
```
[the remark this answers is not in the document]
TwitchBronBron: No, I don't think that's possible. Take this for example: 

```brightscript
video = {
    url: "some_url"
}
```

We would enter the label statement for every inline object property. Then, in my current implementation, it would throw an error because there was not a newline or eof. 

If we always enter the label in this situation, we would then need to UNDO the consumes for the label and colon if we found something other than a newline or eof. I think I personally prefer just doing `checkNextNext`. 

Also, none of the other conditions in `statement()` back themselves out. They all pre-check, and then return the thing if true. By walking into `labelStatement()` and then backing out if not a match, we would be deviating from the norm. Not un-doable, but I still like this implementation better. 

Ultimately your call, so let me know.

src/parser/Parser.ts
TwitchBronBron: No, because this would throw errors for valid brightscript like
```brightscript
video = {
    url: "some_url"
}
```

src/parser/Parser.ts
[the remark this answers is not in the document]
sjbarag: > We would enter the label statement for every inline object property.

That's not the case!  Reading `{` enters a "read associative array literal" phase in the parser without calling back into `statement`, where each property is a literal, a colon, an expression, an optional comma, and either a newline or a colon:  

https://github.com/sjbarag/brs/blob/3b0bc9b5ea514e11574dff8e6f9df37bcb175cd0/src/parser/Parser.ts#L1389-L1443

From what I can tell, no statement is allowed to be an identifier followed by a newline other than these `labels`, so removing `checkNextNext` seems pretty safe in my testing:

```brightscript
' labels aren't allowed outside of functions
' foo:
sub main()
    ' bypasses the `print` statement entirely
    goto derp

    ' labels are allowed in functions
    bar:
    print "bar!"

    ' with no parse errors in associative literals
    derp = {
        foo: "foo",
        bar: "bar",
        baz: "baz"
    }

    ' defines a new label instead of being an expression-statement that references only the `derp` variable
    derp:
end sub
```

Let me know if you find something different, but that snippet seems to parse just fine in RBI and in `brs` without `checkNextNext` 😃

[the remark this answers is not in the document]
TwitchBronBron: Ah, very good point.

TwitchBronBron:
@sjbarag I have removed `checkNextNext`. This is ready for your review again.

src/parser/Parser.ts
```diff
@@ -1114,6 +1123,36 @@ export class Parser {
             return new Stmt.Return(tokens, toReturn);
         }
 
+        /**
+         * Parses a `goto` statement
```
sjbarag: Looks like a bit of a copy-pasta error in the docs for `labelStatement` 😃

TwitchBronBron: Fixed!

TwitchBronBron: Good catch btw.

TwitchBronBron:
@sjbarag I fixed that jsdoc copy paste issue. ready for your review.

PR #251, opened 2019-06-04
rough implementation of RoSGNode/ifAssociativeArray

strattonbrazil:
This implements RoSGNode--specifically the ifAssociativeArray interface.  This is almost a straight copy of the roAssociativeArray implementation, which will serve as a foundation for future interfaces such as supporting fields.  While some of the code might be shareable between them in the Node's current state, I imagine the Node implementation will have a lot of special cases to handle fields, so these functions are placeholders for future interfaces, which will slightly modify their behavior.

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -14,7 +14,7 @@ class Field {
 
 export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
     readonly kind = ValueKind.Object;
-    private elements = new Map<string, BrsType>();
+    elements = new Map<string, BrsType>();
     private fields = new Map<string, Field>();
 
     constructor(elements: AAMember[], readonly type: string = "Node") {
@@ -47,8 +47,8 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         return [
             `<Component: ${componentName}> =`,
             "{",
-            ...Array.from(this.elements.keys()).map(
-                key => `    ${key}: ${this.elements.get(key)!.toString(this as any)}`
+            ...Array.from(this.elements.entries()).map(
+                ([key, value]) => `    ${key}: ${value.toString(this)}`
             ),
             "}",
         ].join("\n");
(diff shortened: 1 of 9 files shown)


src/brsTypes/components/RoSGNode.ts
```diff
@@ -0,0 +1,222 @@
...
+
+        if (parent) {
+            return `<Component: ${componentName}>`;
+        }
+
+        return [
+            `<Component: ${componentName}> =`,
+            "{",
+            ...Array.from(this.elements.keys()).map(
+                key => `    ${key}: ${this.elements.get(key)!.toString(this as any)}`
```
strattonbrazil: The "as any" is to avoid a type error as this class isn't a BrsType, I guess.  Not sure how to improve this.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -0,0 +1,222 @@
...
+    private append = new Callable("append", {
+        signature: {
+            args: [new StdlibArgument("obj", ValueKind.Object)],
+            returns: ValueKind.Void,
+        },
+        impl: (interpreter: Interpreter, obj: BrsType) => {
+            if (obj instanceof RoAssociativeArray || obj instanceof RoSGNode) {
+                this.elements = new Map<string, BrsType>([
+                    ...this.elements,
+                    ...(obj as any).elements,
```
strattonbrazil: Similar to the above, I added this to avoid a type error, but I'm less sure what the problem is here.  Even when just checking for an RoSGNode instance I still get this error.

test/e2e/resources/components/roAssociativeArray.brs
```diff
@@ -5,13 +5,14 @@ sub main()
...
-    print "AA items size: " aa.items().count()              ' => 3
-    print "can delete elements: " aa.delete("baz")          ' => true
-    print "can look up elements: " aa.lookup("foo") = "foo" ' => true
-    print "can check for existence: " aa.doesExist("bar")   ' => true
+    print "AA size: " aa.count()                                ' => 3
+    print "AA keys size: " aa.keys().count()                    ' => 3
+    print "AA items size: " aa.items().count()                  ' => 3
+    print "can delete elements: " aa.delete("baz")              ' => true
+    print "can look up elements: " aa.lookup("foo") = "foo"     ' => true
+    print "can look up elements (brackets): " aa["foo"] = "foo" ' => true
```
strattonbrazil: Added the brackets case.  Not sure if it was tested elsewhere.

src/brsTypes/components/RoSGNode.ts
sjbarag: Yep, I think that might have been solved by a newer typescript version?

src/brsTypes/components/RoSGNode.ts
sjbarag: This one was caused by the fact that `elements` wasn't public on `RoAssociativeArray`.  It is now!

PR #254, opened 2019-06-07
feat(stdlib,interp): Implement autoboxing and roString (ifStringOps only)

sjbarag:
`roString` — specifically the methods described in `ifStringOps` — is one of the most often-used non-SceneGraph types in the BrightScript language.  It's not necessarily directly created by BrightScript authors, but it gets created whenever `foo.split("/")` or `bar.encodeUriComponent()` are called due to the reference implementation's automatic boxing rules.  Basically, if you use a boxable primitive like an object, it'll get automatically boxed for you, and the temporary boxed value will be destroyed at the end of the operation.

This PR implements the `ifStringOps` functions on a new `roString` type, and includes that automatic boxing behavior.

### DIFF
--- a/src/brsTypes/Boxing.ts
+++ b/src/brsTypes/Boxing.ts
@@ -11,12 +11,7 @@ export interface Unboxable {
 }
 
 export function isBoxable(value: BrsType): value is BrsType & Boxable {
-    switch (value.kind) {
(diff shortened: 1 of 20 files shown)


sjbarag:
Ignore the commit history; it'll get squashed

PR #258, opened 2019-06-11
Parser/lexer more class like

TwitchBronBron:
The current parser and lexer don't fully leverage the class structure. They both have a function (`scan` or `parse`) that contain many inner functions. This PR moves all of those inner functions out as methods and properties of `Lexer` or `Parser` itself. This has a few benefits:

1. TypeScript adds methods to the prototype of a class, reducing the memory footprint when instantiating one of these classes. 
2. Every method on a class is available to spy/mock/stub during unit testing, thus providing more options when trying to test very specific use cases. This is not the case when the functions are hidden as local variables. 
3. Subclassing is made easier (for any external project interested in doing so)

I know this looks like a ton of changes. However, 99% of the changes were just changing the keyword `function` to `private`, and then adding `this.` in front of methods and properties. All unit tests are still passing. 

GitHub doesn't do the best job when creating the diff. If you copy the files locally overtop of your master branch, and open the diff viewer in vscode, and enable `ignore trim whitespace`, you'll see a much cleaner diff view. Here's a screenshot of how it looks in vscode's diff viewer.
image

GitHub does have the ability to ignore whitespace changes. That helps tremendously when looking at the changes. 
image


TwitchBronBron:
I tried running `npm run prettier:write` locally, but the CI build is complaining about how I didn't run prettier. 
image

TwitchBronBron:
Found the prettier issues. I had to delete node_modules and run npm install again.

src/lexer/Lexer.ts
```diff
@@ -55,6 +55,46 @@ export class Lexer {
         this.events.once("err", errorHandler);
     }
 
+    /**
+     * The zero-indexed position at which the token under consideration begins.
+     */
+    private start = 0;
```
TwitchBronBron: Moved all of these variables out as properties.

src/parser/Parser.ts
```diff
@@ -184,1368 +188,1390 @@ export class Parser {
...
-     * Parses an array of `Token`s into an abstract syntax tree that can be executed with the `Interpreter`.
-     * @param toParse the array of tokens to parse
-     * @returns an array of `Statement` objects that together form the abstract syntax tree of the
-     *          program
+     * The current token index. This gets reset on every parse
      */
-    parse(toParse: ReadonlyArray<Token>): ParseResults {
-        let current = 0;
-        let tokens = toParse;
+    private current = 0;
```
TwitchBronBron: Moved all of the local variables from parse out into private properties

src/parser/Parser.ts
```diff
@@ -184,1368 +188,1390 @@ export class Parser {
...
+                    endingKeyword,
+                    `Expected 'end ${functionType.text}' to terminate ${functionType.text} block`
                 );
             }
 
-            let name = advance();
-            let type: ValueKind = ValueKind.Dynamic;
-            let typeToken: Token | undefined;
-            let defaultValue;
+            let func = new Expr.Function(args, returnType, body, startingKeyword, endingKeyword);
```
TwitchBronBron: I think prettier reformatted this line because it now fits on one line since it's de-indented

src/parser/Parser.ts
```diff
@@ -184,1368 +188,1390 @@ export class Parser {
...
+                Lexeme.Eof
+            );
+        }
 
-            while (!check(Lexeme.Newline, Lexeme.Colon, ...additionalterminators) && !isAtEnd()) {
-                if (check(Lexeme.Semicolon)) {
-                    values.push(advance() as Stmt.PrintSeparator.Space);
-                }
+        if (expr instanceof Expr.Call) {
+            re
```
TwitchBronBron: This got moved down ~100 lines and was renamed `aaKey`

TwitchBronBron: Renamed to `primaryExpression` to make it a little more clear what it does. This can be reverted if you don't like that.

TwitchBronBron: This is the `key` function from `primary`/`primaryExpression` function ~100 lines above

sjbarag:
Hey dude!  In general, I'm not inclined to turn the parser into a not-opaque structure due to its tight coupling to the language being parsed.  Since [brighterscript](TwitchBronBron/brighterscript) is currently in a proposal state, it seems unlikely that either of us have a good idea of what's actually going to be required to make that work just yet.  Here's what I'm proposing:

1. Keep your fork of `brs` and these changes in it
2. Start writing Brighterscript by having it depend on your fork of `brs`!  Tools like [lerna](https://lerna.js.org/) or [yarn's workspaces](https://yarnpkg.com/lang/en/docs/workspaces/) should make that very easy 😃  You'll have full control over your fork and can make any changes you need as you identify them, so you won't be blocked by this project.
3. Once brighterscript matures and supports some of the [features that've been proposed](https://github.com/TwitchBronBron/brighterscript/issues), we'll revisit whether or not `brs`'s parser can help with a much less abstract goal in mind!  The changes from your `brs` fork could easily be contributed back into the upstream repo (this one) if appropriate, or maybe we'll find that they aren't necessary for one reason or another.  Time will tell!

In the meantime, I'm going to close this PR but I'd be happy to revisit this topic once brighterscript's specific uses have been built out.

PR #261, opened 2019-06-11
Moar Azure Devops

sjbarag:


sjbarag:
GitHub Actions has been sufficient, and also supports Windows builds for when we need them

PR #264, opened 2019-06-19
Implement if sg node children

gincwang:
This PR implements 6 of the most used functions in `ifSGNodeChildren` interface:
- appendChild()
- getChildCount()
- getChildren()
- removeChild()
- getParent()
- createChild()

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -16,6 +16,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
     readonly kind = ValueKind.Object;
     elements = new Map<string, BrsType>();
     private fields = new Map<string, Field>();
+    private children = new Array<RoSGNode>();
 
     constructor(elements: AAMember[], readonly type: string = "Node") {
         super("roSGNode");
@@ -34,6 +35,9 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             this.keys,
             this.items,
             this.lookup,
+            //ifSGNodeChildren
+            this.appendchild,
+            this.getchildcount,
         ]);
     }
 
@@ -211,6 +215,35 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         impl: (interpreter: Interpreter, key: BrsString) => {
             let lKey = key.value.toLowerCase();
             return this.get(new BrsString(lKey));
+        },
+    });
+
+    private getchildcount = new Callable("getchildcount", {
+        signature: {
+            args: [],
+            returns: ValueKind.Int32,
+        },
+        impl: (interpreter: Interpreter) => {
+            return new Int32(this.children.length);
+        },
+    });
+
+    private appendchild = new Callable("appendchild", {
+        signature: {
+            args: [new StdlibArgument("child", ValueKind.Dynamic)],
+            returns: ValueKind.Boolean,
+        },
+        impl: (interpreter: Interpreter, child: BrsType) => {
+            for (let childNode of this.children) {
+                if (childNode === child) {
+                    return BrsBoolean.True;
+                }
+            }
+            if (child instanceof RoSGNode) {
+                this.children.push(child);
+                return BrsBoolean.True;
+            }
+            return BrsBoolean.False;
         },
     });
 }
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -42,6 +42,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             this.getchildren,
             this.removechild,
             this.getparent,
+            this.createchild,
         ]);
     }
 
@@ -332,6 +333,23 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         },
         impl: (interpreter: Interpreter) => {
             return this.parent;
+        },
+    });
+
+    /* Creates a child node of type nodeType, and adds the new node to the end of the
+    subject node list of children */
+    private createchild = new Callable("createchild", {
+        signature: {
+            args: [new StdlibArgument("nodetype", ValueKind.String)],
+            returns: ValueKind.Object,
+        },
+        impl: (interpreter: Interpreter, nodetype: BrsString) => {
+            // currently we can't create a custom subclass object of roSGNode,
+            // so we'll always create generic RoSGNode object as child
+            var child = new RoSGNode([]);
+            this.children.push(child);
+            child.setParent(this);
+            return child;
         },
     });
 }
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -282,7 +282,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             let childrenSize = this.children.length;
             if (numChildrenValue <= -1 && indexValue === 0) {
                 //short hand to return all children
-                return new RoArray(this.children);
+                return new RoArray(this.children.slice());
             } else if (numChildrenValue <= 0 || indexValue < 0 || indexValue > childrenSize - 1) {
                 //these never return any children
                 return new RoArray([]);
@@ -346,9 +346,11 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         impl: (interpreter: Interpreter, nodetype: BrsString) => {
             // currently we can't create a custom subclass object of roSGNode,
             // so we'll always create generic RoSGNode object as child
-            var child = new RoSGNode([]);
-            this.children.push(child);
-            child.setParent(this);
+            var child = createNodeByType(nodetype);
+            if (child instanceof RoSGNode) {
+                this.children.push(child);
+                child.setParent(this);
+            }
             return child;
         },
     });
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -280,7 +280,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             let numChildrenValue = num_children.getValue();
             let indexValue = index.getValue();
             let childrenSize = this.children.length;
-            if (numChildrenValue <= -1 && indexValue == 0) {
+            if (numChildrenValue <= -1 && indexValue === 0) {
                 //short hand to return all children
                 return new RoArray(this.children);
             } else if (numChildrenValue <= 0 || indexValue < 0 || indexValue > childrenSize - 1) {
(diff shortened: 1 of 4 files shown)


src/brsTypes/components/RoSGNode.ts
```diff
@@ -213,6 +230,128 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+    subject node list of children */
+    private createchild = new Callable("createchild", {
+        signature: {
+            args: [new StdlibArgument("nodetype", ValueKind.String)],
+            returns: ValueKind.Object,
+        },
+        impl: (interpreter: Interpreter, nodetype: BrsString) => {
+            // currently we can't create a custom subclass object of roSGNode,
+            // so we'll always create generic RoSGNode object as child
+            var child = new RoSGNode([]);
```
strattonbrazil: Can you instead use _createNodeByType_?  Then when that function is fixed it will work here.

test/brsTypes/components/RoSGNode.test.js
```diff
@@ -278,4 +278,335 @@ describe("RoSGNode", () => {
...
+                let node = new RoSGNode([]);
+
+                let getChildCount = node.getMethod("getchildcount");
+                expect(getChildCount).toBeTruthy();
+
+                let result = getChildCount.call(interpreter);
+                expect(result).toEqual(new Int32(0));
+            });
+
+            //TODO: test when the node does have children
```
strattonbrazil: What's this TODO for?

gincwang: oops this was left when we didn't have any other functions implemented.  I think other test cases have covered this well, will remove

src/brsTypes/components/RoSGNode.ts
gincwang: oh sweet didn't realize we had this, will change it

src/brsTypes/components/RoSGNode.ts
```diff
@@ -213,6 +230,128 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+            ],
+            returns: ValueKind.Object,
+        },
+        impl: (interpreter: Interpreter, num_children: Int32, index: Int32) => {
+            let numChildrenValue = num_children.getValue();
+            let indexValue = index.getValue();
+            let childrenSize = this.children.length;
+            if (numChildrenValue <= -1 && indexValue == 0) {
+                //short hand to return all children
+                return new RoArray(this.children);
```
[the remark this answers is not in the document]
gincwang: good call!  will return a new array copy

src/brsTypes/components/RoSGNode.ts
```diff
@@ -213,6 +230,130 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+    return false if trying to remove anything that's not a node */
+    private removechild = new Callable("removechild", {
+        signature: {
+            args: [new StdlibArgument("child", ValueKind.Dynamic)],
+            returns: ValueKind.Boolean,
+        },
+        impl: (interpreter: Interpreter, child: BrsType) => {
+            var spliceIndex = -1;
+            if (child instanceof RoSGNode) {
+                for (var i = 0; i < this.children.length; i++) {
```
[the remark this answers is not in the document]
gincwang: good call

src/brsTypes/components/RoSGNode.ts
```diff
@@ -16,6 +16,8 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
     readonly kind = ValueKind.Object;
     elements = new Map<string, BrsType>();
     private fields = new Map<string, Field>();
+    private children = new Array<RoSGNode>();
```
sjbarag: Mind switching this to an array literal?  I don't think `new Array<Type>` is being used anywhere else.  Perhaps I should add a lint check for that 😄 

```suggestion
    private children: RoSGNode[] = [];
```

src/brsTypes/components/RoSGNode.ts
```diff
@@ -213,6 +230,129 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+
+    /* Adds a child node to the end of the subject node list of children so that it is
+    traversed last (of those children) during render. */
+    private appendchild = new Callable("appendchild", {
+        signature: {
+            args: [new StdlibArgument("child", ValueKind.Dynamic)],
+            returns: ValueKind.Boolean,
+        },
+        impl: (interpreter: Interpreter, child: BrsType) => {
+            for (let childNode of this.children) {
```
sjbarag: We may be able to reduce this down to `if (this.children.includes(child)) { return BrsBoolean.True }`.  Would you mind switching to that if you get time?

src/brsTypes/components/RoSGNode.ts
```diff
@@ -213,6 +230,129 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+                if (childNode === child) {
+                    return BrsBoolean.True;
+                }
+            }
+            if (child instanceof RoSGNode) {
```
sjbarag: This is gorgeous!

src/brsTypes/components/RoSGNode.ts
```diff
@@ -213,6 +230,129 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+            let childrenSize = this.children.length;
+            if (numChildrenValue <= -1 && indexValue === 0) {
+                //short hand to return all children
+                return new RoArray(this.children.slice());
+            } else if (numChildrenValue <= 0 || indexValue < 0 || indexValue >= childrenSize) {
+                //these never return any children
+                return new RoArray([]);
+            } else {
+                //only valid cases
+                let endSliceIndex = Math.min(indexValue + numChildrenValue, childrenSize);
```
sjbarag: Looks like this `Math.min` isn't necessary – Javascript's `Array.prototype.slice()` won't attempt to read past the end of an array, and effectively does this `min` operation internally!

src/brsTypes/components/RoSGNode.ts
```diff
@@ -213,6 +230,129 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+    subject node list of children */
+    private createchild = new Callable("createchild", {
+        signature: {
+            args: [new StdlibArgument("nodetype", ValueKind.String)],
+            returns: ValueKind.Object,
+        },
+        impl: (interpreter: Interpreter, nodetype: BrsString) => {
+            // currently we can't create a custom subclass object of roSGNode,
+            // so we'll always create generic RoSGNode object as child
+            var child = createNodeByType(nodetype);
```
sjbarag: Could you please switch these `var`s to `let`s?  JavaScript's scoping of `var` is sometimes surprising, and `let` is basically always what you want 😃

src/brsTypes/components/RoSGNode.ts
```diff
@@ -213,6 +230,129 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+    return false if trying to remove anything that's not a node */
+    private removechild = new Callable("removechild", {
+        signature: {
+            args: [new StdlibArgument("child", ValueKind.Dynamic)],
+            returns: ValueKind.Boolean,
+        },
+        impl: (interpreter: Interpreter, child: BrsType) => {
+            var spliceIndex = -1;
+            if (child instanceof RoSGNode) {
+                for (let i = 0; i < this.children.length; i++) {
```
sjbarag: This for-loop looks like it's finding the index of `child` in `this.children`, which is provided by the standard library in [Array.prototype.indexOf](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/indexOf).  Consider using that via `let indexOfChild = this.children.indexOf(child)`!

test/brsTypes/components/RoSGNode.test.js
```diff
@@ -278,4 +278,333 @@ describe("RoSGNode", () => {
...
+                appendChild.call(interpreter, child4);
+
+                // get last 2 children
+                let result = getChildren.call(interpreter, new Int32(2), new Int32(2));
+                let childCount = getChildCount.call(interpreter);
+                expect(childCount).toEqual(new Int32(4));
+                expect(result).toBeInstanceOf(RoArray);
+                expect(result.elements).toEqual([child3, child4]);
+
+                // pass in more children count than there are children, should just return 2 with offset of 2
```
sjbarag: There seem to be a few tests cases in this single `it`.  For future PRs, consider splitting those into separate test cases — possibly in a `describe()` with a `beforeEach()` to initialize the parent and children!

test/brsTypes/components/RoSGNode.test.js
```diff
@@ -278,4 +278,333 @@ describe("RoSGNode", () => {
...
+
+                appendChild.call(interpreter, child1);
+                appendChild.call(interpreter, child2);
+                appendChild.call(interpreter, child3);
+
+                // remove child 2
+                removeChild.call(interpreter, child2);
+                let result = getChildren.call(interpreter, new Int32(-1), new Int32(0));
+                let childCount = getChildCount.call(interpreter);
+                expect(removeChild).toBeTruthy();
```
sjbarag: I was only using `.toBeTruth()` to ensure the method existed before I called it.  Since you've already called `removeChild` by this line, there's no need to check it 😛

test/brsTypes/components/RoSGNode.test.js
```diff
@@ -278,4 +278,333 @@ describe("RoSGNode", () => {
...
+                // try to remove child 1 again, but shouldn't do anything
+                removeChild.call(interpreter, child1);
+                let result3 = getChildren.call(interpreter, new Int32(-1), new Int32(0));
+                let childCount3 = getChildCount.call(interpreter);
+                expect(childCount3).toEqual(new Int32(0));
+                expect(result3).toBeInstanceOf(RoArray);
+                expect(result3.elements).toEqual([]);
+            });
+
+            it("remove non-node things from parent", () => {
```
sjbarag: great test!

src/brsTypes/components/RoSGNode.ts
gincwang: oof what is this, built-in language features???

test/brsTypes/components/RoSGNode.test.js
gincwang: good call!  will break it up :)

test/brsTypes/components/RoSGNode.test.js
gincwang: let me move the check a bit upwards so it makes more sense

test/brsTypes/components/RoSGNode.test.js
sjbarag: Oh I only meant for the next PR, you didn't have to do it for this one!  But I appreciate it 😍

PR #265, opened 2019-06-20
Implement ifSGNodeField interface

alimnios72:
This PR implements 6 of the most used functions in ifSGNodeField interface:

- setField
- addFields
- getField
- addField
- setFields
- removeField
- update (This made the more sense to be implemented as part of the fields since it isn't related with the ifSGNodeChildren)

I'm still working on the `observeField` implemented but I want to do a checkpoint here since I've changed quite a bit the RoSGNode class.

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -8,6 +8,7 @@ import { RoAssociativeArray } from "./RoAssociativeArray";
 import { RoArray } from "./RoArray";
 import { AAMember } from "./RoAssociativeArray";
 import { Float } from "../Float";
+import { Call } from "../../parser/Expression";
 
 class Field {
     // private callbacks;
@@ -42,6 +43,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             this.addfield,
             this.addfields,
             this.getfield,
+            this.removefield,
             this.setfield,
         ]);
     }
@@ -309,6 +311,22 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         },
         impl: (interpreter: Interpreter, fieldname: BrsString) => {
             return this.get(fieldname);
+        },
+    });
+
+    /** Removes the given field from the node */
+    private removefield = new Callable("removefield", {
+        signature: {
+            args: [new StdlibArgument("fieldname", ValueKind.String)],
+            returns: ValueKind.Boolean,
+        },
+        impl: (interpreter: Interpreter, fieldname: BrsString) => {
+            if (this.get(fieldname) !== BrsInvalid.Instance) {
+                this.elements.delete(fieldname.value);
+                this.fields.delete(fieldname.value);
+            }
+
+            return BrsBoolean.True;
         },
     });
 
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -12,7 +12,7 @@ import { Float } from "../Float";
 class Field {
     private type: string;
     private value: BrsType;
-    private observers: String[] = [];
+    private observers: string[] = [];
 
     constructor(value: BrsType, private alwaysNotify: boolean) {
         this.type = ValueKind.toString(value.kind);
@@ -244,8 +244,8 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             returns: ValueKind.Boolean,
         },
         impl: (interpreter: Interpreter, str: BrsString) => {
-            let deleted = this.fields.delete(str.value);
-            return BrsBoolean.from(deleted);
+            this.fields.delete(str.value);
+            return BrsBoolean.True; //RBI always returns true
         },
     });
 
@@ -361,10 +361,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         ) => {
             let defaultValue = this.getDefaultValue(type.value);
 
-            if (
-                defaultValue !== Uninitialized.Instance &&
-                this.get(fieldname) === BrsInvalid.Instance
-            ) {
+            if (defaultValue !== Uninitialized.Instance && !this.fields.has(fieldname.value)) {
                 this.set(fieldname, defaultValue, alwaysnotify.toBoolean());
             }
 
@@ -385,7 +382,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
 
             fields.getValue().forEach((value, key) => {
                 let fieldName = new BrsString(key);
-                if (this.get(fieldName) === BrsInvalid.Instance) {
+                if (!this.fields.has(key)) {
                     this.set(fieldName, value);
                 }
             });
@@ -431,11 +428,8 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             returns: ValueKind.Boolean,
         },
         impl: (interpreter: Interpreter, fieldname: BrsString) => {
-            if (this.get(fieldname) !== BrsInvalid.Instance) {
-                this.fields.delete(fieldname.value);
-            }
-
-            return BrsBoolean.True;
+            this.fields.delete(fieldname.value);
+            return BrsBoolean.True; //RBI always returns true
         },
     });
 
@@ -451,7 +445,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         impl: (interpreter: Interpreter, fieldname: BrsString, value: BrsType) => {
             let field = this.get(fieldname);
 
-            if (field === BrsInvalid.Instance) {
+            if (!this.fields.has(fieldname.value)) {
                 return BrsBoolean.False;
             }
 
@@ -477,8 +471,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
 
             fields.getValue().forEach((value, key) => {
                 let fieldName = new BrsString(key);
-                let field = this.get(fieldName);
-                if (field !== BrsInvalid.Instance) {
+                if (this.fields.has(key)) {
                     this.set(fieldName, value);
                 }
             });
@@ -501,8 +494,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
 
             aa.getValue().forEach((value, key) => {
                 let fieldName = new BrsString(key);
-                let field = this.get(fieldName);
-                if (field !== BrsInvalid.Instance) {
+                if (this.fields.has(key)) {
                     this.set(fieldName, value);
                 }
             });
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -1,4 +1,4 @@
-import { BrsValue, ValueKind, BrsString, BrsInvalid, BrsBoolean } from "../BrsType";
+import { BrsValue, ValueKind, BrsString, BrsInvalid, BrsBoolean, Uninitialized } from "../BrsType";
 import { BrsComponent, BrsIterable } from "./BrsComponent";
 import { BrsType } from "..";
 import { Callable, StdlibArgument } from "../Callable";
@@ -42,6 +42,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             this.addfield,
             this.addfields,
             this.getfield,
+            this.setfield,
         ]);
     }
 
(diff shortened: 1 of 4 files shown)


src/brsTypes/components/RoSGNode.ts
```diff
@@ -7,22 +7,59 @@ import { Int32 } from "../Int32";
...
+import { Call } from "../../parser/Expression";
+import { objectExpression } from "@babel/types";
 
 class Field {
-    constructor(readonly type: string, readonly alwaysNotify: boolean) {}
+    private type: string;
+    private value: BrsType;
+    private observers = new Array<BrsString>();
+
+    constructor(value: BrsType, alwaysNotify?: boolean) {
```
strattonbrazil: Should alwaysNotify be nullable at this level instead of handling it outside the field?  Maybe use a default constructor?  Looks like we're not using it yet, but it would be better if whatever eventually uses it doesn't need to do a null check every time.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -96,18 +146,19 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+        return result || this.getMethod(index.value) || BrsInvalid.Instance;
     }
 
-    set(index: BrsType, value: BrsType) {
+    set(index: BrsType, value: BrsType, alwaysNotify?: boolean) {
         if (index.kind !== ValueKind.String) {
             throw new Error("RoSGNode indexes must be strings");
         }
-        this.elements.set(index.value.toLowerCase(), value);
+        this.fields.set(index.value.toLowerCase(), new Field(value, alwaysNotify));
```
strattonbrazil: Maybe do the null check here.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -231,6 +325,181 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+    });
+
+    /** Adds one or more fields defined as an associative aray of key values. */
+    private addfields = new Callable("addfields", {
+        signature: {
+            args: [new StdlibArgument("fields", ValueKind.Object)],
+            returns: ValueKind.Boolean,
+        },
+        impl: (interpreter: Interpreter, fields: RoAssociativeArray) => {
+            if (!(fields instanceof RoAssociativeArray)) {
```
strattonbrazil: What's the point of this check?  Doesn't the RoAssociativeArray type in the parameter signature prevent this case from ever occurring?

src/brsTypes/components/RoSGNode.ts
```diff
@@ -231,6 +325,181 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+        },
+        impl: (interpreter: Interpreter, fieldname: BrsString, value: BrsType) => {
+            let field = this.get(fieldname);
+
+            if (field === BrsInvalid.Instance) {
+                return BrsBoolean.False;
+            }
+
+            if (
+                ValueKind.toString(field.kind).toLowerCase() !==
```
strattonbrazil: When would these kinds get different casings?  Do they not all come from the same type-safe source?

src/brsTypes/components/RoSGNode.ts
```diff
@@ -231,6 +325,181 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+    });
+
+    /** Updates the value of multiple existing field only if the types match. */
+    private setfields = new Callable("setfields", {
+        signature: {
+            args: [new StdlibArgument("fields", ValueKind.Object)],
+            returns: ValueKind.Boolean,
+        },
+        impl: (interpreter: Interpreter, fields: RoAssociativeArray) => {
+            if (!(fields instanceof RoAssociativeArray)) {
```
strattonbrazil: Same question as above.  Is this check necessary?

src/brsTypes/components/RoSGNode.ts
```diff
@@ -231,6 +325,181 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+                return BrsBoolean.False;
+            }
+
+            fields.getValue().forEach((value, key) => {
+                let fieldName = new BrsString(key);
+                let field = this.get(fieldName);
+                let valueType = ValueKind.toString(value.kind);
+                if (
+                    field !== BrsInvalid.Instance &&
+                    ValueKind.toString(field.kind).toLowerCase() === valueType.toLowerCase()
```
strattonbrazil: Same question as above.

test/brsTypes/components/RoSGNode.test.js
```diff
@@ -235,22 +244,36 @@ describe("RoSGNode", () => {
...
+
                 let node = new RoSGNode([]);
 
                 let keys = node.getMethod("keys");
                 expect(keys).toBeTruthy();
 
                 let result = keys.call(interpreter);
-                expect(result.elements).toEqual(new RoArray([]).elements);
+                expect(result.getElements()).toEqual(
+                    new RoArray([change, focusable, focusedChild, id]).elements
```
strattonbrazil: Does "new RoArray([...]).elements" wrap the elements in an array and then immediately unwrap this?  Why can't we just pass in the array itself to toEqual()?

test/brsTypes/components/RoSGNode.test.js
```diff
@@ -264,17 +287,385 @@ describe("RoSGNode", () => {
                 let items = node.getMethod("items");
                 expect(items).toBeTruthy();
                 let result = items.call(interpreter);
-                expect(result.elements).toEqual(new RoArray([letter1, letter2, cletter]).elements);
+                expect(result.getElements()).toEqual(
+                    new RoArray([id, focusedChild, letter1, letter2, cletter, change, focusable])
```
strattonbrazil: Same question.

test/brsTypes/components/RoSGNode.test.js
```diff
@@ -264,17 +287,385 @@ describe("RoSGNode", () => {
...
+                expect(node.getFields().size).toEqual(6);
+
+                let moreFields = new RoAssociativeArray([
+                    { name: new BrsString("field1"), value: new BrsString("my string") },
+                    { name: new BrsString("field2"), value: new Int32(-10) },
+                    { name: new BrsString("field3"), value: BrsInvalid.Instance },
+                ]);
+                result = addFields.call(interpreter, moreFields);
+                expect(result).toEqual(BrsBoolean.True);
+                expect(node.getFields().size).toEqual(7); //Only adds the non duplicated
```
gincwang: nice!

src/brsTypes/components/RoSGNode.ts
sjbarag: Good point!  We could set a default value with `constructor(value: BrsType, alwaysNotify: boolean = false)`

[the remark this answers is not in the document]
strattonbrazil: It can be saved by the constructor if it has a scope setting (https://kendaleiv.com/typescript-constructor-assignment-public-and-private-keywords/).  This should probably be changed to "readonly alwaysNotify: boolean" so it's publicly accessible, but nothing can change it.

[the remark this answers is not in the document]
alimnios72: I'm declaring the argument as private to use it later when the field changes. I've change the `set` implementation to have a default value for this.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -96,18 +146,19 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         // method with the desired name separately? That last bit would work but it's pretty gross.
         // That'd allow roArrays to have methods with the methods not accessible via `arr["count"]`.
         // Same with RoAssociativeArrays I guess.
-        return (
-            this.elements.get(index.value.toLowerCase()) ||
-            this.getMethod(index.value) ||
-            BrsInvalid.Instance
-        );
+        let field = this.fields.get(index.value.toLowerCase());
+        let result;
```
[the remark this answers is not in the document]
alimnios72: I've removed this var to make things simpler

src/brsTypes/components/RoSGNode.ts
alimnios72: I've changed the implementation of this, it should have the check now

src/brsTypes/components/RoSGNode.ts
```diff
@@ -119,26 +170,61 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+            default:
+                value = Uninitialized.Instance;
+                break;
+        }
+
+        return value;
+    }
+
     /** Removes all elements from the node */
+    // ToDo: Built-in fields shouldn't be removed
```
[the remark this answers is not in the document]
alimnios72: Will update!

src/brsTypes/components/RoSGNode.ts
```diff
@@ -189,8 +275,16 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
         impl: (interpreter: Interpreter, obj: BrsType) => {
-            if (obj instanceof RoAssociativeArray || obj instanceof RoSGNode) {
-                this.elements = new Map<string, BrsType>([...this.elements, ...obj.elements]);
+            if (obj instanceof RoAssociativeArray) {
+                obj.elements.forEach((value, key) => {
+                    this.fields.set(key, new Field(value));
+                });
+            }
+
+            if (obj instanceof RoSGNode) {
```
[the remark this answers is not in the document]
alimnios72: sure, why not

src/brsTypes/components/RoSGNode.ts
alimnios72: It will for the unit tests but won't prevent actual brightscript code to pass random stuff. Also the box returns false in this case so I'm keeping this for consistency

src/brsTypes/components/RoSGNode.ts
alimnios72: good call 🤦‍♂ , updating this

src/brsTypes/components/RoSGNode.ts
alimnios72: Keep just for consistency with the actual device

src/brsTypes/components/RoSGNode.ts
alimnios72: Updating

test/brsTypes/components/RoSGNode.test.js
alimnios72: True

alimnios72:
@sjbarag missed your feedback until just now, I think most of your concerns are fixed already but gonna double check tomorrow!

src/brsTypes/components/RoSGNode.ts
```diff
@@ -231,6 +325,181 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+
+    /** Removes the given field from the node */
+    /** TODO: node built-in fields shouldn't be removable (i.e. id, change, focusable,) */
+    private removefield = new Callable("removefield", {
+        signature: {
+            args: [new StdlibArgument("fieldname", ValueKind.String)],
+            returns: ValueKind.Boolean,
+        },
+        impl: (interpreter: Interpreter, fieldname: BrsString) => {
+            if (this.get(fieldname) !== BrsInvalid.Instance) {
```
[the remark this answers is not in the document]
alimnios72: Good catch, I'll fix this and will add a test case for that scenario

src/brsTypes/components/RoSGNode.ts
```diff
@@ -8,15 +8,13 @@ import { RoAssociativeArray } from "./RoAssociativeArray";
...
 import { AAMember } from "./RoAssociativeArray";
 import { Float } from "../Float";
-import { Call } from "../../parser/Expression";
-import { objectExpression } from "@babel/types";
 
 class Field {
     private type: string;
     private value: BrsType;
-    private observers = new Array<BrsString>();
+    private observers: String[] = [];
```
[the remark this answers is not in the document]
alimnios72: Done!

src/brsTypes/components/RoSGNode.ts
```diff
@@ -119,26 +170,61 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
         },
         impl: (interpreter: Interpreter) => {
-            this.elements.clear();
+            this.fields.clear();
             return BrsInvalid.Instance;
         },
     });
 
     /** Removes a given item from the node */
+    // ToDo: Built-in fields shouldn't be removed
```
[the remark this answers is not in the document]
alimnios72: That's right, I'll leave it as a todo for now so I can focus on the `observeField` method.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -231,6 +325,181 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+            returns: ValueKind.Boolean,
+        },
+        impl: (interpreter: Interpreter, fieldname: BrsString) => {
+            return BrsBoolean.True;
+        },
+    });
+
+    /** Removes the given field from the node */
+    /** TODO: node built-in fields shouldn't be removable (i.e. id, change, focusable,) */
+    private removefield = new Callable("removefield", {
```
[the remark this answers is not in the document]
alimnios72: Just realized they are exactly the same, neither of them ever return false if the field isn't removed. Will add a note here

PR #267, opened 2019-06-21
Implement if sg node dict

underwoo16:
This implements the `findNode()` function from the `ifSGNodeDict` interface documented here:
https://developer.roku.com/docs/references/brightscript/interfaces/ifsgnodedict.md

This PR will need to be updated after #265 has been merged.


src/brsTypes/components/RoSGNode.ts
```diff
@@ -119,6 +121,25 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
 
+    findNodeById(node: RoSGNode, id: BrsString): RoSGNode | BrsInvalid {
+        // test current node in tree
+        let currentId = node.get(new BrsString("id"));
+        if (currentId.toString() === id.toString()) {
+            return node;
+        }
+
+        // visit each child
+        for (let child of node.children) {
```
alimnios72: Isn't this DFS? I think it's called out in the docs that BFS is used.

test/brsTypes/components/RoSGNode.test.js
```diff
@@ -578,4 +578,130 @@ describe("RoSGNode", () => {
...
+            });
+
+            it("finds a node in its direct children", () => {
+                let appendChild = node.getMethod("appendchild");
+                let children = [child1, child2, child3, child4];
+                for (let child of children) {
+                    appendChild.call(interpreter, child);
+                }
+
+                let findNode = node.getMethod("findnode");
```
alimnios72: Don't think this check is needed here and below since you already did it in the first test case

src/brsTypes/components/RoSGNode.ts
```diff
@@ -119,6 +121,25 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         this.parent = BrsInvalid.Instance;
     }
 
+    findNodeById(node: RoSGNode, id: BrsString): RoSGNode | BrsInvalid {
+        // test current node in tree
+        let currentId = node.get(new BrsString("id"));
+        if (currentId.toString() === id.toString()) {
```
alimnios72: I think we should enforce the `id` field to alway be of `BrsString` type otherwise this would fail.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -344,6 +365,25 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             return child;
         },
     });
+
+    /* Returns the node that is a descendant of the nearest component ancestor of the subject node whose id field matches the given name,
+        otherwise return invalid.*/
```
gincwang: can you just add a comment that it's DFS instead of the doc's BFS?

alimnios72: https://developer.roku.com/docs/references/brightscript/interfaces/ifsgnodedict.md#findnodename-as-string-as-object

> The search for the descendant node is a breadth-first search that includes child nodes in nodes that are declared as custom components defined in other XML component files

src/brsTypes/components/RoSGNode.ts
underwoo16: yes this is DFS. This implementation matches the behavior seen on device.

src/brsTypes/components/RoSGNode.ts
underwoo16: will do. went through a bunch of scenarios with Sean yesterday and we feel pretty confident it is DFS.

src/brsTypes/components/RoSGNode.ts
underwoo16: Chatted offline and decided since all `BrsTypes` have `toString` method this will not crash (but it may not do what you are expecting if you set `id = someNonStringValue`)

src/brsTypes/components/RoSGNode.ts
alimnios72: Docs wrong again, thanks for taking the time of checking this on the box

PR #268, opened 2019-06-21
feat(stdlib): Implement observeField method

alimnios72:
This PR adds support for the `observeField` method defined in the `ifSGNodeField` interface by executing every named function attached to an object field.


src/brsTypes/components/RoSGNode.ts
```diff
@@ -32,18 +36,23 @@ class Field {
...
+        let brsCallback: BrsCallback = {
+            interpreter: interpreter,
+            functionName: functionName.value,
+        };
+        this.observers.push(brsCallback);
     }
 
-    private executeCallbacks() {}
+    private executeCallbacks(callback: BrsCallback) {
+        callback.interpreter.callNamedFunction(callback.functionName);
```
gincwang: 🆒 🍻

src/interpreter/index.ts
```diff
@@ -184,6 +184,33 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
         }
     }
 
+    callNamedFunction(functionName: string) {
```
gincwang: Can you parking lot how this actually works for us?  Does this just find the first callback function with that exact function name?  What if different components have the same callback name?

underwoo16: I'm still concerned / confused about how this will find the correct named function for more complex components but this looks sweet for single files.

alimnios72: As synced offline, the `RoSGNode` will iterate over all the observers attached and will call each function by name using the reference to the interpreter.

sjbarag: Great point!  It almost seems like the callable gets found immediately when `observeField` is called, and the list of observers winds up being a reference to each of those?

src/interpreter/index.ts
```diff
@@ -185,6 +185,33 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
...
+        try {
+            let callbackVariable = new Expr.Variable({
+                kind: Lexeme.Identifier,
+                text: functionName,
+                isReserved: false,
+                location: {
```
sjbarag: 🤔 one day I'll pull this out into its own variable 😃

src/interpreter/index.ts
```diff
@@ -185,6 +185,33 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
...
+                    },
+                    file: "(internal)",
+                },
+            });
+            let maybeCallback = this.evaluate(callbackVariable);
+            if (maybeCallback.kind === ValueKind.Callable) {
+                this.visitCall(new Expr.Call(callbackVariable, callbackVariable.name, []));
+            }
+        } catch (err) {
+            throw err;
```
sjbarag: It feels a little weird to try-catch-throw here without doing using `err` at all — would it be reasonable to remove the `try/catch` entirely?

alimnios72: You're right, it doesn't make any sense to just throw the error, gonna remove this.

PR #284, opened 2019-08-08
Implement ifSGNodeChildren and ifSGNodeDict interfaces

underwoo16:
As well as a function for mocking components like so:
 `_brs_.mockComponent("COMPONENT_NAME", {MOCK_OBJECT_TO_RETURN})`

[ifSGNodeChildren](https://developer.roku.com/en-gb/docs/references/brightscript/interfaces/ifsgnodechildren.md) and [ifSGNodeDict](https://developer.roku.com/en-gb/docs/references/brightscript/interfaces/ifsgnodedict.md)


src/interpreter/Environment.ts
```diff
@@ -174,11 +190,30 @@ export class Environment {
...
         newEnvironment.focusedNode = this.focusedNode;
 
         return newEnvironment;
     }
 
+    /**
+     * retrieves mocked object if it exists
+     * @param objName the object to mock
+     */
+    public getMock(objName: string): BrsType {
```
sjbarag: Any particular reason to store this in `Environment`?  I suspect we could use a similar `Map<string, RoAssociativeArray>` stored in `src/mocks/RegisteredMocks.ts` or something, and the `CreateObject` implementation just grabs it from `(up a billion levels)/mocks/RegisteredMocks`

test/e2e/resources/mocks.brs
```diff
@@ -0,0 +1,60 @@
+sub Main()
+    ' testing mock objects with functions
+    _brs_.mockComponent("roTimespan", {
```
sjbarag: 😭 it's so beautiful! ❤️

src/interpreter/Environment.ts
sjbarag: Hmm, interestingly my first at doing what I suggested didn't work sooooo 🤷‍♂ this works!

underwoo16:
@sjbarag I just interpreted the ticket as "create a separate scope for mocking objects" so I did it that way, no particular reason it wasn't in global

PR #286, opened 2019-08-11
Fix lexer.isToken

TwitchBronBron:
`lexer.isToken` is supposed to return a boolean, but it actually returns the last nonfalsey value it finds. If statement that actually compared it to `true`|`false` would result in unexpected behavior.


test/lexer/Lexer.test.js
```diff
@@ -422,4 +422,23 @@ describe("lexer", () => {
             ]);
         });
     });
+
+    describe("isToken", () => {
+        it("works", () => {
```
sjbarag: Mind changing this to `"returns booleans"` or something?

```suggestion
        it("returns booleans", () => {
```

TwitchBronBron:
And no rush on getting this released...I'm not actually depending on it for anything. I was just writing some extra unit tests in my copy of `brs` in `brighterscript`, and when I went to test the method, that's how I found the issue.

PR #290, opened 2019-08-20
Added "quit" and "exit" cli statements and fixed "invalid" message  #272 #273

### LINKED ISSUE #272: An "invalid" text is shown after any print statement
For any ? or print statement you get the correct result but an "invalid" is also shown in the very next line.
This is happening in 0.14.0
Tested both on Windows 10 and Elementary OS Juno (Ubuntu)

### LINKED ISSUE #273: Suggestion: Use "quit"  statement as an alternative to Ctrl+D
Roku supports the quit statement during a telnet debug session to close the channel and debug session. I think this would be nice to have in brs command line interface.

lvcabral:
fixes #272 
fixes #273

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -30,6 +30,8 @@ program
                 process.exitCode = 1;
             }
         } else {
+            console.log("Off-Roku BrightScript interpreter [Version " + packageJson.version + "]");
+            console.log("");
             brs.repl();
         }
     })
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -30,7 +30,7 @@ program
                 process.exitCode = 1;
             }
         } else {
-            console.log("Off-Roku BrightScript interpreter [Version " + packageJson.version + "]");
+            console.log(`Off-Roku BrightScript interpreter [Version ${packageJson.version}]`);
             console.log("");
             brs.repl();
         }
--- a/src/index.ts
+++ b/src/index.ts
@@ -152,7 +152,7 @@ export function repl() {
         let results = run(line, defaultExecutionOptions, replInterpreter);
         if (results) {
             results.map(result => {
-                if (result.toString() !== "invalid") {
+                if (result !== BrsTypes.BrsInvalid.Instance) {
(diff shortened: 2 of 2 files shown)


src/index.ts
```diff
@@ -145,13 +145,18 @@ export function repl() {
...
-
     rl.on("line", line => {
+        if (line.toLowerCase() === "quit" || line.toLowerCase() === "exit") {
+            process.exit();
+        }
         let results = run(line, defaultExecutionOptions, replInterpreter);
         if (results) {
-            results.map(result => console.log(result.toString()));
+            results.map(result => {
+                if (result.toString() !== "invalid") {
```
TwitchBronBron: What about when someone runs something like this:
```brightscript
name = invalid
print name
```

lvcabral: It works, try it, this results object is not the actual response on the REPL.

lvcabral: image

TwitchBronBron: Cool, thanks for verifying!

bin/cli.js
```diff
@@ -30,6 +30,8 @@ program
                 process.exitCode = 1;
             }
         } else {
+            console.log("Off-Roku BrightScript interpreter [Version " + packageJson.version + "]");
```
sjbarag: Great addition!  Consider using a [template literal](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Template_literals#Expression_interpolation) instead of concatenating strings, e.g.

```suggestion
            console.log(`Off-Roku BrightScript interpreter [Version ${packageJson.version}`);
```

src/index.ts
sjbarag: Rather than comparing strings, could we compare against the `BrsInvalid.Instance` singleton?   Something like

```suggestion
                if (result !== BrsTypes.BrsInvalid.Instance) {
```

appears to work the same way and is more explicit (it also doesn't convert to a string twice 😉)

lvcabral:
Changes done

PR #291, opened 2019-08-21
Functions and constructors to auto-convert between Int and Float

### LINKED ISSUE #280: When an integer is passed to a method on a float parameter it should be converted to float.
Currently the interpreter is raising an exception if you send an integer number to a method that expects a float. 
Roku does convert that number to float transparently.

lvcabral:
Roku does truncate float to integer (not round) so I also changed the constructor of Int32 and Int64 to correctly mimic the behavior.

fixes #280

### DIFF
--- a/src/brsTypes/Callable.ts
+++ b/src/brsTypes/Callable.ts
@@ -3,6 +3,8 @@ import * as Brs from ".";
 import * as Expr from "../parser/Expression";
 import { Scope } from "../interpreter/Environment";
 import { Location } from "../lexer";
+import { Int32 } from "./Int32";
+import { Float } from "./Float";
 
 /** An argument to a BrightScript `function` or `sub`. */
 export interface Argument {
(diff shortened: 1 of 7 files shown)


src/brsTypes/Callable.ts
```diff
@@ -255,6 +257,24 @@ export class Callable implements Brs.BrsValue {
                 return;
             }
 
+            if (
+                expected.type.kind === Brs.ValueKind.Float &&
+                received.kind === Brs.ValueKind.Int32
+            ) {
+                let old = args[index] as Int32;
```
sjbarag: Consider using the `received` variable in place of this new `old` one — the TypeScript compiler already knows it's an Int32 because of the `ValueKind` comparison on line 262, so there's no need to re-cast it here!

test/brsTypes/Int64.test.js
```diff
@@ -11,18 +11,18 @@ const {
 const Long = require("long");
 
 describe("Int64", () => {
-    it("rounds initial values to integers", () => {
+    it("truncates initial values to integers", () => {
         let three = new Int64(3.4);
         let four = new Int64(3.5);
```
sjbarag: Would you mind renaming the variable `four` to `threePointFive` or something?   Or drop the intermediate values entirely, since both `3.4` and `3.5` should convert to `3` either way!  The meaning of the variable names is no longer useful, since they have the same resulting value 😄

lvcabral:
Updated with suggestion changes

Issue #310, opened 2019-09-02
for loop with final < initial and step > 0 should not process anything


lvcabral:
This code should not print anything, but the interpreter is getting into loop until integer overflow

```
'ensure do not loop if final is lower than initial (and step > 0)
for a = 0 to -1
    print "this should not be printed"
next
```

lvcabral:
Same scenario if the step is negative and the start < finish
`
'ensure do not loop if final is higher than initial (and step < 0)
for a = 1 to 10 step -1
    print "this should not be printed"
    exit for
next
`

PR #311, opened 2019-09-02
Fixed `for loop` issue that should not process loop when step is on the wrong direction of the start-end parameters #310

lvcabral:


lvcabral:
Added both fixes for #310 and #315 on the same PR

Issue #309, opened 2019-09-01
Interpreter messes up nested "if" handling when there is an "one line if" next to "else" or "endif"


lvcabral:
if type(component.TextAttrs.font) = "roString"
            font = m.fonts.Lookup(component.TextAttrs.font)
            if font = invalid then font = m.fonts.medium
        else if type(component.TextAttrs.font) = "roFont"
            font = component.TextAttrs.font
        else
            font = m.fonts.reg.GetDefaultFont()
        end if

lvcabral:
In the code above, the line `if font=invalid then font=m.fonts.medium` messes up the interpreter (with not exception)  and the logic implies the `else` below is related to the `inline if`

strattonbrazil:
Fixed in sjbarag/brs#309.

sjbarag:
> Fixed in #309.

Looks like #345 fixed it — this is 309 😜

sjbarag:
Thanks for the PR, @lvcabral!

Issue #253, opened 2019-06-06
Single line if else with dot setter


triwav:
`if true then m.top.visible = true else m.top.visible = false` spits out `Expected newline or ':' after dotted 'set' statement`

sjbarag:
Well that's odd :open_mouth: Thanks for providing a test case! :heart:

lvcabral:
I have a fix for this issue, will send a PR soon

strattonbrazil:
Fixed in sjbarag/brs#309

Issue #252, opened 2019-06-04
Nested If Invalid Error


triwav:
The following will incorrectly spit out `Expected a statement to follow 'else'`
```
if true
    if true then t = 1
else
    ' empty line or line with just a comment causes crash
end if
```
It only happens if the inner if is a one line if and the else is empty which is edge casey but still compiles correctly so figured I'd report it.

sjbarag:
Oh bummer!  Thanks for reporting this @triwav :smile:

PR #330, opened 2019-10-04
Fixed issues with inline if (#252, #253 and #309)

lvcabral:


src/parser/Parser.ts
```diff
@@ -933,7 +933,8 @@ export class Parser {
                     });
                 }
 
-                if (match(Lexeme.Else)) {
+                if (previous().kind !== Lexeme.Newline && match(Lexeme.Else)) {
+                    console.log("Else 2");
```
TwitchBronBron: Looks like you forgot to remove this `console.log` statement.

src/parser/Parser.ts
```diff
@@ -911,7 +911,7 @@ export class Parser {
                 }
                 thenBranch = new Stmt.Block([thenStatement], peek().location);
 
-                while (match(Lexeme.ElseIf)) {
+                while (previous().kind !== Lexeme.Newline && match(Lexeme.ElseIf)) {
```
TwitchBronBron: There's already a `checkPrevious()` method that does this in a slightly less verbose way. @sjbarag might ask you to use that.

src/parser/Parser.ts
lvcabral: indeed, will fix it.

src/parser/Parser.ts
lvcabral: will take a look, thanks for the review

sjbarag:
@lvcabral could you please run a `git rebase` against the upstream `master` branch and delete any commits that don't need to be here?  There's currently 526 commits and 61,000 lines added, but something tells me that's not what you meant to do :wink:

lvcabral:
I'm closing this pull request, will resubmit with a new clean branch

PR #331, opened 2019-10-11
fix(dirs): Defer process.cwd() calls until needed

sjbarag:
Settting `process.cwd()` when the `brs` module is imported causes it to be static for the lifetime of a consuming process, and that's typically not what we want!  Fetching `process.cwd()` as part of the `brs` execution path allows the root directory for `pkg:/` paths to vary with the consuming process dynamically.


sjbarag:
Passing in a `root` property solves this, as seen in https://github.com/hulu/roca/pull/46

PR #333, opened 2019-10-16
feat(stdlib): Added XML parsing and basic support for components

alimnios72:
- Added XML parsing support to start testing Scenegraph components
- Updated RoSGNode to support custom components
- Added e2e testing


yarn.lock
```diff
@@ -299,11 +299,24 @@
...
+  dependencies:
+    "@nodelib/fs.stat" "2.0.3"
+    run-parallel "^1.1.9"
+
 "@nodelib/fs.stat@2.0.2", "@nodelib/fs.stat@^2.0.1":
   version "2.0.2"
   resolved "https://registry.yarnpkg.com/@nodelib/fs.stat/-/fs.stat-2.0.2.tgz#2762aea8fe78ea256860182dcb52d61ee4b8fda6"
   integrity sha512-z8+wGWV2dgUhLqrtRYa03yDx4HWMvXKi1z8g3m2JyxAx8F7xk74asqPk5LAETjqDSGLFML/6CDl0+yFunSYicw==
 
+"@nodelib/fs.stat@2.0.3", "@nodelib/fs.stat@^2.0.2":
```
alimnios72: I had some issues when merging, does this seems good to you?

yarn.lock
```diff
@@ -299,11 +299,24 @@
     "@nodelib/fs.stat" "2.0.2"
     run-parallel "^1.1.9"
 
+"@nodelib/fs.scandir@2.1.3":
```
alimnios72: I had some issues when merging, does this seems good to you?

src/brsTypes/BrsType.ts
```diff
@@ -90,6 +92,48 @@ export namespace ValueKind {
...
+
+    switch (type.toLowerCase()) {
+        case "boolean":
+            returnValue = value ? BrsBoolean.from(value === "true") : BrsBoolean.False;
+            break;
+        case "dynamic":
+            returnValue = BrsInvalid.Instance;
+            break;
+        case "integer":
+            returnValue = value ? new Int32(Number.parseInt(value)) : new Int32(0);
```
underwoo16: I assume the parser will handle cases where value may not be a valid number? So it should be caught before this point I think..

src/brsTypes/BrsType.ts
```diff
@@ -90,6 +92,48 @@ export namespace ValueKind {
...
+ */
+export function getBrsTypeFromString(type: string, value?: string): BrsType {
+    let returnValue: BrsType;
+
+    switch (type.toLowerCase()) {
+        case "boolean":
+            returnValue = value ? BrsBoolean.from(value === "true") : BrsBoolean.False;
+            break;
+        case "dynamic":
+            returnValue = BrsInvalid.Instance;
```
underwoo16: Is it valid to define a type as `dynamic` in scenegraph xml? I thought that was only in brightscript code

src/componentprocessor/index.ts
```diff
@@ -0,0 +1,142 @@
...
+            this.xmlNode = new xmldoc.XmlDocument(xmlStr);
+            this.name = this.xmlNode.attr.name;
+
+            return Promise.resolve(this);
+        } catch (err) {
+            // TODO: provide better parse error reporting
+            //   cases:
+            //     * file read error
+            //     * XML parse error
+            return Promise.reject(this);
```
underwoo16: Would printing out the `err` at least provide some info on what went wrong?

src/index.ts
```diff
@@ -34,6 +35,8 @@ export async function execute(filenames: string[], options: Partial<ExecutionOpt
 
     let manifest = await PP.getManifest(executionOptions.root);
 
+    let nodeDefs = await getComponentDefinitionMap(executionOptions.root);
```
underwoo16: should this be optional in case people want to run `brs` on source files only and don't have any components to parse or don't want to parse components?

src/brsTypes/BrsType.ts
alimnios72: You're right, `dynamic` is not a valid type [according to the docs](https://developer.roku.com/docs/references/scenegraph/xml-elements/interface.md) but on the other hand [node](https://developer.roku.com/docs/references/scenegraph/abstract-nodes/node.md) doesn't define a type for `focusedChild` field. I think I'll update that case to accept `"invalid"` for those edge cases

yarn.lock
sjbarag: 🤷‍♂ usually just running `yarn` will resolve those conflicts 😃

src/componentprocessor/index.ts
alimnios72: It gets thinks a bit messy when running the unit tests because one of those checks for invalid xml syntax

src/brsTypes/BrsType.ts
strattonbrazil: While the docs are weird, let's change this to "node" as that seems what _focusedChild_ is in this case.  Let's also update that in the constructor of roSGNode.

src/brsTypes/BrsType.ts
strattonbrazil: This is pulling straight from the XML value attribute, so there's no brightscript parsing before this point.

src/brsTypes/BrsType.ts
```diff
@@ -90,6 +92,48 @@ export namespace ValueKind {
...
 }
 
+/**
+ *  Converts a specified brightscript type in string into BrsType representation, with actual value
+ *  Note: only supports native types so far.  Objects such as array/AA aren't handled at the moment.
+ *  @param {string} type data type of field
+ *  @param {string} value optional value specified as string
+ *  @returns {BrsType} BrsType value representation of the type
+ */
+export function getBrsTypeFromString(type: string, value?: string): BrsType {
```
strattonbrazil: Should we rename this to "getBrsValueFromFieldType" to be super explicit?

src/componentprocessor/index.ts
strattonbrazil: Let's revisit this in the future possibly having an early failure case to check all XML files.

src/index.ts
strattonbrazil: In the case of no components I believe glob should return empty.  We can explore adding an option in the future to exclude component parsing.

PR #335, opened 2019-11-01
Reordering child node parsing in component

MATATAT:
Reordered the child node parsing and added a unit test to support correct parse order.


sjbarag:
Thanks for the PR, @MATATAT , and thanks for merging it @strattonbrazil !

Issue #306, opened 2019-08-28
Make Interpreter available to component constructors


lvcabral:
As all methods do receive the interpreter instance, is also useful to have it available at the constructor too.

lvcabral:
Already fixed and merged

PR #340, opened 2019-11-12
Made interpreter available to any object constructor #306

lvcabral:


lvcabral:
> I don't expect this to be as useful for the non-node types. Do you have any future use cases in mind?

In my fork (brs-emu) I needed to use the interpreter in the constructor of:
* `roScreen`
* `roBitmap`
* `roAudioResource`
* `roUrlTransfer`

PR #342, opened 2019-11-15
Adding functions into custom component scopes

MATATAT:
This PR covers:

1. Adding individual environment scopes to custom components
2. Parsing `<script>` tags to build a tree of functions split by scope
3. Attaching these scopes to the corresponding component's environment

Note that this does not cover supporting multiple `init()` functions. This will come in the next iteration.


src/index.ts
```diff
@@ -34,73 +38,31 @@ export async function execute(filenames: string[], options: Partial<ExecutionOpt
     const executionOptions = Object.assign(defaultExecutionOptions, options);
 
     let manifest = await PP.getManifest(executionOptions.root);
+    let componentDefinitions = await getComponentDefinitionMap(executionOptions.root);
```
strattonbrazil: Does the interpreter not need any other execution options?

src/index.ts
```diff
@@ -34,73 +38,31 @@ export async function execute(filenames: string[], options: Partial<ExecutionOpt
...
-
-    // don't execute anything if there were reading, lexing, or parsing errors
-    if (parsedFiles.some(file => file.isRejected)) {
-        return Promise.reject({
-            messages: parsedFiles
-                .filter(file => file.isRejected)
-                .map(rejection => rejection.reason.message),
-        });
+    if (!interpreter) {
+        throw new Error("Unable to build interpreter with given component definitions.");
```
strattonbrazil: Is this always a component definition error?

src/interpreter/Environment.ts
```diff
@@ -188,10 +188,12 @@ export class Environment {
      *
      * @returns a copy of this environment but with no function-scoped values.
      */
-    public createSubEnvironment(): Environment {
+    public createSubEnvironment(includeModuleScope: boolean = true): Environment {
```
strattonbrazil: Can you add this parameter to the doc string so it's clear what it's doing?

src/index.ts
```diff
@@ -34,73 +38,31 @@ export async function execute(filenames: string[], options: Partial<ExecutionOpt
     const executionOptions = Object.assign(defaultExecutionOptions, options);
 
     let manifest = await PP.getManifest(executionOptions.root);
+    let componentDefinitions = await getComponentDefinitionMap(executionOptions.root);
+
+    let pathFormatter = (component: ComponentDefinition) => {
```
strattonbrazil: Declaring this function inline makes it harder to understand and test.  We should move it out or encapsulate the logic into the interpreter.

src/interpreter/index.ts
```diff
@@ -94,6 +96,45 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
...
+     * Builds out all the sub-environments for the given components. Components are saved into the calling interpreter
+     * instance. This function will mutate the state of the calling interpreter.
+     * @param componentMap Map of all components to be assigned to this interpreter
+     * @param manifest
+     * @param formatPathFn
+     * @param parseFn
+     */
+    public static async withSubEnvsFromComponents(
+        componentMap: Map<string, ComponentDefinition>,
+        formatPathFn: (comp: ComponentDefinition) => void,
```
strattonbrazil: Is this passed in just to make testing easier?  If it never changes I'd rather change it to a helper function in this file and mock it in testing.

src/interpreter/index.ts
```diff
@@ -94,6 +96,45 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
...
+        parseFn: (filenames: string[]) => Promise<Stmt.Statement[]>,
+        options: ExecutionOptions = defaultExecutionOptions
+    ) {
+        let interpreter = new Interpreter(options);
+        interpreter.onError(logError);
+
+        let componentDefinitions = (interpreter.environment.nodeDefMap = componentMap);
+        if (componentDefinitions.size > 0) {
```
strattonbrazil: This seems redundant as Promise.all() and Array.from() should handle empty arrays.

src/interpreter/index.ts
```diff
@@ -94,6 +96,45 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
...
+    public static async withSubEnvsFromComponents(
+        componentMap: Map<string, ComponentDefinition>,
+        formatPathFn: (comp: ComponentDefinition) => void,
+        parseFn: (filenames: string[]) => Promise<Stmt.Statement[]>,
+        options: ExecutionOptions = defaultExecutionOptions
+    ) {
+        let interpreter = new Interpreter(options);
+        interpreter.onError(logError);
+
+        let componentDefinitions = (interpreter.environment.nodeDefMap = componentMap);
```
strattonbrazil: Is componentMap (passed in) and componentDefinitions the same thing?  If so, why do we redeclare it?

src/index.ts
MATATAT: The component parsing? Ultimately it just needs to know the root of either where it's executing or where you tell the CLI to execute. It will scan that space for components. It doesn't need any of the I/O or other options.

src/index.ts
MATATAT: It used to be the case, but after my last minor refactor this is no longer true. I'll update this to indicate a problem building the interpreter in general.

src/interpreter/index.ts
MATATAT: There were a couple things:
1. Since it relies on the root value from the options and I was previously not passing it in it made sense to use a closure to encapsulate that value.
2. I don't think this is a function that the Interpreter needs to be intimate with. Ultimately I wanted to separate the concerns of reading, lexing, and parsing away from the actual interpretation of that data. 

I'd be interested in moving it into a helper function but I'd like to maintain that separation of responsibility.

src/interpreter/index.ts
MATATAT: This is some fallout from one of the refactors, I was trying to maintain the previous parameter names, but it became confusing so I renamed them. I'll remove this.

src/interpreter/index.ts
MATATAT: I ended up pulling this out and having the rewrite happen before the `Interpreter` is built.

sjbarag:
oh dang is it time to finally create a `scenegraph` label in this project?! 

wow

src/index.ts
```diff
@@ -34,73 +38,30 @@ export async function execute(filenames: string[], options: Partial<ExecutionOpt
...
 
     let manifest = await PP.getManifest(executionOptions.root);
+    let componentDefinitions = await getComponentDefinitionMap(executionOptions.root);
+
+    componentDefinitions.forEach((component: ComponentDefinition) => {
+        if (component.scripts.length < 1) return;
+        component.scripts = component.scripts.map((script: ComponentScript) => {
+            script.uri = path.join(
+                options.root ? options.root : __dirname,
+                new URL(script.uri).pathname
```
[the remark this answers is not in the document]
sjbarag: Also I see the TODO about URI validation up above, so perhaps that's where we'd handle this?

sjbarag:
Also, please note that me merging #339 earlier this morning caused a merge conflict!  Shouldn't be too difficult to resolve though 😉

test/interpreter/resources/scripts/baseComponent.brs
```diff
@@ -0,0 +1,4 @@
+' sub init()
```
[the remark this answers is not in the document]
MATATAT: You are correct, and I can add that `TODO`.

src/LexerParser.ts
```diff
@@ -0,0 +1,79 @@
...
+import { promisify } from "util";
+import pSettle from "p-settle";
+const readFile = promisify(fs.readFile);
+
+import { Lexer } from "./lexer";
+import * as PP from "./preprocessor";
+import { Parser } from "./parser";
+
+import * as _lexer from "./lexer";
+export { _lexer as lexer };
```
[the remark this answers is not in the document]
MATATAT: Yeah, this is really weird... I didn't notice this when I was migrating this code. Maybe VS Code added it? Either way will fix up.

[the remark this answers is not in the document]
sjbarag: 0_o probably some auto-import something or other?  How odd!

src/index.ts
```diff
@@ -34,73 +38,30 @@ export async function execute(filenames: string[], options: Partial<ExecutionOpt
     const executionOptions = Object.assign(defaultExecutionOptions, options);
 
     let manifest = await PP.getManifest(executionOptions.root);
+    let componentDefinitions = await getComponentDefinitionMap(executionOptions.root);
+
+    componentDefinitions.forEach((component: ComponentDefinition) => {
+        if (component.scripts.length < 1) return;
```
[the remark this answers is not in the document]
MATATAT: So... this one is a weird situation that probably needs to be refactored some but I didn't want to work on it immediately. If you think it's worth fixing now I can go back and try and rework it. Otherwise we can note it and come back later.

Basically we have a mutation going on here for each component. If there are scripts present we need to modify the current component's set of paths to be correct. If it doesn't have any scripts then we just want to skip over it. I don't really like the fact that we're mutating the paths here but it works for this situation.

src/index.ts
```diff
@@ -34,73 +38,30 @@ export async function execute(filenames: string[], options: Partial<ExecutionOpt
...
+        component.scripts = component.scripts.map((script: ComponentScript) => {
+            script.uri = path.join(
+                options.root ? options.root : __dirname,
```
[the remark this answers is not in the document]
MATATAT: I think I opted to use `__dirname` over `process.cwd()` since you could run this from a different directory. It seems like that would be wrong choice in that situation?

src/index.ts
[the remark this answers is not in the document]
MATATAT: Yeah, we might want to use a `try/catch` here for now and revisit this as we refactor this piece of code. I think it's got a lot of concerns and deserves some separate time for being considered.

test/interpreter/Interpreter.test.js
```diff
@@ -0,0 +1,62 @@
...
+        );
+
+        let baseComp = interpreter.environment.nodeDefMap.get("BaseComponent");
+        expect(baseComp).not.toBeUndefined();
+        expect(baseComp.environment.module.size).toEqual(0);
+
+        let extendedComp = interpreter.environment.nodeDefMap.get("ExtendedComponent");
+        expect(extendedComp).not.toBeUndefined();
+        let actualFns = Array.from(extendedComp.environment.module).map(methodKV => {
+            let [fnNamePair] = methodKV;
```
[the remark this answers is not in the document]
MATATAT: Ah yeah, there was a bit of churn here as I was figuring out how to destructure an array.

MATATAT:
Have updated a decent amount based on feedback. I've tried to separate each commit out into individual concerns to try and help match to feedback and hopefully make re-review quicker.

PR #343, opened 2019-11-19
Splitting uri rewrite out

MATATAT:


MATATAT:
I created this by accident. It should be deleted. Sorry!

PR #344, opened 2019-12-01
Fixed #312 The scope for object m in global functions now references the root object

### LINKED ISSUE #312: Global functions called inside object methods are sharing the object scope.
If a method is part of an object like objA = { funcA: function_a } when you call objA.funcA() inside it the m. object is actually objA but if you call a global function that is not a property of objA then the m. shall still be the global AA.

The sample code  below exemplifies the issue:
```
Sub Main()
    m.someValue = "root"
    objA = { myMethod: do_something, someValue: "not root" }
    print objA.myMethod()
End Sub

function do_something()
    return GenericFunction()
end function

Function GenericFunction()
    return m.someValue
End Function

```
In Roku this code prints "root", in `brs` it prints "not root"

lvcabral:


test/e2e/resources/function/m-pointer-func.brs
```diff
@@ -0,0 +1,13 @@
+Sub Main()
+    m.someValue = "root"
+    objA = { myMethod: do_something, someValue: "not root" }    
+    print objA.myMethod()
+End Sub
+
+function do_something()
+    return GenericFunction()
```
strattonbrazil: Can you print out m.someValue here as well so we're testing both the local and root scope?  I tried from your branch and that worked fine.  

```
diff --git a/test/e2e/Functions.test.js b/test/e2e/Functions.test.js
index 8f56a5b..e0e93b2 100644
--- a/test/e2e/Functions.test.js
+++ b/test/e2e/Functions.test.js
@@ -98,7 +98,10 @@ describe("end to end functions", () => {
     test("function/m-pointer-func.brs", async () => {
         await execute([resourceFile("function", "m-pointer-func.brs")], outputStreams);
 
-        expect(allArgs(outputStreams.stdout.write).filter(arg => arg !== "\n")).toEqual(["root"]);
+            "not root",
+            "root"
+        ]);
     });
 
     test("function/scoping.brs", async () => {
diff --git a/test/e2e/resources/function/m-pointer-func.brs b/test/e2e/resources/function/m-pointer-func.brs
index f27f467..a745791 100644
--- a/test/e2e/resources/function/m-pointer-func.brs
+++ b/test/e2e/resources/function/m-pointer-func.brs
@@ -5,6 +5,7 @@ Sub Main()
 End Sub
 
 function do_something()
+    print m.someValue
     return GenericFunction()
 end function
```

lvcabral: Will add it, thanks for the feedback!

PR #350, opened 2019-12-18
feat(rsg, components): Call init methods when creating components

alimnios72:
These code changes call every `init` method found when creating a component. The calls are made in the following way:
- Init methods found in the children of the component
- Init method of the current component

also this is done starting from the root node in the hierarchy in case that the created component has any extensions.
`non-init` methods are called in the context of the component being created because of the flattening process in #347

### DIFF
--- a/src/componentprocessor/index.ts
+++ b/src/componentprocessor/index.ts
@@ -146,7 +146,7 @@ async function processXmlTree(
             while (baseNode) {
                 let baseNodeDef = nodeDefMap.get(baseNode);
                 if (baseNodeDef) {
-                    nodeDef.children = [...getChildren(baseNodeDef.xmlNode!), ...nodeDef.children];
+                    nodeDef.children = [...nodeDef.children];
                     baseNode = baseNodeDef.xmlNode!.attr.extends;
                 }
             }
--- a/test/componentprocessor/componentprocessor.test.js
+++ b/test/componentprocessor/componentprocessor.test.js
@@ -82,7 +82,7 @@ describe.only("component parsing support", () => {
                 expect(parsedExtendedComp.children).not.toBeUndefined();
                 expect(parsedExtendedComp.children.length).toBeGreaterThan(0);
 
-                let expectedChildOrder = ["label_a", "label_b"];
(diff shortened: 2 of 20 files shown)


src/componentprocessor/index.ts
```diff
@@ -146,7 +146,7 @@ async function processXmlTree(
             while (baseNode) {
                 let baseNodeDef = nodeDefMap.get(baseNode);
                 if (baseNodeDef) {
-                    nodeDef.children = [...getChildren(baseNodeDef.xmlNode!), ...nodeDef.children];
+                    nodeDef.children = [...nodeDef.children];
```
alimnios72: We don't need this anymore because children are now appended at node creation

test/componentprocessor/componentprocessor.test.js
```diff
@@ -82,7 +82,7 @@ describe.only("component parsing support", () => {
                 expect(parsedExtendedComp.children).not.toBeUndefined();
                 expect(parsedExtendedComp.children.length).toBeGreaterThan(0);
 
-                let expectedChildOrder = ["label_a", "label_b"];
+                let expectedChildOrder = ["label_b"];
```
alimnios72: Removed `label_a` from test because we no longer append children at parsing time

test/e2e/resources/components/AdvancedWidget.xml
```diff
@@ -1,7 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <component name="AdvancedWidget" extends="NormalWidget">
-    <script type="text/brightscript" uri="pkg:/components/scripts/AdvancedWidget.brs" />
```
alimnios72: For this test I only want to test children and fields but not scripts. Scripts and init are tested in a different set of components.

sjbarag: Great idea!  That might be worth putting in an XML comment too :)

sjbarag:
Oh dang I forgot to merge this!  Doing that now, one sec 😄

PR #352, opened 2019-12-30
feat(stdlib): Implement GetInterface

sjbarag:
Implementing [`GetInterface`](https://developer.roku.com/docs/references/brightscript/language/global-utility-functions.md#getinterfaceobject-as-object-ifname-as-string-as-interface) requires slightly better tracking of which interfaces are implemented on a BrightScript type — specifically the inclusion of a new `Interface` type in our BrightScript implementation.  A later PR to implement [`FindMemberFunction`](https://developer.roku.com/docs/references/brightscript/language/global-utility-functions.md#findmemberfunctionobject-as-object-funname-as-string-as-interface) will require a bit more tracking, so this PR sets that up as well.

Almost all of the lines in this diff come from 4723b6c, which updates test snapshots after inserting a new `ValueKind` at the front of the enum.  Perhaps that was a bad idea, but c'est la vie 🤷‍♂


sjbarag:
(rebased on `master`; nothing to see here)

alimnios72:
This is awesome

PR #354, opened 2019-12-31
fix(rsg): Prevent infinite loop when a node extends an unknown component

sjbarag:
Long-term we probably don't want to warn-and-fall-back for components that extend something we don't recognize.  The reference implementation considers that a compilation failure, but since we don't yet know about every component in the reference implementation we can't make that assertion!

### DIFF
--- a/src/componentprocessor/index.ts
+++ b/src/componentprocessor/index.ts
@@ -148,6 +148,14 @@ async function processXmlTree(
                 if (baseNodeDef) {
                     nodeDef.children = [...nodeDef.children];
                     baseNode = baseNodeDef.xmlNode!.attr.extends;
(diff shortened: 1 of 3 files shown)


src/componentprocessor/index.ts
```diff
@@ -148,6 +148,14 @@ async function processXmlTree(
                 if (baseNodeDef) {
                     nodeDef.children = [...nodeDef.children];
                     baseNode = baseNodeDef.xmlNode!.attr.extends;
+                } else {
+                    // The reference implementation doesn't allow extensions of unknown node subtypes, but
```
strattonbrazil: Makes sense.  Thanks for the note.

PR #360, opened 2020-01-21
fix(interp): Automatically box return values for 'as object' return types

### LINKED ISSUE #359: (Interpreter): Functions with return type `object` don't auto-box return values
RBI allows functions to declare a return type of `object` that return a value of any primitive type, and automatically boxes those primitive types to their object-like versions.  Effectively all return types are thus allowed from `as object` functions, just like with `as dynamic` (or not declaring a return type).  The only difference is that `as dynamic` (or not declaring a return type) *doesn't automatically box return values*.

```brightscript
function returnsObject_359() as object
    return "foo" ' => roString("foo")
    ' or:
    ' * return 3.14159 ' => roFloat(3.14159)
    ' * return invalid ' => roInvalid
end function

function returnsDynamic_359() as dynamic
    return "foo" ' => String("foo")
    ' or:
    ' * return 3.14159 ' => Float(3.14159)
    ' * return invalid ' => Invalid
end function

sub main()
    print returnsObject_359()
    print returnsDynamic_359()
end sub
```

Current behavior during execution:

```
/path/to/issue_359.brs(2,4-10): Attempting to return value of type String, but function returnsObject_359 declares return value of type Object
```

sjbarag:
RBI will automatically box an intrinsic value when it's returned from a function that declares an `as object` return type.  This makes the function behave quite similarly to an `as dynamic` function, in that it can seemingly return any value.  The only difference is that `as dynamic` won't automatically box something like `return false` into an `roBoolean`, while the `as object` variant will!

fixes #359

### DIFF
--- a/src/brsTypes/Boxing.ts
+++ b/src/brsTypes/Boxing.ts
@@ -1,6 +1,5 @@
 import { BrsComponent } from "./components/BrsComponent";
 import { BrsValue, BrsType, ValueKind } from "./";
-import { RoString } from "./components/RoString";
 
 export interface Boxable {
     box(): BrsComponent;
--- a/src/interpreter/index.ts
+++ b/src/interpreter/index.ts
@@ -1013,6 +1013,14 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
 
                 if (
                     returnedValue &&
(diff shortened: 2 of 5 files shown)


sjbarag:
(oh you already approved!)

PR #365, opened 2020-01-27
fix(interp): Automatically box `as object` function parameters

sjbarag:
Just like with https://github.com/sjbarag/brs/issues/359, function parameters must *also* get automatically boxed.

### DIFF
--- a/test/e2e/resources/stdlib/global-utilities.brs
+++ b/test/e2e/resources/stdlib/global-utilities.brs
@@ -1,3 +1,3 @@
 sub main()
-    print getInterface({}, "ifAssociativeArray");
+    print getInterface(1.123, "ifFloat");
 end sub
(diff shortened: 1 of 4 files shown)


test/e2e/resources/stdlib/global-utilities.brs
```diff
@@ -1,3 +1,3 @@
 sub main()
-    print getInterface({}, "ifAssociativeArray");
```
strattonbrazil: Any reason we don't want to test this as well?

sjbarag: This isn't meant to be an exhaustive check of `getInterface`'s behavior, but testing with `{}` and `"ifAssociativeArray"` didn't force the automatic-promotion behavior that was added in this PR!

PR #368, opened 2020-01-29
feat(stdlib): Implement ifSGNodeField.hasField()

### LINKED ISSUE #367: bug(SceneGraph): Implement `ifSGNode#hasField`
Looks like we missed `ifSGNode#hasField`!  Oh no 😢 

https://developer.roku.com/en-gb/docs/references/brightscript/interfaces/ifsgnodefield.md#hasfieldfieldname-as-string-as-boolean

strattonbrazil:
Implements hasField() (resolves #367 ).  The implementation is simpler than the existing _getField_ method, which returns a value, while this simply returns a boolean if the field is present.  

There is at least one bug in hasField in the official implementation that I'm aware of, where calling hasField() with specific strings will actually create the field, but this is a bug and I don't think we should match that for compatibility sake.


src/brsTypes/components/RoSGNode.ts
```diff
@@ -498,6 +499,19 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
     });
 
+    /** Returns true if the field exists */
+    private hasfield = new Callable("hasfield", {
+        signature: {
+            args: [new StdlibArgument("fieldname", ValueKind.String)],
+            returns: ValueKind.Boolean,
+        },
+        impl: (interpreter: Interpreter, fieldname: BrsString) => {
+            return this.fields.has(fieldname.value.toLowerCase())
```
sjbarag: For future reference, this could be as simple as:

```suggestion
            return BrsBoolean.from(
                this.fields.has(fieldname.value.toLowerCase())
            )
```

What you've got is of course fine though 😄

test/brsTypes/components/RoSGNode.test.js
```diff
@@ -441,6 +441,19 @@ describe("RoSGNode", () => {
             });
         });
 
+        describe("hasfield", () => {
+            it("returns presence of a field", () => {
+                let node = new RoSGNode([{ name: new BrsString("foo"), value: new Int32(17) }]);
+
+                let hasField = node.getMethod("hasfield");
+                let result1 = hasField.call(interpreter, new BrsString("foo"));
+                expect(result1).toEqual(new BrsBoolean(true));
```
sjbarag: Also for future reference, this could be a tick simpler with `expect(result1).toBe(BrsBoolean.True)`.  Because there's only the two possible `BrsBoolean` values, they're exposed as static members of `BrsBoolean` for easy access!

test/e2e/resources/components/roSGNode.brs
```diff
@@ -35,6 +35,9 @@ sub main()
     print "field1 in node now is: " node1.getField("field1")      ' => hello
     print "field3 in node now is: " node1.getField("field3")      ' => false
 
+    print "field3 present? " node1.hasField("field3")
+    print "fieldほ present? " node1.hasField("fieldほ")
```
sjbarag: Oh dang I've never thought about having non-ASCII characters in field names.  I'm glad that works! 🎉

PR #382, opened 2020-03-08
fix(manifest): Fix parsing of numbers in manifest

sjbarag:
We were previously not handling number parsing properly, with two nasty side-effects:

1. Floating-point values were truncated to integer values only (e.g. `"3.14" => 3`)
2. Strings that began with numbers were parsed as numbers and also truncated (e.g. `"1.2.3.4.5" => 1`)

Both of these stemmed from my misreading of the docs for `parseInt`; what I actually wanted was the `Number()` function followed by `Number.isNaN()`!

### DIFF
--- a/src/preprocessor/Manifest.ts
+++ b/src/preprocessor/Manifest.ts
@@ -89,7 +89,8 @@ export function parseManifest(contents: string) {
                 return [key, false];
             }
 
-            let maybeNumber = Number.parseInt(value);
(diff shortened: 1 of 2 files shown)


sjbarag:
@alimnios72 we chatted about this in-person a few days ago!  Should be fixed in the next release (publishing that in a few minutes)

alimnios72:
This is awesome, thank you

PR #393, opened 2020-03-26
bug(parse): Capture `EndIf` token when used alongside `else`

### LINKED ISSUE #392: End If token undefined when dealing with else or elseif
When parsing conditional statements, the `end if` token is sometimes undefined. This seems to happen anytime an `else` keyword is present.

TwitchBronBron:
Fixes #392


test/componentprocessor/componentprocessor.test.js
```diff
@@ -98,7 +98,7 @@ describe.only("component parsing support", () => {
 
                 const expectedPrefix = "pkg:/test/componentprocessor/resources/scripts/";
                 let expectedScripts = ["extendedComponent.brs", "utility.brs"].map(scriptName =>
-                    path.join(expectedPrefix, scriptName)
+                    path.join(expectedPrefix, scriptName).replace(/\\/g, "/")
```
TwitchBronBron: This test was failing on my Windows computer. It's unrelated to this PR, but was a one-liner, figured I would include it. I can pull it out into its own PR if you prefer.

sjbarag: Strange!  A separate PR would be lovely — I suspect there's a few cases where we use `path.join()` in an attempt to be portable when we actually want `path.posix.join()` https://nodejs.org/api/path.html#path_windows_vs_posix , and it'll be easier to review those in isolation 😄

.gitignore
```diff
@@ -3,3 +3,4 @@ lib/
 types/
 node_modules/
 .vscode/
+coverage
```
sjbarag, line 6: What produces a `coverage` file (or directory) that required this?

src/parser/Parser.ts
```diff
@@ -873,7 +873,7 @@ export class Parser {
 
                     match(Lexeme.Newline);
                     elseBranch = block(Lexeme.EndIf);
-                    let endIfToken = advance(); // skip past "end if"
+                    endIfToken = advance();
```
sjbarag, line 876: Dang, variable shadowing!  Great catch ❤️

.gitignore
TwitchBronBron: Running the coverage report using jest (`npx jest --coverage`). I undid the change.

test/componentprocessor/componentprocessor.test.js
TwitchBronBron: k. I'll submit a separate PR with just this fix in it.

TwitchBronBron:
Changes made.

Issue #313, opened 2019-09-03
Error when returning Invalid from a function that returns an Object


lvcabral:
The following code runs fine in Roku and fails in `brs`
```
Sub Main()
    print ReturnInvalid()
End Sub

Function ReturnInvalid() as Object
    return invalid
End Function
```

PR #395, opened 2020-04-09
Allow invalid returns for signatures that return object

underwoo16:
It turns out there is a `RoInvalid` type when you return `invalid` as object or even call `createObject("RoInvalid")`

This PR implements that type and adds unit tests and end to end tests covering these scenarios.

[RoInvalid](https://developer.roku.com/docs/references/brightscript/components/roinvalid.md)
fixes https://github.com/sjbarag/brs/issues/313


underwoo16:
I'm actually going to check the brightscript reference on device to see if it will return `RoInvalid`

PR #396, opened 2020-04-10
chore(deps): Update dependencies

sjbarag:


sjbarag:
Note: Not a full upgrade to the latest because there's a lot of major-version bumps available, and I'll want to upgrade those individually.  Especially [prettier](prettier.io), which changed some defaults and will have a huge diff when it's upgraded.

PR #397, opened 2020-04-11
Fix path.join bug in test when run on windows.

TwitchBronBron:
Fix a test that fails on windows devices due to using `path.join`. The test passes now by using `path.posix.join` instead as suggested in #393


TwitchBronBron:
There actually are two more failing tests on windows... I just didn't have time to dig into them.

PR #399, opened 2020-04-15
Support Mocking Functions

### LINKED ISSUE #381: Mock functions declared in source/ tree
Like #380 , it's sometimes helpful to mock functions being used by the unit being tested!  Something like `sub _brs_.mockFunction(name as string, implementation as function)` could be interesting, and might be a good starting point for #380 ?

underwoo16:
As related to Issue https://github.com/sjbarag/brs/issues/381

This problem seems complicated and this approach may be too unlimited. As it is implemented I believe one could mock any functions, even in local scope. 

I have shown in the end 2 end test how one could mock `formatJson`. This was based on the solution used to implement `_brs_.mockObject` function.

fixes #381


src/interpreter/Environment.ts
```diff
@@ -152,7 +157,11 @@ export class Environment {
         );
 
         if (source) {
-            return source.get(lowercaseName)!;
+            let variableToReturn = source.get(lowercaseName)!;
+            if (this.isMockedFunction(variableToReturn, lowercaseName)) {
```
strattonbrazil, line 161: Nit, but this calls getMockFunction() twice.  Duplicate calls have led to misc. bugs before so is there a way to shorten this?  Like:

    let mockedFunc = getMockFunction(...)
    if (mockedFunc) then return mockedFunc else return source.get(lowercaseName)

underwoo16: I think we probably could clean this up to prevent duplicate calls I just really wanted to make sure that the variable was a `Callable` so that we didn't return a mocked function for some `boolean` variable for example.

I'll take a look

underwoo16: How would you feel about using `this.mockFunctions.has(possibleMockName)` inside of `isMockedFunction`?

src/interpreter/Environment.ts
```diff
@@ -1,6 +1,7 @@
 import { Identifier } from "../lexer";
-import { BrsType, RoAssociativeArray, Int32, BrsInvalid, RoSGNode } from "../brsTypes";
+import { BrsType, RoAssociativeArray, Int32, BrsInvalid, RoSGNode, Callable } from "../brsTypes";
 import { ComponentDefinition } from "../componentprocessor";
+import { Call } from "../parser/Expression";
```
underwoo16, line 4: going to remove this unused import in next commit

src/interpreter/Environment.ts
strattonbrazil: Sounds good to me.

test/e2e/resources/mockFunctions.brs
```diff
@@ -0,0 +1,9 @@
+sub Main()
```
alimnios72, line 1: I think it would be valuable to add a unit test in the components too. For example, defining a function in a component and confirming that the mock replaces the functionality

test/e2e/resources/mockFunctions.brs
```diff
@@ -0,0 +1,9 @@
...
+    ' mock formatJSON
+    _brs_.mockfunction("formatJson", (function(json as object, flags=0 as integer) as dynamic
+        return "{fake:'json'}"
+    end function))
+
+    ' call the mocked function
+    print formatJson({}) ' => {fake:'json'}
```
sjbarag, line 8: Love this demo line!

test/e2e/resources/mockFunctions.brs
sjbarag: I agree, if only because of how tricky it was to get the scoping right for scenegraph nodes 😃   This should work as-is, but it'd be nice to check!

test/e2e/BrsMock.test.js
```diff
@@ -46,4 +46,12 @@ describe("end to end brightscript functions", () => {
             "name",
         ]);
     });
+
+    test("mockFunctions.brs", async () => {
+        await execute([resourceFile("mockFunctions.brs")], outputStreams);
+
+        expect(allArgs(outputStreams.stdout.write).filter(arg => arg !== "\n")).toEqual([
```
sjbarag, line 53: 🤔 I think there's been maybe… one time where we didn't do `.filter(arg => arg !== "\n")`.  @underwoo16 think it's worth us looking into abstracting this a bit more in the future? The boilerplate isn't too bad and it's definitely clear what's happening, but it's still boilerplate.

test/extensions/MockFunction.test.js
```diff
@@ -0,0 +1,39 @@
...
+        let interpreter;
+
+        beforeEach(() => {
+            interpreter = new Interpreter();
+        });
+
+        describe("calling mock function", () => {
+            // TODO: I don't think this environment spy is working as expected.
+            //       am able to step through and see setMockFunction called
+            //       but the mock expectations fail the test
```
sjbarag, line 17: How strange!

test/extensions/MockFunction.test.js
```diff
@@ -0,0 +1,39 @@
...
+                    },
+                    impl: () => {
+                        return true;
+                    },
+                });
+                const mockArgs = [mockName, mock];
+                mockFunction.getAllSignatureMismatches(mockArgs);
+                mockFunction.call(interpreter, ...mockArgs);
+                expect(envMockFunctionSpy).toHaveBeenCalledTimes(1);
+                expect(envMockFunctionSpy).toHaveBeenCalledWith([mockName.toString(), mock]);
```
sjbarag, line 35: I'm perfectly satisfied if we just want to check for that mock existing in `interpreter.environment` after `mockFunction.call()` — if we ever change the implementation we'll have to come back and fix this test again, which is kind of a bummer!

```suggestion
                expect(interpreter.environment.getMockFunction("testfunction")).toBe(mock);
```

src/interpreter/Environment.ts
```diff
@@ -152,7 +156,11 @@ export class Environment {
         );
 
         if (source) {
-            return source.get(lowercaseName)!;
+            let variableToReturn = source.get(lowercaseName)!;
+            if (this.isMockedFunction(variableToReturn, lowercaseName)) {
```
sjbarag, line 160: >

We could probably add an extra check to ensure we only mock names from the `module` and `global` scopes to prevent mocking of things in local scope!

```suggestion
            if (souce !== this.function && this.isMockedFunction(variableToReturn, lowercaseName)) {
```

underwoo16: Can you provide an example brightscript snippet capturing the spirit of `prevent mocking of things in local scope!` ?

I am consistently seeing `this.function` as an empty map so the mock is always being returned.

test/e2e/resources/mockFunctions.brs
underwoo16: we might have to sync up on this because I can't quite picture how this test would be set up :/

src/interpreter/Environment.ts
sjbarag: Sorry for the slow response — took the weekend off from side-projecting 😄 

It's mostly to prevent strange variable shadowing issues like:

```brightscript
sub main()
    print foo() ' => "Named Foo"

    foo = function()
        return "--inline foo--"
    end function

    print foo() ' => "--inline foo--"

    _brs_.mockFunction("foo", function()
        return "mOcKeD fOo"
    end function)

    print foo() ' => "mOcKeD fOo"
end sub

function foo()
    return "Named Foo"
end function
```

Unless I'm misunderstanding things!  That's absolutely possible 😅

underwoo16: Is the snippet you posted the correct expected behavior? Just want to be super clear before I make any changes 👍

underwoo16: Basically if I _DON'T_ implement the check against function scope the output is:
```
Named Foo
--inline foo--
mOcKeD fOo
```

But if I do implement the check against function scope I get:
```
Named Foo
--inline foo--
--inline foo--
```

sjbarag: I suspect we do want to implement that check, and prevent mocking of a local variable `foo = function() …`.  If we find a use-case where we do want that, we can always remove the check.

But!  If you've got a strong use-case for allowing that I'm totally okay with that too.  We make the rules here 😜

Issue #389, opened 2020-03-20
Scripts imported in XML components aren't loaded 


alimnios72:
While there are some [e2e tests](https://github.com/sjbarag/brs/blob/master/test/e2e/BrsComponents.test.js#L284-L326) for custom components with imports in the script tags this doesn't seem to work in a real brightscript code base. XML files are parsed correctly if they are in the `components` but brightscript files are not.

PR #400, opened 2020-04-16
fix(parser): Parse scripts included in components

alimnios72:
Scripts in components weren't parsed because they couldn't be found, `__dirname` references the path of the actual js script and not the brightscript project were we want to run the tests.

Fixes: #389

### DIFF
--- a/src/index.ts
+++ b/src/index.ts
@@ -44,10 +44,7 @@ export async function execute(filenames: string[], options: Partial<ExecutionOpt
         if (component.scripts.length < 1) return;
         try {
             component.scripts = component.scripts.map((script: ComponentScript) => {
-                script.uri = path.join(
-                    options.root ? options.root : executionOptions.root,
-                    new URL(script.uri).pathname
-                );
+                script.uri = path.join(executionOptions.root, new URL(script.uri).pathname);
                 return script;
             });
         } catch (error) {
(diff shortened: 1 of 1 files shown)


src/index.ts
```diff
@@ -45,7 +45,7 @@ export async function execute(filenames: string[], options: Partial<ExecutionOpt
         try {
             component.scripts = component.scripts.map((script: ComponentScript) => {
                 script.uri = path.join(
-                    options.root ? options.root : __dirname,
+                    options.root ? options.root : executionOptions.root,
```
strattonbrazil, line 48: Based on this string:

    const executionOptions = Object.assign(defaultExecutionOptions, options);

is there a time that _root_ has something that isn't in _executionOptions_?  Should this instead be just _executionOptions.root_?

sjbarag: Looks like `Object.assign` returns an object of type `ExecutionOptions & Partial<ExecutionOptions>` (basically `typeof A & typeof B`).  Switching that to `const executionOptions = { ...defaultExecutionOptions, options };` on line 38 should take care of that possibly-undefined `root` and allow this to be simplified to `executionoptions.root`!

alimnios72: Thanks for answering Sean, forgot to check this yesterday.

PR #401, opened 2020-04-20
feat(stdlib): Implement roSGNode#update second parameter

### LINKED ISSUE #378: Implement roSGNode#update({…}, true)
There's an undocumented second parameter to `roSGNode#update` that maps roughly to:

```brightscript
' @param {boolean} [createFields=false] - true if missing fields should be created in the calling node before mapping values, otherwise false
```

It's currently being used in [SGDEX's ButtonBar component](https://github.com/rokudev/SceneGraphDeveloperExtensions/blob/da1b1ad4167a3770566e00072bd4b58a4705c5aa/extensions/SGDEX/ButtonBar/ButtonBar.brs#L131-L133) and has existed in RBI for quite a while, but is still undocumented.

`brs` should support that second parameter!  (Or maybe it's a case of multiple dispatch, like in `Val(s as String) as Float` and `Val(str as String, radix as Integer) as Integer`

jessicacaley:
See issue https://github.com/sjbarag/brs/issues/378

fixes #378

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -600,7 +600,6 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             returns: ValueKind.Uninitialized,
         },
         impl: (interpreter: Interpreter, aa: RoAssociativeArray, createFields: BrsBoolean) => {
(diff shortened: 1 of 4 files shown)


sjbarag:
fixes #378

PR #402, opened 2020-04-30
Implement m.top for components

underwoo16:
- Adds a reference to the node itself on the mPointer as `top`
- Creates a new `mPointer` for each component to prevent `m.top` of created components being accessible everywhere

I can't think of other e2e test cases at the moment since we don't have `callFunc` or `onChange` callbacks working yet.

~~- The tricky part that this needs help on is when visiting calls. Getting the root mPointer prevents function calls inside of components from accessing `m.top`. But only getting the environments mPointer prevents function calls from within objects from accessing the global mPointer.~~
~~- I am not sure how to reconcile the above two scenarios.~~


src/interpreter/index.ts
```diff
@@ -122,7 +122,7 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
                 );
                 let statements = await componentScopeResolver.resolve(component);
                 interpreter.inSubEnv(subInterpreter => {
-                    subInterpreter.environment.setM(interpreter.environment.getM());
+                    subInterpreter.environment.setM(new RoAssociativeArray([]));
```
underwoo16, line 125: One thing I have tried that _seems to work_ is also setting the environments rootM here to the new associative array and then revevrting the below change to call `getRootM` again

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1109,6 +1109,8 @@ export function createNodeByType(interpreter: Interpreter, type: BrsString) {
             }
 
             interpreter.inSubEnv(subInterpreter => {
+                let mPointer = subInterpreter.environment.getM();
+                mPointer.set(new BrsString("top"), node);
```
alimnios72, line 1113: Will this work for the base `RoSGNode`? Seems like this is only being added for custom components

src/interpreter/index.ts
```diff
@@ -122,7 +122,9 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
                 );
                 let statements = await componentScopeResolver.resolve(component);
                 interpreter.inSubEnv(subInterpreter => {
-                    subInterpreter.environment.setM(interpreter.environment.getM());
+                    let componentMPointer = new RoAssociativeArray([]);
+                    subInterpreter.environment.setM(componentMPointer);
```
alimnios72, line 126: I don't really understand this change, what was wrong with setting `interpreter.environment.getM()` here?

src/brsTypes/components/RoSGNode.ts
underwoo16: I thought it only had to be added for custom components. I am unsure if base RoSGNode uses `m.top` at all?

src/interpreter/index.ts
underwoo16: That was passing down the existing `mPointer` so then `top` was being added to the global mPointer and the script that created a component could access `m.top.*` of that component.

underwoo16: For example:

```brightscript
sub main()
    ' create an object that uses m.top
    myComponent = createObject("roSGNode", "MTopWidget")

    ' m.top in this script should not refer to anything
    print m.top
end sub
```
Using `interpreter.environment.getM()` would mean that `m.top` in this script would print out the top value of the `MTopWidget` we created

src/brsTypes/components/RoSGNode.ts
alimnios72: Think I didn't explain myself correctly but I was talking about the base `Node` type. At the very top of this function there is an early return if the type is `Node`, thus basic nodes wouldn't have a `m.top` reference, right?

underwoo16: Ahh I see. Is there a use case for that as well? Do base `Node` types ever have scripts associated with them?

strattonbrazil: I don't think so.

sjbarag: Yeah it should be fine to not apply this kind of override to the generic `Node` type — the only time we'd need access to a `Node`'s interface is from a JavaScript function, at which point we shouldn't need to route through `someInterpreter.environment.getM().get("top")` (since those public fields are already available through `this`)

PR #404, opened 2020-05-01
Fix lex issue with `#ELSE IF` and `#END IF`

### LINKED ISSUE #403: Parse error for upper case #ELSE IF and #END IF
image

image

TwitchBronBron:
Fixes #403

### DIFF
--- a/test/lexer/Lexer.test.js
+++ b/test/lexer/Lexer.test.js
@@ -353,6 +353,19 @@ describe("lexer", () => {
             ]);
         });
 
+        it("reads conditional directives", () => {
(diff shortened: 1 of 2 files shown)


test/lexer/Lexer.test.js
```diff
@@ -353,6 +353,19 @@ describe("lexer", () => {
             ]);
         });
 
+        it("reads conditional directives", () => {
```
strattonbrazil, line 356: Can you change the name of the test to not match the above or just combine them into one string?

TwitchBronBron: Oops. copy-paste error. Fixed. (i combined them into a single test per your suggestion).

PR #405, opened 2020-05-01
feat(stdlib): Implement Group Component

kirkyoder:


src/brsTypes/components/Group.ts
```diff
@@ -0,0 +1,29 @@
+import { RoSGNode } from "./RoSGNode";
+import { ValueKind } from "../BrsType";
+import { AAMember } from "./RoAssociativeArray";
+
+export class Group extends RoSGNode {
+    readonly kind = ValueKind.Object;
```
strattonbrazil, line 6: I don't think this line is necessary since it's provided by RoSGNode.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1073,6 +1070,19 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             return new BrsString(this.type);
         },
     });
+
+    protected registerFields(fields: FieldModel[]) {
+        if (!fields) {
```
strattonbrazil, line 1075: I thought empty arrays are truthy in javascript?  If so, this will never be entered.  

I also don't think it's necessary since the forEach is a noop in this case (see RoSGNode constructor as a reference).  Empty arrays are forEach safe.

.editorconfig
```diff
@@ -5,3 +5,6 @@ charset = utf-8
 indent_style = space
 indent_size = 4
 trim_trailing_whitespace = true
+
+[*.test.js]
+trim_trailing_whitespace = false
```
sjbarag, line 10: Oh no, was this breaking a test?

src/brsTypes/components/Group.ts
```diff
@@ -0,0 +1,29 @@
...
+        { name: "inheritParentTransform", type: "boolean" },
+        { name: "inheritParentOpacity", type: "boolean" },
+        { name: "clippingRect", type: "array" },
+        { name: "renderPass", type: "integer" },
+        { name: "muteAudioGuide", type: "boolean" },
+        { name: "enableRenderTracking", type: "boolean" },
+        { name: "renderTracking", type: "string" },
+    ];
+
+    constructor(members: AAMember[] = []) {
```
sjbarag, line 25: Any chance this could accept a `FieldModel[]` (as defined in `RoSGNode.ts`) instead of an `AAMember[]`?  When we eventually add some typing for the list of supported `type` values, it'll be handy to have those plumbed through everywhere already 😃

lkipke: In this case, I think the logic would need to change so that the child component doesn't need to explicitly call `this.registerFields` for its own fields, i.e. the constructor logic for `Group` (or better yet, `RoSGNode`) would need to look something like:
```
constructor(members: FieldModel[] = []) {
  super("RoSGNode");
  let allFields = [...this.builtInFields, ...members];
  this.registerFields(allFields);
}
```

lkipke: Or actually, I guess you could still have the child call `this.registerFields` for its own fields, instead of passing them to the parent. Not sure which is more desirable 🤷‍♂️

.editorconfig
kirkyoder: Yeah, anywhere we want to compare against a string value like: `foo: ` that (intentionally) has trailing whitespace, it was being auto stripped. 😄

src/brsTypes/components/Group.ts
kirkyoder: Nice catch! 👍

src/brsTypes/components/RoSGNode.ts
kirkyoder: I need to disconnect my brain from a land where we've got interop between CoffeeScript & TypeScript. 😄 My intent here was to guard against `null` being supplied by a caller, but that's unnecessary since we're fully within the realm of TS here.

.editorconfig
sjbarag: Derp, sorry about that!

src/brsTypes/components/Group.ts
sjbarag: True!  I think I like the explicit `this.registerFields()` call over passing things through the constructor — if only because you could make the same argument about `this.registerMethods()` and constructor calls would be really large after adding both of those 😅

src/brsTypes/components/Group.ts
```diff
@@ -0,0 +1,26 @@
...
+        super(members, "Group");
+        this.registerFields(this.builtInFields);
```
lkipke, line 24: I might be misunderstanding, but won't this overwrite the `members` that `Group` is created with? As far as I can tell, the order of operations is:
1. Call `super(members, "Group")`
2. In the `RoSGNode` constructor, call `this.registerFields` with `builtInFields`
3. In the `RoSGNode` constructor, add all `members` to `this.fields` (intentionally overwriting any of the default `builtInFields` with the `members` that the node was created with).
4. Come back to `Group` constructor.
5. Call `this.registerFields` with `builtInFields`. **At this point, we are overwriting any of the `members` that `Group` passed into `super()` previously.**

src/brsTypes/components/Group.ts
lkipke: Also, second thought: don't we need `members` to have a value associated with them? Versus the `builtInFields`, which are all initialized to the default value. I assumed that when we had a component in Scenegraph like:
```
<Group foo="bar">
```
then we would be initializing it along the lines of
```
let members = [{ name: "foo", value: "bar" }];
let group = new Group(members);
```

And then the `builtInFields` are all of the interfaces that _can_ be defined on a certain node, initialized to default values, then overwritten by `members` if applicable.

Of course, that was just my assumption based on how the class is constructed, and I haven't had a chance to dive into the flow of parsing/initializing yet, so please correct me if that assumption is wrong 😅

src/brsTypes/components/Group.ts
kirkyoder: I think that sounds right, but only in the case where there is a collision between field names would the given value be overwritten. I think you bring up a good point, though, regarding whether `builtInFields` should supersede provided fields a la `members`.

src/brsTypes/components/Group.ts
kirkyoder: I _believe_ `members` would represent the _children_ of the component in question... but that's a good point to clarify. 😄 
```
<Group>
    <Rectangle />
    <String />
</Group>
```

package.json
```diff
@@ -11,6 +11,7 @@
   },
   "scripts": {
     "build": "tsc",
+    "dev": "tsc -w",
```
lkipke, line 14: We can alias this to something other than `dev` if it makes more sense, just thought it would be nice to have an alias. (Also, I know that you can just run `yarn build -- -w`, but if we're aliasing `tsc`, may as well alias `tsc -w`).

src/brsTypes/index.ts
```diff
@@ -39,6 +42,13 @@ export * from "./components/RoInt";
 export * from "./components/RoInvalid";
 export * from "./Callable";
 
+export type NodeCtor = () => RoSGNode;
+
+export const BrsBuiltInComponents = new Map<string, NodeCtor>([
```
lkipke, line 47: If there's a better place for this (or if it deserves its own file), let me know!

src/brsTypes/components/RoSGNode.ts
```diff
@@ -69,23 +71,18 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
     private parent: RoSGNode | BrsInvalid = BrsInvalid.Instance;
-    readonly builtInFields = [
+    readonly builtInFields: FieldModel[] = [
         { name: "change", type: "roAssociativeArray" },
         { name: "focusable", type: "boolean" },
         { name: "focusedChild", type: "node" },
         { name: "id", type: "string" },
     ];
 
     constructor(members: AAMember[], readonly type: string = "Node") {
```
lkipke, line 81: So, do we want this class to have the name `"Node"`, or `"roSGNode"`?

package.json
strattonbrazil: Seems fine.  I think "watch" would also work.  Whatever we settle on can you add a reference to this in the README with the other build commands?

src/brsTypes/index.ts
strattonbrazil: Seems fine.

src/brsTypes/components/RoSGNode.ts
strattonbrazil: I think _Node_ is fine for now.  It's simple to swap if we change our minds.

package.json
MATATAT: `watch` seems more descriptive to me.

src/brsTypes/components/Group.ts
MATATAT: I'm also finding the inconsistency with the naming confusing. Fields and members are comparable? If it's the children we might want to be more explicit about that as well.

src/brsTypes/components/Group.ts
MATATAT: Does this not just register fields with the `Group` and not the `RsSGNode`? If the `members` are being pulled from the `Group` then I would not expect them to be overwritten by the `builtInFields`. Roku seems to mostly flatten things into a single scope so the hierarchy becomes important. But it might be worth experimenting with the platform to see what it's doing explicitly.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1073,47 +1070,67 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
             typeDef = interpreter.environment.nodeDefMap.get(typeDef.extends);
             if (typeDef) typeDefStack.push(typeDef);
         }
 
+        // Start from the "basemost" component of the tree.
+        typeDef = typeDefStack.pop();
+
+        // If this extends a built-in component other than Node, create it.
+        // Default to Node otherwise.
+        let node: RoSGNode = new RoSGNode([], type.value);
```
underwoo16, line 1108: do we need this line since RoSGNode is in BrsBuiltInComponents?

src/brsTypes/index.ts
alimnios72: In the future we might need to use the factory pattern to create these but for now I also think it is fine

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1073,47 +1070,67 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
         //use typeDef object to tack on all the bells & whistles of a custom node
-        let node = new RoSGNode([], type.value);
         let typeDefStack: ComponentDefinition[] = [];
         let currentEnv = typeDef.environment;
 
         // Adding all component extensions to the stack to call init methods
         // in the correct order.
         typeDefStack.push(typeDef);
-        while (typeDef && typeDef.extends) {
+        while (typeDef) {
```
alimnios72, line 1098: Good catch!

src/brsTypes/index.ts
lkipke: Agreed, with the map I was following the pattern of `BrsObjects`: https://github.com/sjbarag/brs/blob/master/src/brsTypes/components/BrsObjects.ts#L20-L39

test/e2e/resources/components/Group.xml
```diff
@@ -1,7 +0,0 @@
-<?xml version="1.0" encoding="UTF-8"?>
```
lkipke, line 1: We shouldn't need this file anymore, so I deleted it. Tests all still pass!

src/brsTypes/components/Group.ts
```diff
@@ -0,0 +1,26 @@
+import { RoSGNode, FieldModel } from "./RoSGNode";
+import { AAMember } from "./RoAssociativeArray";
+
+export class Group extends RoSGNode {
+    readonly defaultFields: FieldModel[] = [
+        { name: "visible", type: "boolean", value: "true" },
```
lkipke, line 6: I instantiated these all with the default values from the Roku docs.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1073,47 +1071,76 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
     });
+
+    /**
+     * Takes a list of models and creates fields with default values, and adds them to this.fields.
+     * By default, it will not override existing fields.
+     */
+    protected registerFields(fields: FieldModel[], allowOverrides: boolean = false) {
+        fields.forEach(field => {
+            let fieldName = field.name.toLowerCase();
+            if (!allowOverrides && this.fields.has(fieldName)) {
```
lkipke, line 1082: This is so that we don't overwrite our instantiated `members` with default field values.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -69,24 +71,20 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
-        this.builtInFields.forEach(field => {
-            this.fields.set(
-                field.name.toLowerCase(),
-                new Field(getBrsValueFromFieldType(field.type), false)
-            );
-        });
+        // All nodes start have some built-in fields when created.
+        this.registerFields(this.defaultFields);
 
+        // TODO: filter out any non-allowed members (custom fields on an object are not allowed on a Roku device).
```
strattonbrazil, line 87: Can you clarify this statement?

src/brsTypes/components/RoSGNode.ts
strattonbrazil: Can you clarify what this is needed for?  If _registerFields_ is only used for _brs_-provided components is this to prevent a maintainer from accidentally adding a field in a subcomponent that overwrites one in the parent?  Should this silently raise an error?  Should this fail on the first field?  That seems hard to debug.  When will we ever allow overrides?  

If this will be used by custom components is this the desired behavior?  What does the Roku do in this situation?  
```
<component name="A">
    <interface>
        <field name="foo" type="integer">
    </interface>
</component>

<component name="B" extends="A">
    <interface>
        <field name="foo" type="string">
    </interface>
</component>

```

lkipke: Great questions! You bring up a good point about `interface`s. The problem I was trying to solve was this:
```
<Group opacity="0.5" />
```
In this scenario, we would first add `members` to `this.fields` in the `RoSGNode` constructor. Then, we would go through the _default_ fields (and `opacity` is a default field), so the custom `opacity` value that we got from the XML would be overwritten by the default.

I think it might be more clear to have a `registerDefaultFields` and `registerMembers`, and have the child call those methods in the appropriate order, i.e.:

```
class Group {
  constructor(members, name) {
    super([], name); // note that we will not pass members into the super constructor

    this.registerDefaultFields(this.defaultFields);
    this.registerMembers(members);
  }
}
```

alimnios72: Do you see any case where we'd want to override members with defaults? I can't think on any right now. If you think `allowOverrides` will never be used I'd suggest to go for the second approach.

lkipke: Yep agreed, I just made those changes!

src/brsTypes/components/RoSGNode.ts
```diff
@@ -69,27 +71,21 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
-            );
-        });
+        // All nodes start have some built-in fields when created.
+        this.registerDefaultFields(this.defaultFields);
 
-        members.forEach(member =>
-            this.fields.set(member.name.value.toLowerCase(), new Field(member.value, false))
-        );
+        // After registering default fields, then register fields instantiated with initial values.
+        this.registerInitializedFields(initializedFields);
```
lkipke, line 88: I renamed this to `initializedFields` to make the relationship between these and `defaultFields` more clear.

PR #407, opened 2020-05-06
fix(parser,stdlib) Run errors ignored

### LINKED ISSUE #383: Parsing errors in files executed with `Run` silently ignored
Consider two files:

#### main.brs
```brightscript
sub main()
    print "main.brs start"
    run("pkg:/foo.brs")
    print "main.brs end"
end sub
```

#### foo.brs
```brightscript
sub main()
    print "foo.brs start"

    m = {
        it: __it,
        assert: {
            isInvalid: sub(a, b)
                if a <> invalid then print b
            end sub
        }
    }

    m.it("is a test" sub()
    '              ~~~ note the missing `,`
        m.assert.isInvalid(12345, "TEST TEST TEST")
    end sub)
    print "foo.brs end"
end sub

sub __it(a as string, b as object)
    b()
end sub
```

Executing `foo.brs` directly results in parse errors:

```
$ brs foo.brs
foo.brs(13,21-24): Expected ')' after function call arguments
foo.brs(16,11-12): Found unexpected token ')'
foo.brs(18,0-7): Found unexpected token 'end sub'
Error occurred parsing
```

But executing `main.brs` executes… weirdly:

```
$ brs main.brs
main.brs start
foo.brs end
foo.brs start
main.brs end
```

We'll need to figure out what's happening here!

alimnios72:
- Added error listener to lexer-parser process (`lexParseSync`) to send errors to stderr
-  `lexParseSync` now throws an error stopping execution when parsing fails. This fixes `run` and `RunInScope` functions which use `lexParseSync`
- Removed `try`/`catch` block from RunInScope` to let the error bubble up to main error handler.

Fixes #383

### DIFF
--- a/src/stdlib/Run.ts
+++ b/src/stdlib/Run.ts
@@ -36,16 +36,10 @@ function runFiles(interpreter: Interpreter, filenames: BrsString[], args: BrsTyp
         return BrsInvalid.Instance;
     }
 
-    try {
-        let ast = brs.lexParseSync(pathsToFiles, interpreter.options);
-        // execute the new files in a brand-new interpreter, as no scope is shared with the `Run`-ed files in RBI
-        let sandbox = new Interpreter(interpreter.options);
-        return sandbox.exec(ast, ...args)[0] || BrsInvalid.Instance;
-    } catch (err) {
-        // swallow errors and just return invalid; RBI returns invalid for "file doesn't exist" errors,
-        // syntax errors, etc.
-        return BrsInvalid.Instance;
-    }
+    let ast = brs.lexParseSync(pathsToFiles, interpreter.options);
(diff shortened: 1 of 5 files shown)


src/stdlib/Run.ts
```diff
@@ -36,16 +36,10 @@ function runFiles(interpreter: Interpreter, filenames: BrsString[], args: BrsTyp
         return BrsInvalid.Instance;
     }
 
-    try {
```
sjbarag, line 39: I wouldn't mind keeping the try/catch here to maintain consistency with RBI?  Since [roca](https://github.com/hulu/roca) uses `_brs_.runInScope` leaving the default `Run` in "silent mode" should be safe while still giving us the debugging boost we need 😃

alimnios72: That makes sense, I'll revert the change

PR #409, opened 2020-05-12
Fix bug where child fields are always treated as strings

lkipke:
# Change summary
I noticed that when you referenced a component as a child and tried to set its fields, it only worked when they were strings. For example:
```
<component name="foo">
  <children>
    <Group
        id="group"
        childRenderOrder="renderLast"
        opacity="0.2"
    />
  <children>
</component>
```
```
foo = createObject("roSGNode", "foo")
group = foo.findNode("group")
print group.childRenderOrder    ' => "renderLast"
print group.opacity                     ' => "1.0" (or whatever default is)
```

This PR fixes that case by finding and using the type of the field being set.

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -1189,9 +1189,18 @@ function addChildren(interpreter: Interpreter, node: RoSGNode, typeDef: Componen
         if (newChild instanceof RoSGNode && appendChild) {
             appendChild.call(interpreter, newChild);
             let setField = newChild.getMethod("setfield");
-            for (let [key, value] of Object.entries(child.fields)) {
-                if (setField) {
-                    setField.call(interpreter, new BrsString(key), new BrsString(value));
+            if (setField) {
+                let nodeFields = newChild.getFields();
+                for (let [key, value] of Object.entries(child.fields)) {
+                    let field = nodeFields.get(key);
+                    if (field) {
+                        setField.call(
+                            interpreter,
+                            new BrsString(key),
+                            // use the field type to construct the field value
+                            getBrsValueFromFieldType(field.getType(), value)
+                        );
+                    }
                 }
             }
         }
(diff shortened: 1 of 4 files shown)


src/brsTypes/components/RoSGNode.ts
```diff
@@ -1189,9 +1189,18 @@ function addChildren(interpreter: Interpreter, node: RoSGNode, typeDef: Componen
         if (newChild instanceof RoSGNode && appendChild) {
             appendChild.call(interpreter, newChild);
             let setField = newChild.getMethod("setfield");
-            for (let [key, value] of Object.entries(child.fields)) {
-                if (setField) {
-                    setField.call(interpreter, new BrsString(key), new BrsString(value));
+            if (setField) {
```
strattonbrazil, line 1192: Good optimization.  Is this check still needed though?  When does a type of RoSGNode not have _setField_ implemented?  I see several similar checks in the file that I don't understand.  @alimnios72 , do you know why they're needed?

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1189,9 +1189,18 @@ function addChildren(interpreter: Interpreter, node: RoSGNode, typeDef: Componen
...
+                let nodeFields = newChild.getFields();
+                for (let [key, value] of Object.entries(child.fields)) {
+                    let field = nodeFields.get(key);
+                    if (field) {
+                        setField.call(
+                            interpreter,
+                            new BrsString(key),
+                            // use the field type to construct the field value
+                            getBrsValueFromFieldType(field.getType(), value)
```
strattonbrazil, line 1201: Nice fix.

src/brsTypes/components/RoSGNode.ts
alimnios72: The check was added because `getMethod` can return `undefined` https://github.com/sjbarag/brs/blob/master/src/brsTypes/components/BrsComponent.ts#L40  and the typescript compiler won't let you use `setField` without a check

PR #411, opened 2020-05-12
Implement Label component

lkipke:
# Change Summary
This implements the `<Label>` component.

### DIFF
--- a/test/brsTypes/components/Label.test.js
+++ b/test/brsTypes/components/Label.test.js
@@ -1,0 +1,51 @@
+const brs = require("brs");
+const { Label } = brs.types;
+
+describe("Font", () => {
+    describe("stringification", () => {
+        it("inits a new Font component", () => {
+            let group = new Label();
+
+            expect(group.toString()).toEqual(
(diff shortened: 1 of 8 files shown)


test/brsTypes/components/Label.test.js
```diff
@@ -0,0 +1,51 @@
+const brs = require("brs");
+const { Label } = brs.types;
+
+describe("Font", () => {
```
kirkyoder, line 4: I think this should be `"Label"` 🙃

test/brsTypes/components/Label.test.js
```diff
@@ -0,0 +1,51 @@
...
+    describe("stringification", () => {
+        it("inits a new Font component", () => {
```
kirkyoder, line 6: and ditto here ☝️

test/brsTypes/components/Label.test.js
lkipke: Oh noooo, stupid copy 🍝

PR #412, opened 2020-05-13
feat(stdlib): Added support for FindMemberFunction

### LINKED ISSUE #78: Implement global reflective functions to return interface information
See the [BrightScript documentation](https://sdkdocs.roku.com/display/sdkdoc/Global+Utility+Functions#GlobalUtilityFunctions-GetInterface(objectasObject,ifnameasString)asInterface) for details on what each one does, but this should cover:

* `GetInterface(object as Object, ifname as String) as Interface`
* `FindMemberFunction(object as Object, funName as String) As Interface`

This is currently blocked until objects are implemented, but when implemented objects one should keep this function in mind as it will need access to which interface a function came from.

alimnios72:
FindMemberFunction can now be used on `brs`

fixes #78

### DIFF
--- a/src/stdlib/GlobalUtilities.ts
+++ b/src/stdlib/GlobalUtilities.ts
@@ -39,3 +39,23 @@ export const GetInterface = new Callable("GetInterface", {
         return object.interfaces.get(ifname.value.toLowerCase()) || BrsInvalid.Instance;
     },
 });
+
+export const FindMemberFunction = new Callable("FindMemberFunction", {
+    signature: {
+        args: [
+            new StdlibArgument("object", ValueKind.Object),
+            new StdlibArgument("funname", ValueKind.String),
+        ],
+        returns: ValueKind.Interface,
+    },
+    impl: (interpreter, object: BrsComponent, funname: BrsString): BrsInterface | BrsInvalid => {
+        let iface: BrsType = BrsInvalid.Instance;
+        object.interfaces.forEach(interfaceName => {
+            if (interfaceName.methodNames.has(funname.value.toLowerCase())) {
+                iface = interfaceName;
+            }
+        });
+
+        return iface;
+    },
+});
--- a/src/stdlib/GlobalUtilities.ts
+++ b/src/stdlib/GlobalUtilities.ts
@@ -48,10 +48,10 @@ export const FindMemberFunction = new Callable("FindMemberFunction", {
         ],
         returns: ValueKind.Interface,
     },
-    impl: (interpreter, object: BrsComponent, funname: BrsString): BrsInterface | BrsInvalid => {
+    impl: (interpreter, object: BrsComponent, funName: BrsString): BrsInterface | BrsInvalid => {
         let iface: BrsType = BrsInvalid.Instance;
         object.interfaces.forEach(interfaceName => {
-            if (interfaceName.methodNames.has(funname.value.toLowerCase())) {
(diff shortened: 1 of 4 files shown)


src/stdlib/GlobalUtilities.ts
```diff
@@ -39,3 +39,23 @@ export const GetInterface = new Callable("GetInterface", {
         return object.interfaces.get(ifname.value.toLowerCase()) || BrsInvalid.Instance;
     },
 });
+
+export const FindMemberFunction = new Callable("FindMemberFunction", {
+    signature: {
+        args: [
+            new StdlibArgument("object", ValueKind.Object),
+            new StdlibArgument("funname", ValueKind.String),
```
strattonbrazil, line 47: Nit, but can we camelcase uses of _funname_ as is done in the Roku docs?

https://developer.roku.com/en-ca/docs/references/brightscript/language/global-utility-functions.md#findmemberfunctionobject-as-object-funname-as-string-as-interface

src/stdlib/GlobalUtilities.ts
```diff
@@ -39,3 +39,23 @@ export const GetInterface = new Callable("GetInterface", {
...
+        ],
+        returns: ValueKind.Interface,
+    },
+    impl: (interpreter, object: BrsComponent, funname: BrsString): BrsInterface | BrsInvalid => {
+        let iface: BrsType = BrsInvalid.Instance;
+        object.interfaces.forEach(interfaceName => {
+            if (interfaceName.methodNames.has(funname.value.toLowerCase())) {
```
strattonbrazil, line 54: I assume the method name on the interface is stored lowercase when registering?

src/stdlib/GlobalUtilities.ts
alimnios72: I thought of that but wanted to keep things consistent, if you see `GetInterface` above we do the same for `ifname` arg. Also in `RoSGNode.ts` all of the arguments are in lowercase (i.e. `fieldname`, `alwaysnotify`, `functionname`, etc) I think we should keep it as it right now to avoid changing all other occurrences

src/stdlib/GlobalUtilities.ts
alimnios72: yes: https://github.com/sjbarag/brs/blob/master/src/brsTypes/components/BrsComponent.ts#L28

src/stdlib/GlobalUtilities.ts
```diff
@@ -39,3 +39,23 @@ export const GetInterface = new Callable("GetInterface", {
...
+
+export const FindMemberFunction = new Callable("FindMemberFunction", {
+    signature: {
+        args: [
+            new StdlibArgument("object", ValueKind.Object),
+            new StdlibArgument("funname", ValueKind.String),
+        ],
+        returns: ValueKind.Interface,
+    },
+    impl: (interpreter, object: BrsComponent, funname: BrsString): BrsInterface | BrsInvalid => {
```
MATATAT, line 51: I saw your comments on the `StdLibArgument` but I think we can switch this `funname` to `funName`.

src/stdlib/GlobalUtilities.ts
sjbarag: Agreed — this might be part of a larger validation pass where we ensure our capitalization is correct across the codebase.  It's a "fit and finish" type of issue that we'll definitely need to handle eventually!

PR #413, opened 2020-05-13
feat(stdlib): implement Font component

lkipke:
# Change Summary
This implements the `<Font>` component.

### DIFF
--- a/src/brsTypes/BrsType.ts
+++ b/src/brsTypes/BrsType.ts
@@ -132,6 +132,10 @@ export function getBrsValueFromFieldType(type: string, value?: string): BrsType 
         case "string":
             returnValue = value ? new BrsString(value) : new BrsString("");
             break;
+        case "uri":
+            // BrightScript represents the "uri" field as a string internally.
+            returnValue = value ? new BrsString(value) : new BrsString("");
+            break;
         default:
             returnValue = Uninitialized.Instance;
             break;
--- a/src/brsTypes/BrsType.ts
+++ b/src/brsTypes/BrsType.ts
@@ -129,11 +129,8 @@ export function getBrsValueFromFieldType(type: string, value?: string): BrsType 
         case "assocarray":
             returnValue = BrsInvalid.Instance;
             break;
-        case "string":
-            returnValue = value ? new BrsString(value) : new BrsString("");
-            break;
         case "uri":
-            // BrightScript represents the "uri" field as a string internally.
+        case "string":
             returnValue = value ? new BrsString(value) : new BrsString("");
             break;
         default:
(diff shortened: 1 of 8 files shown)


src/brsTypes/components/Font.ts
```diff
@@ -0,0 +1,17 @@
+import { RoSGNode, FieldModel } from "./RoSGNode";
+import { AAMember } from "./RoAssociativeArray";
+
+export class Font extends RoSGNode {
+    readonly defaultFields: FieldModel[] = [
+        { name: "uri", type: "uri" },
```
lkipke, line 6: [The docs](https://developer.roku.com/en-gb/docs/references/scenegraph/typographic-nodes/font.md) technically say `uri` is the type for Font, but it's pretty clear they represent it as a `BrsString`. Happy to make this `string` instead of `uri` if that seems more appropriate.

src/brsTypes/BrsType.ts
```diff
@@ -132,6 +132,10 @@ export function getBrsValueFromFieldType(type: string, value?: string): BrsType
         case "string":
             returnValue = value ? new BrsString(value) : new BrsString("");
             break;
+        case "uri":
```
kirkyoder, line 135: 💭 if we leave this as a `string`, should we just make pull this up as an additional case statement alongside `"string"`, since the logic is equivalent?

lkipke: Ah, yes, good call!

src/brsTypes/components/Font.ts
sjbarag: `uri` appears to be a valid type for scenegraph components, but it appears to be an alias for `string` — I don't see any URI validation after creating a `Font` node in RBI!

lkipke: Yeah, I saw that as well -- plus, I noticed that I could perform all of the normal `string` operations on it.

test/brsTypes/components/Font.test.js
```diff
@@ -0,0 +1,23 @@
...
+                `<Component: roSGNode:Font> =
+{
+    change: <UNINITIALIZED>
+    focusable: false
+    focusedchild: invalid
+    id: 
+    uri: 
+    size: 1
+    fallbackglyph: 
+}`
```
sjbarag, line 19: This is _just barely_ different than RBI — looks like `id` and `uri` default to empty strings!

```
<Component: roSGNode:Font> =
{
    fallbackGlyph: ""
    size: 1
    uri: "foo"
    change: <Component: roAssociativeArray>
    focusable: false
    focusedChild: <Component: roInvalid>
    id: ""
}
```

Would you mind setting defaults for those in `Font.ts` (and maybe `RoSGNode.ts` for the base)?

sjbarag: After chatting out of band, we're going to solve this in https://github.com/sjbarag/brs/issues/416 !

PR #414, opened 2020-05-13
feat(stdlib): implement Poster component

lkipke:
# Change Summary
This implements the `<Poster>` component.

### DIFF
--- a/src/brsTypes/components/Poster.ts
+++ b/src/brsTypes/components/Poster.ts
@@ -1,0 +1,32 @@
+import { FieldModel } from "./RoSGNode";
+import { AAMember } from "./RoAssociativeArray";
+import { Group } from "./Group";
+
+export class Poster extends Group {
+    readonly defaultFields: FieldModel[] = [
+        { name: "uri", type: "string" },
+        { name: "width", type: "float", value: "0.0" },
+        { name: "height", type: "float", value: "0.0" },
+        { name: "loadSync", type: "boolean", value: "false" },
+        { name: "loadWidth", type: "float", value: "0.0" },
+        { name: "loadHeight", type: "float", value: "0.0" },
+        { name: "loadDisplayMode", type: "string", value: "noScale" },
+        { name: "loadStatus", type: "string", value: "noScale" },
+        { name: "bitmapWidth", type: "float", value: "0.0" },
+        { name: "bitmapHeight", type: "float", value: "0.0" },
+        { name: "bitmapMargins", type: "assocarray" },
+        { name: "blendColor", type: "string", value: "0xFFFFFFFF" },
+        { name: "loadingBitmapUri", type: "string" },
+        { name: "loadingBitmapOpacity", type: "float", value: "1.0" },
+        { name: "failedBitmapUri", type: "string" },
+        { name: "failedBitmapOpacity", type: "float", value: "1.0" },
+        { name: "audioGuideText", type: "string" },
+    ];
+
+    constructor(initializedFields: AAMember[] = [], readonly name: string = "Poster") {
+        super([], name);
+
+        this.registerDefaultFields(this.defaultFields);
+        this.registerInitializedFields(initializedFields);
+    }
+}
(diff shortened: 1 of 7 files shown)


src/brsTypes/components/Poster.ts
```diff
@@ -0,0 +1,32 @@
...
+        { name: "bitmapMargins", type: "assocarray" },
+        { name: "blendColor", type: "string", value: "0xFFFFFFFF" },
+        { name: "loadingBitmapUri", type: "string" },
+        { name: "loadingBitmapOpacity", type: "float", value: "1.0" },
+        { name: "failedBitmapUri", type: "string" },
+        { name: "failedBitmapOpacity", type: "float", value: "1.0" },
+        { name: "audioGuideText", type: "string" },
+    ];
+
+    constructor(initializedFields: AAMember[] = [], readonly name: string = "Poster") {
```
kirkyoder, line 26: Resurrecting [this feedback](https://github.com/sjbarag/brs/pull/405#discussion_r419706979) from @sjbarag on https://github.com/sjbarag/brs/pull/405.

Do we still want to transition `AAMember[]` to `FieldModel[]`? 🤔

lkipke: Great question! So I've actually found that we _never_ use `initializedFields`, outside of the unit tests for `RoSGNode`. In theory, it makes sense:
```
<Node somefield="somevalue" />
```
Where `{ somefield: somevalue }` would be represented as an `AAMember`. Actually, though, in our current implementation, we always add the fields _after_ creating the object, (both when [the component is a child](https://github.com/sjbarag/brs/blob/master/src/brsTypes/components/RoSGNode.ts#L1197-L1216) and for `interface` fields for [custom components](https://github.com/sjbarag/brs/blob/master/src/brsTypes/components/RoSGNode.ts#L1138).

I'm not sure if this was intentional, or if we'd rather add the fields in the constructor. If the former, then we should remove `initializedFields` altogether. If that latter, then we should refactor how we create component objects and add fields.

lkipke: Either way, probably out of scope for this PR 🤷‍♂️

src/brsTypes/components/Poster.ts
```diff
@@ -0,0 +1,32 @@
...
+        { name: "height", type: "float", value: "0.0" },
+        { name: "loadSync", type: "boolean", value: "false" },
+        { name: "loadWidth", type: "float", value: "0.0" },
+        { name: "loadHeight", type: "float", value: "0.0" },
+        { name: "loadDisplayMode", type: "string", value: "noScale" },
+        { name: "loadStatus", type: "string", value: "noScale" },
+        { name: "bitmapWidth", type: "float", value: "0.0" },
+        { name: "bitmapHeight", type: "float", value: "0.0" },
+        { name: "bitmapMargins", type: "assocarray" },
+        { name: "blendColor", type: "string", value: "0xFFFFFFFF" },
```
lkipke, line 18: So the type for this is actually `color` in the docs, which I thought was an alias for `string`. But upon closer inspection in RBI, it appears to be an alias for `integer`. Also, RBI does seem to do some "is this a valid color hex"-checking because, for example, setting it to `10` didn't work, but setting it to `-10` did.

sjbarag: That's pretty strange.  I've definitely seen odd behavior with colors in RBI but wasn't sure if it was simply a type-casting issue.  Being able to provide a negative value _feels_ like an integer overflow, but `0xFFFFFFFF` fits evenly into a 32-bit unsigned integer.  Using a 16-bit integer would be strange though, since that fits only up to 65,535 (and 0xFFFFFF without the opacity portion converts to 16,777,215).

🤔 I have no idea how RBI handles this then.

sjbarag: Filed https://github.com/sjbarag/brs/issues/417 to investigate and fix this at a later date!

test/e2e/resources/components/Poster.brs
```diff
@@ -0,0 +1,12 @@
+sub Main()
+    poster = createObject("roSGNode", "Poster")
+    print "poster node type:" type(poster)                                       ' => Group
```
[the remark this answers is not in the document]
sjbarag: Filed #418 for this!

PR #415, opened 2020-05-13
feat(stdlib): implement LayoutGroup component

lkipke:
# Change Summary
This implements the `<LayoutGroup>` component.

### DIFF
--- a/src/brsTypes/components/LayoutGroup.ts
+++ b/src/brsTypes/components/LayoutGroup.ts
@@ -1,0 +1,20 @@
+import { FieldModel } from "./RoSGNode";
+import { AAMember } from "./RoAssociativeArray";
+import { Group } from "./Group";
+
+export class LayoutGroup extends Group {
+    readonly defaultFields: FieldModel[] = [
+        { name: "layoutDirection", type: "string", value: "vert" },
+        { name: "horizAlignment", type: "string", value: "left" },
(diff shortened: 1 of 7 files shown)


test/brsTypes/components/LayoutGroup.test.js
```diff
@@ -0,0 +1,39 @@
+const brs = require("brs");
+const { LayoutGroup } = brs.types;
+
+describe("Rectangle", () => {
```
kirkyoder, line 4: don't forget to update this name 🌮

lkipke: Ugh, I need to write a script to do this for me or something so I don't miss silly copy changes 😅

src/brsTypes/components/LayoutGroup.ts
```diff
@@ -0,0 +1,20 @@
+import { FieldModel } from "./RoSGNode";
+import { AAMember } from "./RoAssociativeArray";
+import { Group } from "./Group";
+
+export class LayoutGroup extends Group {
```
sjbarag, line 5: We'll eventually need to implement `boundingRect()` on this differently than we do the other components, simply because of the fact that `itemSpacings` has to be included.  But since we haven't implemented that _anywhere_ yet, it doesn't matter just yet! 🎉

sjbarag:
Thanks for the rebase!  I'll get a new version published today.

PR #420, opened 2020-05-19
feat(stdlib): implement ArrayGrid component

lkipke:
# Change Summary
This implements the `<ArrayGrid>` component.

I wasn't sure how much of the logic we wanted to add initially, vs. just having a skeleton `ArrayGrid` that could be instantiated in BrightScript. It seems like most of the special logic would come from the `contentNode` field; however, nodes as fields are not yet supported in `brs`, so we can't do too much with it yet.

Let me know what you think!


strattonbrazil:
> I wasn't sure how much of the logic we wanted to add initially

I think getting the skeleton up to unblock general loading across an app as you've done makes more sense.  We can implement the details if users ever need this functionality in tests (which may be never).

Issue #408, opened 2020-05-07
Avoid adding inherited fields and children in two places


alimnios72:
The component processor creates a tree representation of XML components adding inherited fields and children to [each component](https://github.com/sjbarag/brs/blob/master/src/componentprocessor/index.ts#L85) but the same task is done when custom [components are created](https://github.com/sjbarag/brs/blob/master/src/brsTypes/components/RoSGNode.ts#L1103). It probably makes more sense to only recurse at component creation since that will keep preprocessing simpler and faster.

PR #424, opened 2020-05-27
On-change callbacks

lkipke:
# Change Summary
This finishes wiring up the `onChange` callbacks. A lot of the work was already there; this PR just modifies the calling pattern so that the correct environment is being used for the callbacks.

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -63,7 +63,9 @@ class Field {
      * child's observer should be called.
      */
     createExclusiveObserver(interpreter: Interpreter, callable: Callable) {
-        this.observers = this.observers.filter((observer) => observer.callable.name !== callable.name);
+        this.observers = this.observers.filter(
+            observer => observer.callable.name !== callable.name
+        );
         this.addObserver(interpreter, callable);
     }
 
@@ -1210,7 +1212,10 @@ function addFields(interpreter: Interpreter, node: RoSGNode, typeDef: ComponentD
             if (value.onChange) {
                 let field = node.getFields().get(key.toLowerCase());
                 if (field) {
-                    field.createExclusiveObserver(interpreter, interpreter.getCallableFunction(value.onChange.toLowerCase()))
+                    field.createExclusiveObserver(
+                        interpreter,
+                        interpreter.getCallableFunction(value.onChange.toLowerCase())
+                    );
                 }
             }
         }
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -17,9 +17,11 @@ import { RoArray } from "./RoArray";
 import { AAMember } from "./RoAssociativeArray";
 import { ComponentDefinition, ComponentNode } from "../../componentprocessor";
 import { ComponentFactory, BrsComponentName } from "./ComponentFactory";
+import { Environment } from "../../interpreter/Environment";
 
 interface BrsCallback {
     interpreter: Interpreter;
+    environment: Environment;
     callable: Callable;
 }
 
@@ -48,22 +50,28 @@ class Field {
     }
 
     setValue(value: BrsType) {
-        if (this.alwaysNotify || this.value !== value) {
+        let oldValue =  this.value;
+        this.value = value;
+        if (this.alwaysNotify || oldValue !== value) {
             this.observers.map(this.executeCallbacks);
         }
-        this.value = value;
     }
 
     addObserver(interpreter: Interpreter, callable: Callable) {
         let brsCallback: BrsCallback = {
             interpreter: interpreter,
+            environment: interpreter.environment,
             callable: callable,
         };
         this.observers.push(brsCallback);
     }
 
     private executeCallbacks(callback: BrsCallback) {
-        callback.callable.call(callback.interpreter);
+        let { interpreter, callable, environment } = callback;
+        interpreter.inSubEnv(subInterpreter => {
+            callable.call(subInterpreter);
+            return BrsInvalid.Instance;
+        }, environment);
     }
 }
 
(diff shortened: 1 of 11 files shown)


src/brsTypes/BrsType.ts
```diff
@@ -109,12 +109,14 @@ export function getBrsValueFromFieldType(type: string, value?: string): BrsType
     let returnValue: BrsType;
 
     switch (type.toLowerCase()) {
+        case "bool":
```
lkipke, line 112: These changes will be resolved once #422 is merged.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -48,22 +50,28 @@ class Field {
...
+        if (this.alwaysNotify || oldValue !== value) {
             this.observers.map(this.executeCallbacks);
         }
-        this.value = value;
     }
 
     addObserver(interpreter: Interpreter, callable: Callable) {
         let brsCallback: BrsCallback = {
             interpreter: interpreter,
+            environment: interpreter.environment,
```
lkipke, line 63: This callback didn't have the correct environment, because the `environment` property in the `Interpreter` is mutable and changes often, so I opted to save the environment as well.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1134,8 +1142,12 @@ export function createNodeByType(interpreter: Interpreter, type: BrsString): RoS
         while (typeDef) {
             let init: BrsType;
 
-            addChildren(interpreter, node, typeDef);
-            addFields(interpreter, node, typeDef);
+            interpreter.inSubEnv(subInterpreter => {
+                addChildren(subInterpreter, node!, typeDef!);
```
lkipke, line 1146: Typescript did not seem to believe that `node` and `typeDef` were defined once I put them in this anonymous function, so I had to add the `!`.

test/e2e/BrsComponents.test.js
```diff
@@ -526,4 +526,25 @@ describe("end to end brightscript functions", () => {
...
+
+    test.only("components/scripts/FieldChangeRunner.brs", async () => {
+        outputStreams.root = __dirname + "/resources";
+        await execute(
+            [resourceFile("components", "scripts", "FieldChangeRunner.brs")],
+            outputStreams
+        );
+
+        expect(allArgs(outputStreams.stdout.write).filter(arg => arg !== "\n")).toEqual([
+            "child: text field changed. new value:",
```
lkipke, line 538: Because of the way that we instantiate fields, these `onChange` callbacks are called on the initial set of the field. This is different than how RBI behaves (it doesn't call `onChange` for the initial value set), so I'm looking into how to refactor the call to make it not do this.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -48,22 +50,40 @@ class Field {
...
         }
-        this.value = value;
+    }
+
+    /**
+     * Adds an observer and removes any existing observers with the same name.
+     * In the case where a child and a parent have the same observer name, only the
+     * child's observer should be called.
+     */
+    createExclusiveObserver(interpreter: Interpreter, callable: Callable) {
```
alimnios72, line 65: Should we move this logic into `addObserver` and just have one method for adding observers? I guess we don't want more than one callback with the same name in any case.

test/e2e/BrsComponents.test.js
lkipke: This was happening because both the child and the parent had the field defined on them based on how we do component processing. Once I changed that, it worked!

src/componentprocessor/index.ts
```diff
@@ -130,7 +130,10 @@ async function processXmlTree(
                         // the original value without throwing an error for simplicity
                         // TODO: throw exception when fields are duplicated.
                         inheritedFields = { ...nodeFields, ...inheritedFields };
-                        newNodeDef.fields = inheritedFields;
+
```
lkipke, line 133: This is related to (but doesn't fix; feels out of scope) #408 .

src/brsTypes/components/RoSGNode.ts
```diff
@@ -48,22 +50,28 @@ class Field {
...
         };
         this.observers.push(brsCallback);
     }
 
     private executeCallbacks(callback: BrsCallback) {
-        callback.callable.call(callback.interpreter);
+        let { interpreter, callable, environment } = callback;
+        interpreter.inSubEnv(subInterpreter => {
+            callable.call(subInterpreter);
+            return BrsInvalid.Instance;
```
sjbarag, line 73: Does TypeScript complain about not returning a `BrsType` here?  That's a bummer since we're not actually using the return value, but I suppose we also don't want to return the result of that `call` all the time. 🤔 

If this feels like a pain in the butt to use, I'm open to suggestions!

src/brsTypes/components/RoSGNode.ts
sjbarag: Straaaaange

test/e2e/BrsComponents.test.js
```diff
@@ -526,4 +526,21 @@ describe("end to end brightscript functions", () => {
             "5",
         ]);
     });
+
+    test("components/scripts/FieldChangeRunner.brs", async () => {
+        outputStreams.root = __dirname + "/resources";
```
sjbarag, line 531: Let's be sure to reset `outputStreams.root` — looks like it only gets set up once at the beginning of this suite!

https://github.com/sjbarag/brs/blob/5e843c121d17c6f982876762a10b9f9176860960/test/e2e/BrsComponents.test.js#L9-L21

test/e2e/BrsComponents.test.js
```diff
@@ -526,4 +526,21 @@ describe("end to end brightscript functions", () => {
...
+        );
+
+        expect(allArgs(outputStreams.stdout.write).filter(arg => arg !== "\n")).toEqual([
+            "runner: node text field value before modifying:",
+            "initial",
+            "child: text field changed. new value:",
+            "modified",
+            "runner: node text field value after modifying:",
+            "modified",
+        ]);
```
sjbarag, line 544: 🔥 that output looks perfect!

test/e2e/BrsComponents.test.js
lkipke: It turns out that all of these tests succeed because `root` was set in one of them, then never cleaned up (so the change propagated to all of them, allowing them to succeed). Because of that, and since these are all component tests, I just assigned `outputStreams.root` in the `beforeEach` for the entire suite.

src/brsTypes/components/RoSGNode.ts
lkipke: Yep, it does complain about that. I think it works for now 🤷‍♂️

PR #425, opened 2020-05-27
 Implement global scope references (m.global)

niveditasonker:
- Added a reference to the node itself on the mPointer as global
- Added a simple test to check if the fields created in the globalWidget are globally available when referencing `m.global `


src/brsTypes/components/RoSGNode.ts
```diff
@@ -66,7 +66,7 @@ class Field {
         callback.callable.call(callback.interpreter);
     }
 }
-
+let myGlobal: RoSGNode;
```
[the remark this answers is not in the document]
strattonbrazil: I'm not too concerned about it living in this file compared to the interpreter. I don't see any advantage one way or the other. 

Minor nit on the name though as I prefer not seeing "my-" as a prefix as I read that as the author's interpretation of something that I would see in a blog. I think _mGlobal_, _globalNode_, or _sharedGlobal_ are all fine.

test/e2e/resources/components/componentGlobalUsage.brs
```diff
@@ -0,0 +1,7 @@
+sub main()
+    ' create an object that uses m.global
```
strattonbrazil, line 2: Maybe clarify comment, "create an object that writes value to global, create another object that tries to read that value from global".

test/e2e/resources/components/componentGlobalUsage.brs
```diff
@@ -0,0 +1,7 @@
...
+    myComponent = createObject("roSGNode", "MGlobalWidget")
+    mySecondComponent = createObject("roSGNode", "MGlobalWidget")
```
alimnios72, line 4: Could this be another type of component so it's clear that works among components?

src/brsTypes/components/RoSGNode.ts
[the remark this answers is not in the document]
sjbarag: I was tempted to suggest putting this in `Environment.ts` but I keep forgetting `m.global` is _only_ supported in the context of a scenegraph node.  Do we need to make any changes to `getGlobalAA()` as well?  TBH that function's name has always confused me 😅

sjbarag:
> It looks good but seems like this `global` won't be part of the AA returned by `GetGlobalAA` if the environment constructor is called with a `rootM` argument https://github.com/sjbarag/brs/blob/master/src/interpreter/Environment.ts#L24 isn't it?

Agreed — just ran a quick test on a very boring component and `getGlobalAA()` from within `init()` needs to include both `global` and `top`:

```brightscript
sub init()
    print getGlobalAA() ' => <Component: roAssociativeArray> =
    ' {
    '    global: <Component: roSGNode:Node>
    '    top: <Component: roSGNode:AVeryBoringComponent>
    ' }
end sub
```

test/e2e/ComponentGlobal.test.js
```diff
@@ -0,0 +1,32 @@
+const { execute } = require("../../lib");
+const { createMockStreams, resourceFile, allArgs } = require("./E2ETests");
+const lolex = require("lolex");
+
+describe("m.global usage in scenegraph components", () => {
+    let outputStreams;
+    let clock;
+
+    beforeAll(() => {
+        clock = lolex.install({ now: 1547072370937 });
```
sjbarag, line 10: I don't think we're doing anything in these tests that requires mocking out the current system time, so this (and all other references to `clock` in this file) can be removed 😃

PR #427, opened 2020-06-01
feat(stdlib): implement callFunc

### LINKED ISSUE #388: callFunc is not implemented
The current implementation of SceneGraph nodes in `brs` doesn't support `callFunc` to call methods defined in the XML interface.

lkipke:
# Change Summary
This implements `callFunc` and closes #388.

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -343,15 +343,11 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
                 return interpreter.inSubEnv(subInterpreter => {
                     let functionToCall = subInterpreter.getCallableFunction(functionname.value);
 
-                    // Args for callFunc are optional, they're based on the function we're calling,
-                    // and we need to match the signature of that function, so we can't always forward functionargs.
-                    //
-                    // You _can_ pass "invalid" as an argument to a function, so we'll use Uninitialized
-                    // as a default value here to differentiate.
-                    if (functionargs === Uninitialized.Instance) {
-                        return functionToCall.call(subInterpreter);
-                    } else {
+                    // Determine whether the function should get arguments are not.
+                    if (functionToCall.getFirstSatisfiedSignature([functionargs])) {
                         return functionToCall.call(subInterpreter, functionargs);
+                    } else {
+                        return functionToCall.call(subInterpreter);
                     }
                 }, componentDef.environment);
             } else {
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -338,7 +338,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             returns: ValueKind.Dynamic,
         },
         impl: (interpreter: Interpreter, functionname: BrsString, functionargs: BrsType) => {
-            // We need to search the callee's environment for this function, not the caller's.
+            // We need to search the callee's environment for this function rather than the caller's.
             let componentDef = interpreter.environment.nodeDefMap.get(this.nodeSubtype);
             if (componentDef) {
                 return interpreter.inSubEnv(subInterpreter => {
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -141,6 +141,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             ifSGNodeFocus: [this.hasfocus, this.setfocus, this.isinfocuschain],
             ifSGNodeDict: [this.findnode, this.issamenode, this.subtype],
         });
+        this.appendMethod("callFunc", this.callFunc);
     }
 
     toString(parent?: BrsType): string {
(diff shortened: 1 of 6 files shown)


src/brsTypes/components/RoSGNode.ts
```diff
@@ -133,6 +133,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             ifSGNodeFocus: [this.hasfocus, this.setfocus, this.isinfocuschain],
             ifSGNodeDict: [this.findnode, this.issamenode, this.subtype],
         });
+        this.appendMethod("callFunc", this.callFunc);
```
alimnios72, line 136: I don't think this is documented anywhere but if you run `print findMemberFunction(node, "callfunc")` on RBI it will print `<Interface: ifSGNodeDict>`, can you add the method to that interface?

src/brsTypes/components/RoSGNode.ts
```diff
@@ -317,6 +318,33 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+        impl: (interpreter: Interpreter, functionname: BrsString, functionargs: BrsType) => {
+            // We need to search the callee's environment for this function rather than the caller's.
+            let componentDef = interpreter.environment.nodeDefMap.get(this.nodeSubtype);
+            if (componentDef) {
+                return interpreter.inSubEnv(subInterpreter => {
+                    let functionToCall = subInterpreter.getCallableFunction(functionname.value);
+                    return functionToCall.call(subInterpreter, functionargs);
+                }, componentDef.environment);
+            } else {
+                throw new Error(
```
alimnios72, line 341: This will stop execution of the script, right? That differ from the warning returned by RBI which doesn't stop execution.
```
=================================================================
Warning calling function in MyNode
no function interface specified for something
=================================================================
```

src/brsTypes/components/RoSGNode.ts
```diff
@@ -317,6 +318,33 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+                    `Error: Unable to find function "${functionname}" used in callFunc.`
+                );
+            }
```
alimnios72, line 344: Can we return `invalid` here? That's what I got from RBI in all cases (even when function doesn't exist)

lkipke: Hmmm I think it should return whatever the function returns (so if the function returns `boolean`, then this call should return `boolean`, etc). That should handle the `invalid` case if the function doesn't return anything.

But if the function doesn't exist (like you called out in a different comment), then we should `console.error` and return `invalid` instead of throwing an error. I'll change that!

src/brsTypes/components/RoSGNode.ts
lkipke: Great find, can do!

src/brsTypes/components/RoSGNode.ts
alimnios72: Oh yeah, you're right. Those invalids were actually returned by the functions :homerdisappears:

src/brsTypes/components/RoSGNode.ts
```diff
@@ -317,6 +317,45 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+            if (componentDef) {
+                return interpreter.inSubEnv(subInterpreter => {
+                    let functionToCall = subInterpreter.getCallableFunction(functionname.value);
+
+                    // Args for callFunc are optional, they're based on the function we're calling,
+                    // and we need to match the signature of that function, so we can't always forward functionargs.
+                    //
+                    // You _can_ pass "invalid" as an argument to a function, so we'll use Uninitialized
+                    // as a default value here to differentiate.
+                    if (functionargs == Uninitialized.Instance) {
```
lkipke, line 343: The comment pretty much sums it up, but let me know if it doesn't make sense or there's a better solution.

src/componentprocessor/index.ts
```diff
@@ -21,6 +21,14 @@ interface ComponentField {
     [key: string]: FieldAttributes;
 }
 
+interface FunctionAttributes {
+    name: string;
+}
+
+interface ComponentFunction {
```
sjbarag, line 28: superduper nit: you might want to call this `ComponentFunctions` since it's a map of function name to function-attributes!

src/componentprocessor/index.ts
```diff
@@ -175,12 +185,15 @@ async function processXmlTree(
  * @param node Xml node with fields
  * @return The fields parsed as ComponentField
  */
-function getFields(node: XmlDocument): ComponentField {
+function processInterface(
```
sjbarag, line 188: Mind updating the JSDoc when you get a second?

src/brsTypes/components/RoSGNode.ts
```diff
@@ -325,6 +325,43 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+                    if (functionToCall.getFirstSatisfiedSignature([functionargs])) {
+                        return functionToCall.call(subInterpreter, functionargs);
+                    } else {
+                        return functionToCall.call(subInterpreter);
+                    }
+                }, componentDef.environment);
+            } else {
+                console.error(
+                    `Warning calling function in ${this.nodeSubtype}: no function interface specified for ${functionname}`
+                );
```
sjbarag, line 358: Pro tip for next time: if you use `interpreter.stderr.write` instead of `console.error`, you won't have to mock anything in your tests!  Just `outputStreams.stderr` the same way we handle `stdout` in the e2e tests 😄 

I'll update this line before I merge though!

src/componentprocessor/index.ts
lkipke: Good call, I was following the pattern of `ComponentField` above, but probably just want to update both those.

src/brsTypes/components/RoSGNode.ts
lkipke: done!

src/componentprocessor/index.ts
lkipke: done!

src/componentprocessor/index.ts
lkipke: done!

PR #431, opened 2020-06-04
feat: expose m.global in _brs_

lkipke:
# Change Summary
This exposes the `m.global` node (implemented in #425) in the `_brs_` object so that unit tests can easily access it. It also modifies the unit tests to include tests for `_brs_.global`.


src/brsTypes/components/RoSGNode.ts
```diff
@@ -1104,6 +1102,9 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
     }
 }
 
+// A node that represents the m.global, referenced by all other nodes
+export let mGlobal: RoSGNode = new RoSGNode([]);
```
lkipke, line 1106: I had to move this down here because Typescript didn't want me to instantiate an `RoSGNode` above the class definition.

lkipke:
@niveditasonker github wouldn't let me tag you as a reviewer, but since you just implemented `m.global`, thought I'd add you here 😄

PR #432, opened 2020-06-04
fix(rsg): Make types initialize a little better

lkipke:
# Change Summary
Before this change, when you had something like this:
```
node = createObject("RoSGNode", "Node")
node.addField("myArray", "assocarray")
node.myArray = { abcd: 1234 }
```
then `brs` would not allow the assignment of `node.myArray = { abcd: 1234 }`. This was because it was reading `node.myArray`'s value kind as `Invalid`, which didn't match the value kind that was being assigned (associative array is an `Object`), so `brs` didn't let the new value get set. This was only happening for arrays and associative arrays because of how `brs` created their initial types.

There were a lot of tests that had to be updated, sorry, I realize this changes a lot of test files!

### DIFF
--- a/src/brsTypes/BrsType.ts
+++ b/src/brsTypes/BrsType.ts
@@ -121,7 +121,7 @@ export function getValueKindFromFieldType(type: string) {
         case "string":
             return ValueKind.String;
         case "function":
-            return ValueKind.Callable
+            return ValueKind.Callable;
         case "node":
         case "roarray":
         case "array":
--- a/src/brsTypes/BrsType.ts
+++ b/src/brsTypes/BrsType.ts
@@ -102,7 +102,8 @@ export namespace ValueKind {
 
 /**
  *  Converts a specified brightscript type in string into BrsType representation, with actual value
- *  Note: only supports native types so far.  Objects such as array/AA aren't handled at the moment.
+ *  Note: only native types are fully supported so far. Objects such as node/array/AA are created with default values,
+ *  instead of the value that gets passed in.
  *  @param {string} type data type of field
  *  @param {string} value optional value specified as string
  *  @returns {BrsType} BrsType value representation of the type
--- a/src/brsTypes/BrsType.ts
+++ b/src/brsTypes/BrsType.ts
@@ -5,6 +5,8 @@ import { Int32 } from "./Int32";
 import { Float } from "./Float";
 import { roBoolean } from "./components/RoBoolean";
 import { roInvalid } from "./components/RoInvalid";
+import { RoAssociativeArray } from "./components/RoAssociativeArray";
+import { RoArray } from "./components/RoArray";
 
 /** Set of values supported in BrightScript. */
 export enum ValueKind {
@@ -123,13 +125,13 @@ export function getBrsValueFromFieldType(type: string, value?: string): BrsType 
         case "float":
             returnValue = value ? new Float(Number.parseFloat(value)) : new Float(0);
             break;
-        case "roArray":
+        case "roarray":
         case "array":
-            returnValue = BrsInvalid.Instance;
+            returnValue = new RoArray([]);
             break;
-        case "roAssociativeArray":
+        case "roassociativearray":
         case "assocarray":
-            returnValue = BrsInvalid.Instance;
+            returnValue = new RoAssociativeArray([]);
             break;
         case "uri":
         case "str":
--- a/src/brsTypes/BrsType.ts
+++ b/src/brsTypes/BrsType.ts
@@ -101,6 +101,39 @@ export namespace ValueKind {
 }
 
 /**
+ * Converts the data type of a field to the appropriate value kind.
+ * @param type data type of field
+ */
+export function getValueKindFromFieldType(type: string) {
+    switch (type.toLowerCase()) {
+        case "bool":
+        case "boolean":
+            return ValueKind.Boolean;
+        case "int":
+        case "integer":
+            return ValueKind.Int32;
+        case "longinteger":
+            return ValueKind.Int64;
+        case "float":
+            return ValueKind.Float;
+        case "uri":
+        case "str":
+        case "string":
+            return ValueKind.String;
+        case "function":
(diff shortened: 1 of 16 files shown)


src/brsTypes/BrsType.ts
```diff
@@ -123,13 +125,13 @@ export function getBrsValueFromFieldType(type: string, value?: string): BrsType
         case "float":
             returnValue = value ? new Float(Number.parseFloat(value)) : new Float(0);
             break;
-        case "roArray":
```
lkipke, line 126: We're calling `type.toLowerCase()` on [line 113](https://github.com/sjbarag/brs/pull/432/files#diff-cef4e231ed0b97ed2e188d7534bed1d2R113), so this should be lowercase instead of camelcase.

src/brsTypes/BrsType.ts
```diff
@@ -123,13 +125,13 @@ export function getBrsValueFromFieldType(type: string, value?: string): BrsType
...
+        case "roarray":
         case "array":
-            returnValue = BrsInvalid.Instance;
+            returnValue = new RoArray([]);
```
lkipke, line 130: This still doesn't technically do everything it should (it ignores the `value` parameter and always creates empty arrays/associative arrays). But it unblocks the scenario I put in the description for this PR, and it gets us _closer_ to complete.

Piera: So is the comment on line 105 still accurate?  "...Objects such as array/AA aren't handled at the moment."

test/brsTypes/components/ArrayGrid.test.js
```diff
@@ -9,27 +9,27 @@ describe("ArrayGrid", () => {
...
     id: 
     visible: true
     opacity: 1
-    translation: invalid
+    translation: <Component: roArray>
     rotation: 0
-    scale: invalid
-    scalerotatecenter: invalid
+    scale: <Component: roArray>
+    scalerotatecenter: <Component: roArray>
```
Piera, line 21: For all of these components, how did you know which of these members were of type `<Compenent: roArray>`?

src/brsTypes/BrsType.ts
lkipke: Good callout! I can edit that comment to be more clear that they're partially handled, but not fully 👍

test/brsTypes/components/ArrayGrid.test.js
lkipke: Good question! In a component definition, we have a list of default fields, and each field has a `type` (see [here](https://github.com/sjbarag/brs/blob/master/src/brsTypes/components/Group.ts#L8) for an example). That `type` corresponds to the [Roku documentation](https://developer.roku.com/docs/references/scenegraph/layout-group-nodes/group.md) for whichever component we're creating.

src/brsTypes/components/RoAssociativeArray.ts
```diff
@@ -125,7 +125,7 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
             returns: ValueKind.Boolean,
         },
         impl: (interpreter: Interpreter, str: BrsString) => {
-            let deleted = this.elements.delete(str.value);
+            let deleted = this.elements.delete(str.value.toLowerCase());
```
lkipke, line 128: Added this because I noticed it as I was testing. We're storing all the fields as lowercase, so we need to access them that way as well.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -216,8 +217,13 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
 
         if (!field) {
             field = new Field(value, alwaysNotify);
-        } else if (field.getType() === valueType) {
-            //Fields are not overwritten if they haven't the same type
+        } else if (
+            field.getType() === ValueKind.toString(ValueKind.Invalid) ||
```
lkipke, line 221: This allows us to set a field that was previously `invalid` or `uninitialized`.

lkipke:
@alimnios72 Just a heads up, I made a couple more logic changes here around handling `invalid` or `uninitialized` as a special case.

src/brsTypes/BrsType.ts
```diff
@@ -123,13 +159,13 @@ export function getBrsValueFromFieldType(type: string, value?: string): BrsType
...
             break;
-        case "roAssociativeArray":
+        case "roassociativearray":
         case "assocarray":
-            returnValue = BrsInvalid.Instance;
+            returnValue = new RoAssociativeArray([]);
```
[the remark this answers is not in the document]
sjbarag: Nah, this seems fine!  Thanks for calling it out :)

src/brsTypes/components/RoSGNode.ts
```diff
@@ -251,18 +324,24 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
     }
 
-    set(index: BrsType, value: BrsType, alwaysNotify: boolean = false) {
+    set(index: BrsType, value: BrsType, alwaysNotify: boolean = false, kind?: FieldKind) {
         if (index.kind !== ValueKind.String) {
             throw new Error("RoSGNode indexes must be strings");
         }
+
+        let fieldType = kind || FieldKind.fromBrsType(value);
+        // RBI does not set a field if the new value isn't valid
```
lkipke, line 333: As far as I could tell from playing around with it in RBI, this was true for all of the functions that could set a field's value: `addField`, `addFields`, `update`, `setField`, `setFields`. If you're trying to add a field (i.e. `update` or similar), and the value is `invalid`, the field won't get added at all. If you're trying to set a new value on an existing field (`myNode.foo = invalid`), then it won't set the new value to `invalid`.

Buuut if anyone has counter-evidence pls let me know

lkipke: ok so it turns out you _can_ set a `node`, `assocarray`, or `array` to invalid in RBI. Let me update this

PR #436, opened 2020-06-08
feat(stdlib): implement ContentNode

lkipke:
# Change Summary
This implements `ContentNode`.

`ContentNode` is a bit unique in how it manages its fields. Like most components, it has a set of built-in fields that it supports. Unlike most other components, these fields are all "hidden" from the node by default. Then, when a field is accessed in any way, it becomes "visible". The unit tests show this behavior as well, but for clarity, here's an example:
```
node = createObject("RoSGNode", "ContentNode")
print node
' Notice that even though ContentNode has dozens of fields, none of them display.
'
' <Component: roSGNode:ContentNode> =
' {
'    change: <UNINITIALIZED>
'    focusable: false
'    focusedchild: invalid
'    id: 
'}

' Getting a field makes it show up (in all caps)
print node.title
print node.count()  ' => 5 (the node now includes title)
print node
' ...
' TITLE: ""
' ...

' Setting a field makes it show up (in all caps)
node.description = "some description"
print node.count()  ' => 6 (the node now includes description)
print node
' ...
' TITLE: ""
' DESCRIPTION: "some description"
' ...
```

Because of this unique behavior, `ContentNode` had to override some functions from `RoSGNode`, along with a few other changes to support the overrides.

### DIFF
--- a/src/brsTypes/components/BrsComponent.ts
+++ b/src/brsTypes/components/BrsComponent.ts
@@ -4,7 +4,7 @@ import { Callable } from "../Callable";
 import { BrsInterface } from "../BrsInterface";
 
 export class BrsComponent {
-    private methods: Map<string, Callable> = new Map();
+    private methods: Map<string, Callable> = new Map<string, Callable>();
     private readonly componentName: string;
 
     readonly interfaces = new Map<string, BrsInterface>();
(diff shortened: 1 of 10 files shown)


sjbarag:
I'm pretty sure I know how to resolve those conflicts now that #438 , but would you mind taking a look when you get a second?

lkipke:
> I'm pretty sure I know how to resolve those conflicts now that #438 , but would you mind taking a look when you get a second?

done!

PR #438, opened 2020-06-09
feat(stdlib): add RoSGNodeEvent

lkipke:
# Change Summary
This implements `RoSGNodeEvent`. Also, when we implemented `onChange` callbacks in #424, I forgot that those callbacks could optionally receive an event. This PR fixes that, so that now `onChange` can receive the event.

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -7,6 +7,7 @@ import {
     Uninitialized,
     getBrsValueFromFieldType,
 } from "../BrsType";
+import { RoSGNodeEvent } from "./RoSGNodeEvent";
 import { BrsComponent, BrsIterable } from "./BrsComponent";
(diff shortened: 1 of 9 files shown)


src/brsTypes/components/RoSGNode.ts
```diff
@@ -529,7 +552,13 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         impl: (interpreter: Interpreter, fieldname: BrsString, functionname: BrsString) => {
             let field = this.fields.get(fieldname.value.toLowerCase());
             if (field instanceof Field) {
-                field.addObserver(interpreter, interpreter.getCallableFunction(functionname.value));
+                let event = new RoSGNodeEvent(this, fieldname, field.getValue());
```
sjbarag, line 555: Looks like this isn't being used anymore 😉

PR #441, opened 2020-06-11
fix(stdlib): escape from value in string replace

lkipke:
# Change Summary
Before, we weren't escaping the `from` value in `string.replace`, which meant that `brs` was throwing an error due to an invalid `RegExp`. This PR adds an escape.

### DIFF
--- a/test/brsTypes/components/RoString.test.js
+++ b/test/brsTypes/components/RoString.test.js
@@ -284,6 +284,18 @@ describe("RoString", () => {
                     )
                 ).toEqual(new BrsString("tossed salad and scrambled eggs"));
             });
+
+            it("escapes strings with reserved regex characters", () => {
+                let s = new RoString(new BrsString("oh baby {1}"));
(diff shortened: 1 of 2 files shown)


test/brsTypes/components/RoString.test.js
```diff
@@ -284,6 +284,18 @@ describe("RoString", () => {
...
+            it("escapes strings with reserved regex characters", () => {
+                let s = new RoString(new BrsString("oh baby {1}"));
+                replace = s.getMethod("replace");
+                expect(
+                    replace.call(
+                        interpreter,
+                        new BrsString("{1}"),
+                        new BrsString("I hear the blues a-callin'")
+                    )
+                ).toEqual(new BrsString("oh baby I hear the blues a-callin'"));
```
sjbarag, line 297: I think I was binging through _Frasier_ while I built `roString`?  I genuinely can't remember the timeline anymore.

PR #442, opened 2020-06-11
Allow mocking of "unknown" functions

lkipke:
# Change Summary
This change allows us to mock functions that don't exist in source. This seems somewhat counter-intuitive: why would we let a test implement a function that doesn't exist in source, yet gets called from source code?

The reason is because `brs` doesn't have all the built-in global functions implemented yet, so this unblocks the testing of source code that calls one of those not-yet-implemented functions. The example I was running into was `tr` -- I wanted to test the code that called `tr`, but `tr` isn't implemented, and it also wasn't mockable, so I had no way to test it.

Also, maybe this would be useful for TDD?

Feel free to reject this if `tr` is a one-off situation or there's a better way. I wasn't sure how many other similar cases there were!

### DIFF
--- a/src/interpreter/Environment.ts
+++ b/src/interpreter/Environment.ts
@@ -2,7 +2,6 @@ import { Identifier } from "../lexer";
 import { BrsType, RoAssociativeArray, Int32, BrsInvalid, RoSGNode, Callable } from "../brsTypes";
 import { ComponentDefinition } from "../componentprocessor";
 import { BrsError } from "../Error";
-// import { Interpreter } from ".";
 
 /** The logical region from a particular variable or function that defines where it may be accessed from. */
 export enum Scope {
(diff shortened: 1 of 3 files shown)


src/interpreter/Environment.ts
```diff
@@ -172,6 +173,26 @@ export class Environment {
...
+            // allow mocking of functions that "don't exist" in source. This unblocks
+            // unit testing of code that calls the not-yet-implemented global functions.
+            //
+            // TODO: remove this once all built-in global functions are implemented?
+
+            // Don't allow mocks of reserved functions like "init" and "main".
+            if (!name.isReserved) {
+                let mockedFunc = this.getMockFunction(lowercaseName);
+                if (mockedFunc !== BrsInvalid.Instance) {
+                    console.error(
```
lkipke, line 187: I went with `console.error` here because importing `Interpreter` caused circular dependency chaos.

sjbarag: Boooooo.  That's fair.  We'll eventually get a working solution for that.

PR #443, opened 2020-06-11
fix(parser): make if statements with 'not' work

### LINKED ISSUE #329: IF statement with NOT before a logic comparison expression shows an error
The following code generates an error:
```
foo = 5
if not foo = 1
    print "foo is not 1"
end if
```
If you add parenthesis it works:
```
foo = 5
if not (foo = 1)
    print "foo is not 1"
end if
```

lkipke:
# Change Summary
This allows `if` statements with a logical `not` operator, and closes #329.

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -1316,7 +1316,7 @@ export class Parser {
         function prefixUnary(): Expression {
             if (match(Lexeme.Not, Lexeme.Minus)) {
                 let operator = previous();
-                let right = prefixUnary();
+                let right = expression();
                 return new Expr.Unary(operator, right);
             }
 
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -1316,7 +1316,7 @@ export class Parser {
         function prefixUnary(): Expression {
             if (match(Lexeme.Not, Lexeme.Minus)) {
                 let operator = previous();
-                let right = expression();
+                let right = relational();
                 return new Expr.Unary(operator, right);
             }
 
--- a/test/e2e/resources/conditionals.brs
+++ b/test/e2e/resources/conditionals.brs
@@ -54,7 +54,52 @@ aa = {}
 if foo > 5 then aa.value = ">5" else aa.value = 8
 print aa.value
 
+' if not with boolean
+if not true then
+    print "this will never print"
+end if
+
+if not false then
+    print "not false"
+end if
+
+' if not with an expression
 bar = "abc"
 if not bar = "def" then
-    print "not equal"
+    print "bar does not equal 'def'"
+end if
+
+' if not with or
+if not bar = "def" or false then
+    print "if not with or variation 1"
+end if
+
+if not bar = "abc" or true then
+    print "if not with or variation 2"
+end if
+
+' if not with and
+if not bar = "def" and true then
+    print "if not with and"
+end if
+
+' if not with two expressions
+foo = 1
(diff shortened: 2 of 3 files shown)


test/e2e/resources/conditionals.brs
```diff
@@ -53,3 +53,53 @@ end if
...
+end if
+
+' if not with and
+if not bar = "def" and true then
+    print "if not with and"
+end if
+
+' if not with two expressions
+foo = 1
+if not bar = "def" and foo = 1 then
```
Piera, line 88: nit/not sure it matters but in the rest of this file, relational operators are used (e.g., `if foo < 5` on line 46) instead of equality operators (...`and foo = 1`).  I only bring it up bc of the specific change in Parser.ts (`let right = relational()`).

test/e2e/resources/conditionals.brs
```diff
@@ -53,3 +53,53 @@ end if
...
+end if
+
+' if not multiple times
+foo = 1
+if not bar = "def" and not foo = 2 then
+    print "if not multiple times"
+end if
+
+' if not with <> operator
+if not bar <> "abc" then
```
Piera, line 103: nit/related to comment above: we're calling out an "inequality" operator here; should there be coverage for relational operators too?

test/e2e/resources/conditionals.brs
lkipke: Great callouts! I will definitely add tests for other operators 👍

src/parser/Parser.ts
```diff
@@ -1316,7 +1316,7 @@ export class Parser {
         function prefixUnary(): Expression {
             if (match(Lexeme.Not, Lexeme.Minus)) {
                 let operator = previous();
-                let right = prefixUnary();
+                let right = relational();
```
lkipke, line 1319: Considering the order of operations goes `relational` > `bitshift` > `additive` > `multiplicative` > `exponential` > `prefixUnary`, I was going to add some unit tests that did things like `not 2 * 3`. However, that elicits strange behavior on RBI (`not 2 * 3 = -7` for some reason...) and `brs` prevents comparing boolean `NOT` to non-boolean `2 * 3`.

Piera: 🤔  interesting!

sjbarag: Hmm, seems like there's still some bugs in `brs` that we'll need to shake out then! From RBI:

```brightscript
print not  0 ' => -1
print not  1 ' => -2
print not -1 ' =>  0
print not  2 ' => -3
print not -2 ' =>  1
```

which looks like RBI is using a [two's complement](https://en.wikipedia.org/wiki/Signed_number_representations#Two's_complement) representation for numbers and does something slightly unexpected with it.  `not someInteger` in RBI seems to take `someInteger`, convert to a binary representation, perform a bitwise negation on it (all 0s become 1s, all 1s become 0s), then convert back to a decimal representation in the Two's complement understanding.  So for `not 0` (using just 4-bits for now because I'm lazy), we have:

```
not 0 (decimal) == not 0b0000 == 0b1111 == -1 (decimal in two's complement)
```

How strange!  In the `not 2 * 3` example, we get:

```
not (2 * 3) == not 6 (decimal) == not 0b0110 == 0b1001 == -7 (decimal in two's complement)
```

All-up, it looks like RBI supports negating decimals, but `brs` doesn't!  I'll file a bug to fix that a bit later.  At least the parser is still correct 🎉

Issue #448, opened 2020-06-23
Inherited interface fields that reference a non-existent callback results in exception


kirkyoder:
When a derived component object is created whose parent's interface defines observable fields (e.g. `onChange`), an exception will be thrown unless the given observable's callback exists on the derived component.

#### `DerivedComponent.xml`
```xml
<component name="DerivedComponent" extends="ParentComponent">
</component>
```

#### `DerivedComponent.brs`
```brs
sub init()
    ' some initialization logic here
end sub

' Note that no "exitPage" subroutine is defined
```

#### `ParentComponent.xml`
```xml
<component name="ParentComponent">
    <interface>
        <field id="exitPage" type="boolean" onChange="exitPage" alwaysNotify="true" />
    </interface>
</component>
```

## Exception

```
Runtime error encountered in BRS implementation:

Error: exitPage was not found in scope
    at Interpreter.getCallableFunction (/Users/kirk.yoder/enlistments/cube-roku/node_modules/brs/lib/interpreter/index.js:210:15)
    at impl (/Users/kirk.yoder/enlistments/cube-roku/node_modules/brs/lib/brsTypes/components/RoSGNode.js:324:64)
    at interpreter.inSubEnv.subInterpreter (/Users/kirk.yoder/enlistments/cube-roku/node_modules/brs/lib/brsTypes/Callable.js:107:20)
    at Interpreter.inSubEnv (/Users/kirk.yoder/enlistments/cube-roku/node_modules/brs/lib/interpreter/index.js:138:20)
    at Callable.call (/Users/kirk.yoder/enlistments/cube-roku/node_modules/brs/lib/brsTypes/Callable.js:97:28)
    at addFields (/Users/kirk.yoder/enlistments/cube-roku/node_modules/brs/lib/brsTypes/components/RoSGNode.js:1140:30)
    at interpreter.inSubEnv.subInterpreter (/Users/kirk.yoder/enlistments/cube-roku/node_modules/brs/lib/brsTypes/components/RoSGNode.js:1097:17)
    at Interpreter.inSubEnv (/Users/kirk.yoder/enlistments/cube-roku/node_modules/brs/lib/interpreter/index.js:138:20)
    at Object.createNodeByType (/Users/kirk.yoder/enlistments/cube-roku/node_modules/brs/lib/brsTypes/components/RoSGNode.js:1095:25)
[...]

PR #449, opened 2020-06-23
fix(interp,stdlib): getCallableFunction ignores non-implemented functions

kirkyoder:
Currently, when `getCallableFunction` attempts to resolve a callable field defined on a component's interface, if it is unable to resolve the given function (i.e. it's not implemented in the given component), an exception will be thrown and execution halted. BrightScript, however, silently fails whenever such a situation is encountered, permitting execution to proceed.

This change updates `brs` to mirror the BrightScript behavior such that a warning is logged to stdout, but execution is allowed to proceed, whenever a non-implemented callable field is encountered.

This change resolves https://github.com/sjbarag/brs/issues/448.

### DIFF
--- a/src/interpreter/index.ts
+++ b/src/interpreter/index.ts
@@ -253,7 +253,7 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
             return maybeCallback;
         }
 
-        console.warn(`${functionName} was not found in scope`);
+        console.warn(`Warning: "${functionName}" was not found in scope.`);
         return;
     }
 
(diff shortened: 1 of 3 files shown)


src/interpreter/index.ts
```diff
@@ -247,12 +247,14 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
...
             },
         });
+
         let maybeCallback = this.evaluate(callbackVariable);
         if (maybeCallback.kind === ValueKind.Callable) {
             return maybeCallback;
         }
 
-        throw new NotFound(`${functionName} was not found in scope`);
+        console.warn(`Warning: "${functionName}" was not found in scope.`);
```
kirkyoder, line 256: Let me know if `console.warn()` is appropriate, here, or if `stderr.write()` would be better.

sjbarag: I typically prefer `stderr.write()` so it doesn't appear in end-to-end test output (plus it's easily capturable/testable that way!)

Issue #416, opened 2020-05-14
toString() method not printing string fields with quotes


lkipke:
When a component has string fields and we call `toString()` on a component, we expect that any string fields it has get printed out as `""`. Instead, we are only seeing a print out of one space (` `).

# Expected behavior
```
ComponentWithStringFields.toString();
// <Component: roSGNode:ComponentWithStringFields> =
// {
//     id: ""
//     otherStringField: "" 
// }
```

# Actual behavior
```
ComponentWithStringFields.toString();
// <Component: roSGNode:ComponentWithStringFields> =
// {
//     id: 
//     otherStringField: 
// }
```

lkipke:
Note: this is the function that will need to be modified:
https://github.com/sjbarag/brs/blob/master/src/brsTypes/BrsType.ts#L218-L220

PR #450, opened 2020-06-23
feat(stdlib): Implement Timer Component

kirkyoder:
This change implements the SceneGraph [`Timer`](https://developer.roku.com/docs/references/scenegraph/control-nodes/timer.md) component.

### DIFF
--- a/test/brsTypes/components/Timer.test.js
+++ b/test/brsTypes/components/Timer.test.js
@@ -1,0 +1,24 @@
+const brs = require("brs");
+const { Timer } = brs.types;
+
+describe("Timer", () => {
+    describe("stringification", () => {
+        it("initializes a new Timer component", () => {
+            let timer = new Timer();
+
+            expect(timer.toString()).toEqual(
+                `<Component: roSGNode:Timer> =
+{
+    change: <UNINITIALIZED>
+    focusable: false
+    focusedchild: invalid
+    id: 
+    control: 
+    repeat: false
+    duration: 0
+    fire: <UNINITIALIZED>
+}`
+            );
+        });
+    });
+});
--- a/test/e2e/resources/components/Timer.brs
+++ b/test/e2e/resources/components/Timer.brs
@@ -1,0 +1,9 @@
+sub Main()
+    timer = createObject("roSGNode", "Timer")
+    print "timer node type:" type(timer)
+    print "timer node subtype:" timer.subtype()
+    print "timer node control:" timer.control
+    print "timer node repeat:" timer.repeat
+    print "timer node duration:" timer.duration
+    print "timer node fire:" timer.fire
+end sub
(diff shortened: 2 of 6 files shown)


kirkyoder:
> On second thought... This will allow you to create a `Timer`, but the `Timer` won't actually work the way that it does in RBI, right? As in, this won't allow you to trigger the timer to call your `fire` function.

That's right! I was mostly interested in just unblocking unit testing of components that rely on `Timer`.

I'm open to opinions, here, but I figured we'd just circle back once we actually needed the ability to test observer callback functions. 💭

test/brsTypes/components/Timer.test.js
```diff
@@ -0,0 +1,24 @@
...
+        it("initializes a new Timer component", () => {
+            let timer = new Timer();
+
+            expect(timer.toString()).toEqual(
+                `<Component: roSGNode:Timer> =
+{
+    change: <UNINITIALIZED>
+    focusable: false
+    focusedchild: invalid
+    id:
```
Piera, line 15: 🤔  Quick question -- `id` and `control` are both left blank -- how is this different than <UNINITIALIZED>?

lkipke: There are a couple of things going on here. Firstly, this technically shouldn't be blank, it's a string so it should be `""`. It's a minor bug in `brs`, and #416 is filed to address it.

The other thing happening is that `<UNINITIALIZED>`is the value you get when you add a field, but not a value for that field. In our current implementation, this is essentially what's happening:
```
myNode.addField("change", "function")
```
Note that we know `change` _exists_, but it doesn't have a value, so it's `UNINITIALIZED`. That being said, the fact that this is happening here is because we haven't implemented default values for complex types like `function`, `node`, `array`, etc. So in the future, `change` will actually get a default value instead of `UNINITIALIZED`. There's another open PR to partially address that.

Piera: Ah, got it!  Thanks for the explanation!

test/e2e/resources/components/Timer.brs
```diff
@@ -0,0 +1,9 @@
+sub Main()
```
cchrissy, line 1: noob question, what's the relationship between these updated files?

kirkyoder: So there are two sets of tests: (1) end-to-end tests that validate our expectations in a BrightScript environment and (2) unit tests that validate logical behaviors of the `brs` parser.

`test/e2e/BrsComponents.test.js` executes `test/e2e/resources/components/Timer.brs` and validates that the corresponding BrightScript logic spits out an expected set of strings.

`test/brsTypes/components/Timer.test.js` contains the unit tests, which are pretty sparse here, since we haven't added any additional logic beyond being able to create a `Timer` component.

kirkyoder:
@sjbarag, if you get a chance, could you give this guy your blessing and land it for me? 😄

Piera:
@sjbarag Possible to give Chrissy and I write access so that our approvals will be ~official~ indicated with little green check marks?

PR #455, opened 2020-07-02
feat(stdlib): Add mock component partial

### LINKED ISSUE #380: Support partial mocking of functions in-scope for a component
For testing purposes, it's sometimes helpful to mock out a function that's brought into scope from a SceneGraph component's `<script>` tag.  Those functions can lead to unexpected side-effects (e.g. network requests) or performance bottlenecks (e.g. prime factorization I guess?) that aren't strictly necessary when testing a component in isolation.

Consider adding a `_brs_.mockComponentPartial` or something that accepts an associative array of functions to be overwritten in the scope of components that are created later.

philmein23:
Prompt - https://github.com/sjbarag/brs/issues/380

This PR adds the ability to bring in multiple mock functions, via an associative array, into scope from SceneGraph's components `script` tag.

closes #380


test/e2e/resources/mockComponentPartial.brs
```diff
@@ -0,0 +1,42 @@
+sub main()
+    print formatJson({}) ' => {real: 'json'}
+    print get() ' => Status: 500
+    print post() ' => Status: 200
+    print isValid() ' => false
+
+    _brs_.mockComponentPartial({
+        formatJson: _brs_.mockFunction("formatJson", function(json as object, flags=0 as integer) as dynamic
```
lkipke, line 8: So I think that the reason this test is passing is because we're calling `_brs_.mockFunction` here. This is equivalent to making that call without also calling `_brs_.mockComponentPartial`:
```brightscript
print isValid() ' => false

_brs_.mockFunction("formatJson", function(json as object, flags=0 as integer) as dynamic
    return "{fake:'json'}"
end function)
...
```
The reason it works is because `_brs_.mockFunction` [mocks any non-local function](https://github.com/sjbarag/brs/blob/master/src/interpreter/Environment.ts#L163-L196), so it mocks the functions you declared at the bottom of this file.

I think the way that we'd probably want to test this is something like:
```
_brs_.mockComponentPartial({
        formatJson: function(json as object, flags=0 as integer) as dynamic
            return "{fake:'json'}"
        end function)
})
```
The difference here is that we're passing in an associative array, where the values are the mock functions themselves, rather than calling `_brs_.mockFunction`.

philmein23:
> Given that `_brs_.mockFunction` handles mocking any module or global function, can someone remind me what the use case is for `_brs_.mockComponentPartial`? I'm thinking that either the issue this PR addresses might no longer be needed, or maybe there is an environment-conflict where `_brs_.mockFunction` registers a mock in a different environment than the one we need it mocked in.

I think the difference between this and `_brs_.mockFunction` is that we are able to mock multiple functions at once, but also mock the ones we bring into scope from other components via the script tag. For example, Component A imports two utility components via script tag - Http.brs and Types.brs - so now Component A has access to those functions. 

With that being said, it looks like there is code that already sets up the overarching environment where if Component A imports a few components, then the scope of each of those components will already be accessible by Component A. With that assumption in mind is what led to the code you see now. :D

src/extensions/mockComponentPartial.ts
```diff
@@ -0,0 +1,32 @@
...
+            [...mockFunctions.elements]
+                .filter(([_, value]) => value.kind === ValueKind.Callable)
+                .forEach(([key, value]) => {
+                    if (value instanceof Callable) {
+                        interpreter.environment.setMockFunction(key.toString(), value);
+                    }
+                });
+        } else {
+            interpreter.stderr.write(
+                "mockFunctions paramter needs to be an instance of roAssociativeArray"
```
lkipke, line 27: nit: `parameter`

test/e2e/BrsComponents.test.js
```diff
@@ -657,4 +657,22 @@ describe("end to end brightscript functions", () => {
             "Node",
         ]);
     });
+
+    test("components/scripts/MockFUnctionsMain.brs", async () => {
```
lkipke, line 661: nit: `MockFunctionsMain.brs`

test/e2e/resources/components/ComponentMockFunctions.xml
```diff
@@ -0,0 +1,14 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<component name="ComponentMockFunctions">
+    <script type="text/brightscript" uri="pkg:/components/scripts/ComponentMockFunctions.brs" />
+    <script type="text/brightscript" uri="pkg:/components/scripts/MockFunctionsUtils-A.brs" />
+    <script type="text/brightscript" uri="pkg:/components/scripts/MockFunctionsUtils-B.brs" />
+    <interface>
+        <field id="json" type="string" />
```
lkipke, line 7: It doesn't look like we're using these fields, unless i'm mistaken

src/extensions/mockComponentPartial.ts
```diff
@@ -0,0 +1,32 @@
...
+    BrsType,
+    ValueKind,
+    Callable,
+    StdlibArgument,
+    BrsInvalid,
+    RoAssociativeArray,
+} from "../brsTypes";
+import { Interpreter } from "../interpreter";
+
+export const mockComponentPartial = new Callable("mockComponentPartial", {
```
lkipke, line 11: I think it might make more sense to call this `mockFunctions` to align with the existing `_brs_.mockFunction`

test/e2e/resources/components/ComponentMockFunctions.xml
philmein23: Whoops, you're right. I initially included them for another purpose but forgot to remove those as the use-case was no longer needed. :)

philmein23:
> Looks awesome!! Thanks for doing this!

Thanks for helping out man! Really appreciate your great feedback.

src/extensions/mockFunctions.ts
```diff
@@ -0,0 +1,44 @@
...
+    BrsString,
+    ValueKind,
+    Callable,
+    StdlibArgument,
+    BrsInvalid,
+    RoAssociativeArray,
+} from "../brsTypes";
+import { Interpreter } from "../interpreter";
+
+export const mockFunctions = new Callable("mockFunctions", {
```
lkipke, line 12: Ok, I realize that I was the one to suggest we call this `mockFunctions` instead of `mockComponentPartial`... but given Sean's feedback and the corresponding changes, we _are_ partially mocking a component. So my vote is actually to change this back to `mockComponentPartial` (i'm the worst) 😬

src/extensions/mockFunctions.ts
```diff
@@ -0,0 +1,44 @@
...
+            interpreter.stderr.write(
+                `Unable to mock functions for unknown roSGNode subtype ${subtype.value}`
+            );
+            return BrsInvalid.Instance;
+        }
+        if (mockFunctions instanceof RoAssociativeArray) {
+            [...mockFunctions.elements]
+                .filter(([_, value]) => value.kind === ValueKind.Callable)
+                .forEach(([key, value]) => {
+                    if (value instanceof Callable) {
```
lkipke, line 33: Do we just need this `if` statement to make the compiler happy? I'm curious how we would run into a scenario where we'd have `value.kind === ValueKind.Callable` but not `value instanceof Callable`

src/interpreter/Environment.ts
```diff
@@ -249,12 +249,15 @@ export class Environment {
     public createSubEnvironment(includeModuleScope: boolean = true): Environment {
         let newEnvironment = new Environment(this.rootM);
         newEnvironment.global = new Map(this.global);
-        newEnvironment.module = includeModuleScope
-            ? new Map(this.module)
-            : new Map<string, BrsType>();
         newEnvironment.mPointer = this.mPointer;
         newEnvironment.mockObjects = this.mockObjects;
-        newEnvironment.mockFunctions = this.mockFunctions;
```
lkipke, line 257: nice catch!

src/extensions/mockFunctions.ts
philmein23: haha, now that I think about it once more it really makes sense- will revert the naming to how it was before.

src/extensions/mockFunctions.ts
philmein23: Right. I mean if we are already filtering for `ValueKind.Callable` then we should not need an `if` check for `Callable` in the next expression. I think I added that to satisfy the compiler as you said.

src/interpreter/Environment.ts
philmein23: Shoutout to @sjbarag for that.

philmein23:
> This looks really great! This will be awesome to have!
> 
> Could we actually do one more unit test, with inheritance? I think that a subclass should also respect the mocked functions, because any logic within the superclass, even if mocked, is still the logic of the subclass. Looking at the code, I think that this should already work, but it'd be nice to have a test for it!
> 
> Maybe something like:
> 
> ```
> <component name="SubclassWithMocks" extends="BaseWithMocks">
> ...
> ```
> 
> ```brightscript
> ' inside BaseWithMocks.brs
> 
> function callMe()
>   return "super"
> end function
> ```
> 
> ```brightscript
> ' inside SubclassWithMocks.brs
> 
> function callMe()
>   return "sub"
> end function
> ```
> 
> and then mocking it:
> 
> ```brightscript
> sub main()
>     _brs_.mockComponentPartial("BaseWithMocks", {
>         callMe: sub()
>             return "mock"
>         end sub
>     })
> 
>     c = createObject("roSGNode", "BaseWithMocks") ' calls to `callMe()` should return "mock"
>     t = createObject("roSGNode", "SubclassWithMocks")  ' calls to `callMe()` should _also_ return "mock" because it inherits that functionality
> ```

@lkipke @sjbarag  - Currently working on this part but I am stuck at the moment. Seems like calling the same function in the subclass returns the real implementation rather than "mock". I took a look at the `ComponentDefinition` of it, and although the subclass inherits the superclass's functions via the `module` field, the `mockFunctions` field contains an empty map. I suspect that's the reason and there should be a way to copy `mockFunctions` from the superclass to that of the subclass.

src/extensions/mockFunctions.ts
sjbarag: Yep, pretty sure that's a compiler affordance, but I suspect there's a better way around it :)

philmein23:
> This looks really great! This will be awesome to have!
> 
> Could we actually do one more unit test, with inheritance? I think that a subclass should also respect the mocked functions, because any logic within the superclass, even if mocked, is still the logic of the subclass. Looking at the code, I think that this should already work, but it'd be nice to have a test for it!
> 
> Maybe something like:
> 
> ```
> <component name="SubclassWithMocks" extends="BaseWithMocks">
> ...
> ```
> 
> ```brightscript
> ' inside BaseWithMocks.brs
> 
> function callMe()
>   return "super"
> end function
> ```
> 
> ```brightscript
> ' inside SubclassWithMocks.brs
> 
> function callMe()
>   return "sub"
> end function
> ```
> 
> and then mocking it:
> 
> ```brightscript
> sub main()
>     _brs_.mockComponentPartial("BaseWithMocks", {
>         callMe: sub()
>             return "mock"
>         end sub
>     })
> 
>     c = createObject("roSGNode", "BaseWithMocks") ' calls to `callMe()` should return "mock"
>     t = createObject("roSGNode", "SubclassWithMocks")  ' calls to `callMe()` should _also_ return "mock" because it inherits that functionality
> ```

I'm going to leave this set of unit tests out for now as references to the BaseClass::CallMe is not being respected when it's attempting to be mocked out.

sjbarag:
@philmein23 if you get a second, would you mind rebasing on top of `master`?  #460 changed the default style of the project (I didn't want to upgrade prettier at the time but I had to to enable a typescript 3.8 feature 😢), so your builds are currently failing because of the change in formatting.

philmein23:
> @philmein23 if you get a second, would you mind rebasing on top of `master`? #460 changed the default style of the project (I didn't want to upgrade prettier at the time but I had to to enable a typescript 3.8 feature 😢), so your builds are currently failing because of the change in formatting.

Wow, that took a minute but it should be good to go now.

test/extensions/MockComponentPartial.test.js
```diff
@@ -0,0 +1,92 @@
...
+const { ValueKind, Callable, BrsString, RoAssociativeArray } = brs.types;
+
+const { ComponentDefinition } = require("../../lib/componentprocessor");
+
+const { Interpreter } = require("../../lib/interpreter");
+const { mockComponentPartial } = require("../../lib/extensions/mockComponentPartial");
+const { BrsInvalid } = require("../../lib/brsTypes/BrsType");
+
+describe("mockComponentPartial", () => {
+    describe("methods", () => {
```
sjbarag, line 11: Probably don't need this inner `describe` since there's only one thing inside it, but I'll remove it before I merge 😃

PR #459, opened 2020-07-14
fix(err): Silence "Something terrible happened" errors

### LINKED ISSUE #444: Get rid of the "Something terrible happened messages"
When something terrible happens, we throw an error with "Something terrible happened ... " as the message. Get rid of those messages, and instead throw the actual error.

elsassph:
Just rethrow error: BrsErrors will be silent, other exceptions will be shown.
Ensure errors aren't logged multiple times.

Fixes #444

### DIFF
--- a/src/Error.ts
+++ b/src/Error.ts
@@ -28,7 +28,7 @@ export class BrsError extends Error {
             formattedLocation = `${location.file}(${location.start.line},${location.start.column},${location.end.line},${location.end.line})`;
         }
 
-        return `${formattedLocation}: ${this.message}`;
+        return `${formattedLocation}: ${this.message}\n`;
     }
 }
 
--- a/src/interpreter/index.ts
+++ b/src/interpreter/index.ts
@@ -222,8 +222,9 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
             } else {
                 throw err;
             }
+        } finally {
+            return results;
         }
-        return results;
     }
 
     getCallableFunction(functionName: string): Callable | undefined {
--- a/src/interpreter/index.ts
+++ b/src/interpreter/index.ts
@@ -176,9 +176,6 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
             this._environment = newEnv;
             return func(this);
         } catch (err) {
-            if (!(err instanceof BrsError)) {
-                throw `Runtime error encountered in BRS implementation: ${err}`;
-            }
             throw err;
         } finally {
             this._environment = originalEnvironment;
--- a/src/interpreter/index.ts
+++ b/src/interpreter/index.ts
@@ -997,9 +997,8 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
                 });
             } catch (reason) {
                 if (!(reason instanceof Stmt.BlockEnd)) {
-                    throw new Error(
-                        "Something terrible happened and we didn't throw a `BlockEnd` instance."
-                    );
+                    // re-throw interpreter errors
+                    throw reason;
                 }
 
                 let returnedValue = (reason as Stmt.ReturnValue).value;
--- a/src/interpreter/index.ts
+++ b/src/interpreter/index.ts
@@ -177,7 +177,7 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
             return func(this);
         } catch (err) {
             if (!(err instanceof BrsError)) {
-                console.error("Runtime error encountered in BRS implementation: ", err);
+                throw `Runtime error encountered in BRS implementation: ${err}`;
             }
             throw err;
         } finally {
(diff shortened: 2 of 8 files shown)


bin/cli.js
```diff
@@ -22,10 +22,9 @@ program
             try {
                 await brs.execute(brsFiles, { root: program.root });
             } catch (err) {
-                if (err.messages && err.messages.length) {
-                    err.messages.forEach(message => console.error(message));
-                } else {
-                    console.error(err.message);
+                // rethrow unhandled errors
+                if (!err.location && !err.messages) {
+                    throw err;
```
sjbarag, line 27: Any particular reason to re-throw errors here?  It's unclear to me what that provides over logging `err.message` like we do, since this file is only executed from a shell via `brs <args>`.

src/Error.ts
```diff
@@ -28,7 +28,7 @@ export class BrsError extends Error {
             formattedLocation = `${location.file}(${location.start.line},${location.start.column},${location.end.line},${location.end.line})`;
         }
 
-        return `${formattedLocation}: ${this.message}`;
+        return `${formattedLocation}: ${this.message}\n`;
```
sjbarag, line 31: Perhaps I'm missing something, but I'm not sure what this has to do with suppressing the "something terrible happened" errors.  What's the motivation behind this change?

src/interpreter/index.ts
```diff
@@ -225,9 +222,8 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
             } else {
                 throw err;
             }
-        } finally {
-            return results;
         }
+        return results;
```
sjbarag, line 226: Looks like this change is what changed the behavior for `RunInScope`, right?

test/cli/cli.test.js
```diff
@@ -32,4 +32,29 @@ describe("cli", () => {
         });
         expect(stdout.trim()).toEqual("hi from foo()");
     });
+
+    it("throws syntax errors once", async () => {
+        let filename = "errors/syntax-error.brs";
+        let stderr = await runWithErrors(filename);
+        expect(stderr.split(filename).length).toEqual(3);
```
sjbarag, line 39: Mind adding a comment to explain why we expect this to be 3?  It's not very clear that we're checking the number of stack frames by counting the chunks around the executed filename 🤔

test/extensions/RunInScope.test.js
```diff
@@ -51,11 +51,13 @@ describe("global Run function", () => {
...
+    it("throws an exception for runtime errors", () => {
         fs.readFileSync.mockImplementationOnce(() => `sub main(): _ = {}: _.crash(): end sub`);
-        expect(RunInScope.call(interpreter, new BrsString("pkg:/errors/exec.brs"))).toBe(
-            BrsInvalid.Instance
-        );
+        try {
+            RunInScope.call(interpreter, new BrsString("pkg:/errors/exec.brs"));
+        } catch (err) {
+            expect(err.message).toBe("'crash' is not a function and cannot be called.");
+        }
```
sjbarag, line 60: Interestingly, this is actually not the behavior we want here!  Both the global `Run` and the custom `_brs_.runInScope` need to return `invalid`, or tests via `@hulu/roca` will abort immediately upon the first syntax error, causing malformed [tap](https://testanything.org/) output and broken formatting into the Mocha "spec" format.

This test breaking in 95ba0ac2 caught that change! 🎉

bin/cli.js
elsassph: Let me check again, maybe that's incorrect.

src/Error.ts
elsassph: I'll check again, but I've found that errors were somehow all printed on a single line.

test/cli/cli.test.js
elsassph: That's because it is split without removing blank lines so there's an exta entry. I can change that.

test/extensions/RunInScope.test.js
elsassph: OK

bin/cli.js
elsassph: The rethrow is only for non BRS errors, so that should only happen for a real runtime exception.

src/Error.ts
lkipke: Also, we print that string of non-fatal errors☝️ at the beginning of every unit test run. Is there some way we could let the user know that those errors are non-fatal? Maybe we could add `Warning:` before them?

I've heard feedback from users that it's difficult to tell which error is the one that's causing the failure when those non-fatal errors are printed directly above the fatal one.

bin/cli.js
lkipke: I think Sean is saying that this is the `brs` cli, so either way we'll be logging the error to the command line and exiting with error (whether we re-throw, or just use `console.error`). What is the difference in intended output with these changes?

src/Error.ts
elsassph: I appreciate that it's wrong to add the newline when formatting the error. I couldn't narrow why the errors are printed without newlines...

sjbarag: > Also, we print that string of non-fatal errors☝️ at the beginning of every unit test run. Is there some way we could let the user know that those errors are non-fatal? Maybe we could add `Warning:` before them?

Filed #467 to track that!

> I've found that errors were somehow all printed on a single line.

Agreed, I've seen that too!  Looking deeper into it, we probably should move that into `getLoggerUsing` so that the error messages themselves don't need to worry about how they're logged.  If we switch back to `console.error`, we'd wind up with extra newlines if we append `\n` here!

PR #460, opened 2020-07-16
chore(deps): Upgrade to prettier 2.x

sjbarag:
The defaults changed for prettier -- mostly `arrowParens: true` -- andI've been meaning to enable that anyway.  Let's use that major-version bump as an excuse to reformat everything.  It pains me a little to have such a large diff but at least it's mostly `()` insertion.

Realistically, the main motivation for the upgrade is support for TypeScript 3.8 language features like `import type …`.

### DIFF
--- a/src/brsTypes/components/RoArray.ts
+++ b/src/brsTypes/components/RoArray.ts
@@ -180,7 +180,7 @@ export class RoArray extends BrsComponent implements BrsValue, BrsIterable {
 
             this.elements = [
                 ...this.elements,
-                ...array.elements.filter(element => !!element), // don't copy "holes" where no value exists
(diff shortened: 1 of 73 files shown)


src/brsTypes/components/RoArray.ts
```diff
@@ -214,7 +214,7 @@ export class RoArray extends BrsComponent implements BrsValue, BrsIterable {
             if (flags.toString().match(/([^ir])/g) != null) {
                 interpreter.stderr.write("roArray.Sort: Flags contains invalid option(s).\n");
             } else {
-                this.elements = this.elements.sort(function(a, b) {
+                this.elements = this.elements.sort(function (a, b) {
```
sjbarag, line 217: thanks, I hate it

PR #462, opened 2020-07-16
fix(rsg,mock): Allow mocked components as children of created components

sjbarag:
We previously didn't create a mocked component when building out a node's child tree, e.g.:

```xml
<?xml version="1.0" encoding="UTF-8"?>
<!-- components/HasALabel.xml -->

<component name="HasALabel">
    <children>
        <Label id="foo"/>
    </children>
</component>
```

```brightscript
' example.brs

_brs_.mockComponent("Label", {
    id: "foo",
    extra_field: "some extra field"
})
hal = createObject("HasALabel")
print hal.findNode("foo").hasField("extra_field")
'                         ~~~~~~~~~~~~~~~~~~~~~~~
'                         used to be false, is now true!
```

Now, any child nodes of a created node will create a mock component if one is found!  This should enable more robust testing scenarios, as well as allowing consumers to provide their own mocks for components we haven't implemented yet.


src/brsTypes/components/RoSGNode.ts
```diff
@@ -21,6 +21,7 @@ import { ComponentDefinition, ComponentNode } from "../../componentprocessor";
 import { ComponentFactory, BrsComponentName } from "./ComponentFactory";
 import { Environment } from "../../interpreter/Environment";
 import { roInvalid } from "./RoInvalid";
+import type * as MockNodeModule from "../../extensions/MockNode";
```
sjbarag, line 24: New in TypeScript 3.8, which required upgrading to `prettier@2.x` in #461

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1320,6 +1321,13 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
 export const mGlobal = new RoSGNode([]);
 
 export function createNodeByType(interpreter: Interpreter, type: BrsString): RoSGNode | BrsInvalid {
+    // If the requested component has been mocked, construct and return the mock instead
+    let maybeMock = interpreter.environment.getMockObject(type.value.toLowerCase());
+    if (maybeMock instanceof RoAssociativeArray) {
+        let mock: typeof MockNodeModule = require("../../extensions/MockNode");
```
sjbarag, line 1327: Copy-pasted (ish) from https://www.typescriptlang.org/docs/handbook/modules.html#optional-module-loading-and-other-advanced-loading-scenarios

PR #465, opened 2020-07-22
fix(stdlib): remove duplicate extends warning

lkipke:
# Change Summary

This removes a duplicate warning we logged to `stdout` when we encountered an unknown component extension. It's exactly the same scenario, so there's no need to log it twice.

### DIFF
--- a/src/componentprocessor/index.ts
+++ b/src/componentprocessor/index.ts
@@ -156,24 +156,6 @@ async function processXmlTree(
         if (xmlNode) {
             nodeDef.children = getChildren(xmlNode);
             nodeDef.scripts = getScripts(xmlNode);
-            let baseNode = xmlNode.attr.extends;
-            while (baseNode) {
-                let baseNodeDef = nodeDefMap.get(baseNode);
-                if (baseNodeDef) {
-                    nodeDef.children = [...nodeDef.children];
-                    baseNode = baseNodeDef.xmlNode!.attr.extends;
-                } else {
-                    if (!(baseNode in BrsComponentName)) {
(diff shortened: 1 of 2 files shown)


src/componentprocessor/index.ts
```diff
@@ -156,24 +156,6 @@ async function processXmlTree(
         if (xmlNode) {
             nodeDef.children = getChildren(xmlNode);
             nodeDef.scripts = getScripts(xmlNode);
-            let baseNode = xmlNode.attr.extends;
-            while (baseNode) {
-                let baseNodeDef = nodeDefMap.get(baseNode);
-                if (baseNodeDef) {
-                    nodeDef.children = [...nodeDef.children];
```
lkipke, line 163: It turns out this whole `while` loop doesn't actually do anything, because
```
nodeDef.children = [...nodeDef.children]
```
is a noop!

src/componentprocessor/index.ts
```diff
@@ -156,24 +156,6 @@ async function processXmlTree(
...
-                    baseNode = baseNodeDef.xmlNode!.attr.extends;
-                } else {
-                    if (!(baseNode in BrsComponentName)) {
-                        // The reference implementation doesn't allow extensions of unknown node subtypes, but
-                        // BRS hasn't implemented every node type in the reference implementation!  For now,
-                        // let's warn when we detect unknown subtypes.
```
sjbarag, line 169: Would you mind copying this comment over to `ComponentScopeResolver.ts` with our `console.error`?

lkipke: You ask so much of me! 😉

PR #469, opened 2020-07-24
feat(cicd): Set up GitHub Actions for push & pull_request

sjbarag:
GitHub Actions offers higher concurrency limits on free tiers and offers non-linux OSes!  Seems like a win :)

### DIFF
--- a/.github/workflows/pull-request.yaml
+++ b/.github/workflows/pull-request.yaml
@@ -1,0 +1,40 @@
+{
+  name: "PR & Push",
+  on: {
+    push: {
+      branches: [ "master" ]
+    },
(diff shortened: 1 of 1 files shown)


sjbarag:
Eventually we'll drop Circle CI in favor of pure github actions.  For now they can both exist, since I still have to migrate the nightly build job and the "publish when a tag gets pushed" job (though that'll get replaced with a "publish when a release is triggered" job)

sjbarag:
> I have never used github actions before but code makes sense.

Please excuse my use of ~json instead of yaml; most of my time with yaml is spent changing indentation and committing 😓

PR #472, opened 2020-07-24
chore(ci): Disable circleci builds for push & pull_request

sjbarag:
We'll migrate nightly builds and automatic publishes shortly, but these builds are the primary use for Circle and can be replaced with the GitHub Actions integration introduced via #469.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -76,12 +76,6 @@ jobs:
 
 workflows:
   version: 2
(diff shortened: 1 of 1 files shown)


sjbarag:
Branch protection rules updated ahead of this landing:

image

sjbarag:
Confirmation that those circleci builds no longer run:

image

PR #476, opened 2020-07-27
fix(ci): Use actions/setup-node to seed auth token in .npmrc

sjbarag:
While `yarn` doesn't seem to look at an environment variable for authentication, it does look at ~/.npmrc and the setup-node action reads an auth token from the `NODE_AUTH_TOKEN` environment variable:

https://github.com/actions/setup-node/blob/59e61b89511ed136a0b17773f07c349fa5c01e8b/src/authutil.ts#L46-L48

All of which is to say this _should_ work now.
I think.
:fingers_crossed:

### DIFF
--- a/.github/workflows/nightly-build.yaml
+++ b/.github/workflows/nightly-build.yaml
@@ -2,7 +2,7 @@
   name: "Nightly build",
   on: {
     schedule: [
-      { cron: "05 16 * * *" }
+      { cron: "20 17 * * *" }
     ]
   },
   jobs: {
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-build.yaml
```diff
@@ -2,7 +2,7 @@
   name: "Nightly build",
   on: {
     schedule: [
-      { cron: "05 16 * * *" }
+      { cron: "20 17 * * *" }
```
sjbarag, line 5: Should publish a nightly build at 10:20 PT (17:20 UTC).  Assuming everything works.

sjbarag:
Has to be merged before it can run, so here's goes something

PR #477, opened 2020-07-27
feat(perf): Memoize component scope resolution lexing and parsing

sjbarag:
Many components can include the same files via `<script/>` tag, and we previously lexed, preprocessed, and parsed those files every time they were encountered in any component.  Even if a component and its extension both import `uri="pkg:/components/foo.brs"`, we'd parse it twice.  Bummer!

We can save a significant amount of time by memoizing that lexing, preprocessing, and parsing process across components.

### DIFF
--- a/src/parser/ComponentScopeResolver.ts
+++ b/src/parser/ComponentScopeResolver.ts
@@ -1,9 +1,16 @@
-import { ComponentDefinition } from "../componentprocessor";
+import { ComponentDefinition, ComponentScript } from "../componentprocessor";
 import * as Stmt from "./Statement";
 import { BrsComponentName } from "../brsTypes";
 
 export class ComponentScopeResolver {
     private readonly excludedNames: string[] = ["init"];
+
+    /**
+     * A map of file URIs to a promise that resolves to an array of that file's statements.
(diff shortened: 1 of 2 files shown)


sjbarag:
As far as real-world numbers go, I ran a simple "hello world" script from the root of a large production channel (300+ components), which triggered a full lex/parse cycle across all components.  With some timing logs added to `src/index.ts`, here's the results:


#### Before
```
$ /path/to/brs/bin/cli.js foo.brs
manifest parsing: 3.225ms
component definitions: 194.009ms
script path rewriting: 7.750ms
create lexerParserFn: 0.115ms
create interpreter: 11839.352ms
lex & parse explicit files: 0.838ms
pre-interpreter steps: 12048.050ms
hello, world
```

#### After
```
$ /path/to/brs/bin/cli.js foo.brs
manifest parsing: 2.883ms
component definitions: 194.693ms
script path rewriting: 8.228ms
create lexerParserFn: 0.078ms
create interpreter: 1145.850ms
lex & parse explicit files: 4.493ms
pre-interpreter steps: 1358.864ms
hello, world
```

PR #478, opened 2020-07-27
fix(ci): Make nightly builds actually work

sjbarag:
Looks like the issue was mostly me manually setting CI=true via the `.env` block?  :shrug:

### DIFF
--- a/.github/workflows/nightly-build.yaml
+++ b/.github/workflows/nightly-build.yaml
@@ -2,7 +2,7 @@
   name: "Nightly build",
   on: {
     schedule: [
-      { cron: "20 17 * * *" }
+      { cron: "05 7 * * *" }
     ]
   },
   jobs: {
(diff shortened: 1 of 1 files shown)


.github/workflows/nightly-build.yaml
```diff
@@ -2,7 +2,7 @@
   name: "Nightly build",
   on: {
     schedule: [
-      { cron: "20 17 * * *" }
+      { cron: "05 7 * * *" }
```
alimnios72, line 5: Does the `5` need to be prefixed with `0`? Also, in which timezone does this run?

.github/workflows/nightly-build.yaml
```diff
@@ -14,29 +14,18 @@
           name: "Set up node 12",
           uses: "actions/setup-node@v1",
           with: {
-            registry-url: "https://registry.npmjs.org",
+            registry-url: "https://registry.npmjs.com",
```
alimnios72, line 17: Why do we switch from `.org` to `.com`?

.github/workflows/nightly-build.yaml
sjbarag: I suppose I should either do both or neither, huh?  😅 

Also, UTC!

.github/workflows/nightly-build.yaml
sjbarag: Our `.npmrc` specifies `.com` so I went for consistency 🤷

alimnios72: That makes sense

PR #479, opened 2020-07-27
chore(ci): Publish new releases via GitHub Actions

sjbarag:
This is the last bit we use circleci for - may as well consolidate!  And now that we've got per-release publishes working, we may as well remove the CircleCI config entirely.

So long, and thanks for all the builds ❤️


sjbarag:
Build technically failed, but that's because circleci had nothing to do :)

sjbarag:
May as well try this in `master` — I've got a new version to publish anyway.

PR #480, opened 2020-07-29
Add RoDeviceInfo Interface

philmein23:
This PR adds the RoDeviceInfo interface to BRS. Most of the methods just return static data because we aren't actually running this on a platform. However, there are a few methods relating to time zone and locality, which pulls from the user's local environment variables.


src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,687 @@
...
+        return BrsBoolean.False;
+    }
+
+    private getModel = new Callable("getModel", {
+        signature: {
+            args: [],
+            returns: ValueKind.String,
+        },
+        impl: (_interpreter) => {
+            return new BrsString("4280x");
```
lkipke, line 85: Hmmm i'm a little hesitant to return this type of static data for all of these functions. I'd definitely prefer to return empty strings (`BrsString("")`), because otherwise it would require users to _know_ these "seeded" values so they could expect them in unit tests.

Curious to hear what @sjbarag thinks as well.

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,687 @@
...
+        },
+    });
+
+    private getRandomUUID = new Callable("getRandomUUID", {
+        signature: {
+            args: [],
+            returns: ValueKind.String,
+        },
+        impl: (_interpreter) => {
+            return new BrsString("123e4567-e89b-12d3-a456-426655440000");
```
lkipke, line 208: Instead of returning a static `uuid` here, let's generate a random one the same way that Brightscript does. We can use the [uuid npm package](https://www.npmjs.com/package/uuid).

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,687 @@
...
+        },
+    });
+
+    private hasFeature = new Callable("hasFeature", {
+        signature: {
+            args: [new StdlibArgument("feature", ValueKind.String)],
+            returns: ValueKind.Boolean,
+        },
+        impl: (_interpreter) => {
+            return BrsBoolean.True;
```
lkipke, line 229: I would prefer to return `BrsBoolean.False` here (technically the device doesn't have _any_ of these features!).

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,687 @@
...
+        },
+    });
+
+    private timeSinceLastKeyPress = new Callable("timeSinceLastKeyPress", {
+        signature: {
+            args: [],
+            returns: ValueKind.Int32,
+        },
+        impl: (_interpreter) => {
+            return new Int32(360);
```
lkipke, line 271: Maybe let's return 0 here? 0 is a little more predictable/"default" than anything else.

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,687 @@
...
+    });
+
+    private getDrmInfo = new Callable("getDrmInfo", {
+        signature: {
+            args: [],
+            returns: ValueKind.Object,
+        },
+        impl: (_interpreter) => {
+            let result = new Array<AAMember>();
+            result.push({ name: new BrsString("playready"), value: new BrsString("tee;ss") });
```
lkipke, line 282: Let's just return an empty associative array.

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,687 @@
...
+        },
+    });
+
+    private setCaptionsMode = new Callable("setCaptionsMode", {
+        signature: {
+            args: [new StdlibArgument("mode", ValueKind.String)],
+            returns: ValueKind.Boolean,
+        },
+        impl: (_interpreter) => {
+            return BrsBoolean.True;
```
lkipke, line 326: We want `getCaptionsMode` to return this value I believe. So let's initialize a private variable in this class to keep track of captions mode, something like `this.captionsMode`. Also, can you check to see what the behavior is if the user tries to set captions mode to a string that's not one of the 4 supported options? I suspect it's a no-op, but regardless we should emulate it.

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,687 @@
...
+        signature: {
+            args: [],
+            returns: ValueKind.String,
+        },
+        impl: (_interpreter) => {
+            return new BrsString("12h");
+        },
+    });
+
+    private enableAppFocusEvent = new Callable("enableAppFocusEvent", {
```
lkipke, line 350: Similar to the captions mode, let's keep track of this value in a variable.

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,687 @@
...
+        signature: {
+            args: [new StdlibArgument("enable", ValueKind.Boolean)],
+            returns: ValueKind.Dynamic,
+        },
+        impl: (_interpreter) => {
+            return new BrsString("Event notifications are enabled");
+        },
+    });
+
+    private enableScreensaverExitedEvent = new Callable("enableScreensaverExitedEvent", {
```
lkipke, line 360: Variable for this one as well 😄

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,687 @@
...
+        signature: {
+            args: [new StdlibArgument("enable", ValueKind.Boolean)],
+            returns: ValueKind.Dynamic,
+        },
+        impl: (_interpreter) => {
+            return new BrsString("Screen saver exit event notifications are enabled");
```
lkipke, line 366: Does this actually return a string? The BrightScript docs say that it returns a boolean reflecting whether or not the operation was a success.

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,687 @@
...
+        },
+    });
+
+    private enableLowGeneralMemoryEvent = new Callable("enableLowGeneralMemoryEvent", {
+        signature: {
+            args: [new StdlibArgument("enable", ValueKind.Boolean)],
+            returns: ValueKind.Dynamic,
+        },
+        impl: (_interpreter) => {
+            return new BrsString("lowGeneralMemoryLevel event notifications are enabled");
```
lkipke, line 376: Same question as above about this return value 😄

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,687 @@
...
+        },
+    });
+
+    private isStoreDemoMode = new Callable("isStoreDemoMode", {
+        signature: {
+            args: [],
+            returns: ValueKind.Boolean,
+        },
+        impl: (_interpreter) => {
+            return BrsBoolean.True;
```
lkipke, line 396: Let's return false here as the default.

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,687 @@
...
+        signature: {
+            args: [],
+            returns: ValueKind.String,
+        },
+        impl: (_interpreter) => {
+            return new BrsString("WiFi Connection");
+        },
+    });
+
+    private getExternalIp = new Callable("getExternalIp", {
```
lkipke, line 430: Hmm these IP functions are interesting. Technically, we could get the user's IP and return that. I'm not sure whether we would want to do that though. I'd say let's return blank strings/arrays for now. If we decide we want it in the future, we could always add dynamic IP later.

src/brsTypes/components/RoDeviceInfo.ts
philmein23: The way I am understanding this is creating an instance variable `private enableAppFocus = BrsBoolean.False`, create the variable's getter/setter functions. Next, when the consumer of this method passes a boolean argument, true, for example, then inside `impl` we should set the variable (using the setter function) ? Is my line of thought correct or far off? :)

The same thought process also extends to other functions that take arguments as well.

src/global.d.ts
```diff
@@ -1,3 +1,4 @@
 // `nanomatch` unfortunately doesn't provide it doesn't provide its own types, there are none in DefinitelyTyped either
```
alimnios72, line 1: You may want to update the comment to include `uuid` package

test/e2e/resources/components/roDeviceInfo.brs
```diff
@@ -0,0 +1,61 @@
...
+    print deviceInfo.enableCodecCapChangedEvent(true)
+    print deviceInfo.getAudioOutputChannel()
+    print deviceInfo.getAudioDecoderInfo().count()
+    audioDecoderInfo = deviceInfo.canDecodeAudio({ "audio_format": { "codec": "aac"} })
+    print audioDecoderInfo.result
+    print deviceInfo.getSoundEffectsVolume()
+    print deviceInfo.isAudioGuideEnabled()
+    print deviceInfo.enableAudioGuideChangedEvent(true)
+
+end sub
```
alimnios72, line 61: nit, new line at the end of the file

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,759 @@
...
+    });
+
+    private getTimeZone = new Callable("getTimeZone", {
+        signature: {
+            args: [],
+            returns: ValueKind.String,
+        },
+        impl: (_interpreter) => {
+            let timeZone = process.env.TZ;
+            return timeZone ? new BrsString(timeZone) : BrsInvalid.Instance;
```
alimnios72, line 344: I think this should return and empty `BrsString` if there is no TZ because that's the function signature

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,759 @@
...
+        },
+    });
+
+    private hasFeature = new Callable("hasFeature", {
+        signature: {
+            args: [new StdlibArgument("feature", ValueKind.String)],
```
alimnios72, line 350: nit, please add a whitespace between brackets

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,759 @@
...
+    });
+
+    private getCurrentLocale = new Callable("getCurrentLocale", {
+        signature: {
+            args: [],
+            returns: ValueKind.String,
+        },
+        impl: (_interpreter) => {
+            let locale = process.env.LOCALE;
+            return locale ? new BrsString(locale) : BrsInvalid.Instance;
```
alimnios72, line 365: same thing about the `invalid` not being a string

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,759 @@
...
+    });
+
+    private getCountryCode = new Callable("getCountryCode", {
+        signature: {
+            args: [],
+            returns: ValueKind.String,
+        },
+        impl: (_interpreter) => {
+            let countryCode = process.env.LOCALE;
+            return countryCode ? new BrsString(countryCode) : BrsInvalid.Instance;
```
alimnios72, line 376: same thing about the invalid not being a string

src/brsTypes/components/RoDeviceInfo.ts
philmein23: @alimnios72  - It looks like the code gets auto-prettified when I make a commit so it undos the whitespace changes I make to the brackets. :(

alimnios72: oh ok, we'll have to leave it that way :)

src/global.d.ts
```diff
@@ -1,3 +1,6 @@
...
 // such a generic declaration works for now, but should be improved in the future
 declare module "nanomatch";
+
+// added 'uuid' as a module declartion in order for other modules to import it
+declare module "uuid";
```
sjbarag, line 6: Consider taking a devDependency on `@types/uuid` with `yarn add --dev @types/uuid` instead of manually declaring it — then we'll get typings at compile time to ensure we're not using it wrong!

src/brsTypes/components/RoDeviceInfo.ts
sjbarag: You're pretty close, except that we don't need an explicit getter or setter function like you've created — just a simple `this.enableAppFocusStatus = enable; return this.enableAppFocusState;` (or as a oneliner, `return (this.enableAppFocusStatus = enable;`) is enough!  Because those variables are private to this instance of `roDeviceInfo`, no other TypeScript code outside of this class will be able to access them.  Your getters and setters expose that `this.enableAppFocusStatus` property to other TypeScript contexts, but don't appear to be called anywhere!

src/brsTypes/components/RoDeviceInfo.ts
```diff
@@ -0,0 +1,759 @@
...
+        return this.videoFormat;
+    }
+
+    setAudioFormat(audioFormat: ValueKind.Object) {
+        this.audioFormat = audioFormat;
+    }
+
+    getAudioFormat() {
+        return this.audioFormat;
+    }
```
sjbarag, line 213: Since nothing outside of this component is using these getters and setters, consider removing them in favor of reading and writing the private variables `this.audioFormat` directly!  It's less code to maintain and a bit of complexity we can remove.

src/brsTypes/components/RoDeviceInfo.ts
sjbarag: I'll update this before I merge.

PR #487, opened 2020-08-03
fix(stdlib): Create nested children correctly

### LINKED ISSUE #485: Custom Node that extends from Group can't instantiate child nodes?
I have a **MaskGroup** node(mock) that extends from **Group**:
```
<component name="MaskGroup" extends="Group">
    <interface>
        <field id="masksize" type="array" />
        <field id="maskOffset" type="array" />
    </interface>
    <script type="text/brightscript" uri="MaskGroup.brs" />
</component>
```
But when appending it to a another node:

**SomeNode.xml**
```
<component name="SomeNode" extends="Group">
    <script type="text/brightscript" uri="SomeNode.brs" />
    <children>
        <MaskGroup id="maskGroup">
           <Rectangle id="rect1" />
        </MaskGroup>
        <Rectangle id="rect2" />
    </children>
```
</component>

**SomeNode.brs** :
```
sub init()
    m.rect1 = m.top.findNode("rect1")
    m.rect2 = m.top.findNode("rect2")
    ?"rect1"; m.rect1
    ?"rect2:"; m.rect2
end sub
```
It seems to not find rect1. 
_Output from code above:_

```
rect1 invalid
rect2: <Component: roSGNode:Rectangle> =
{
    change: <Component: roAssociativeArray>
    focusable: false
    focusedchild: invalid
    id: rect2
    visible: true
    opacity: 1
    translation: <Component: roArray>
    rotation: 0
    scale: <Component: roArray>
    scalerotatecenter: <Component: roArray>
    childrenderorder: renderLast
    inheritparenttransform: true
    inheritparentopacity: true
    clippingrect: <Component: roArray>
    renderpass: 0
    muteaudioguide: false
    enablerendertracking: false
    rendertracking: disabled
    width: 0
    height: 0
    color: 0xFFFFFFFF
    blendingenabled: true
}

```

Also, if I do `maskGroup.getChildCount()`, it returns `0`.

lkipke:
# Change Summary

Fixes #485.

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -1443,31 +1443,30 @@ function addChildren(
 
     children.forEach((child) => {
         let newChild = createNodeByType(interpreter, new BrsString(child.name));
-        if (newChild instanceof RoSGNode && appendChild) {
-            appendChild.call(interpreter, newChild);
-            let setField = newChild.getMethod("setfield");
-            if (setField) {
-                let nodeFields = newChild.getFields();
-                for (let [key, value] of Object.entries(child.fields)) {
-                    let field = nodeFields.get(key.toLowerCase());
-                    if (field) {
-                        setField.call(
-                            interpreter,
-                            new BrsString(key),
-                            // use the field type to construct the field value
-                            getBrsValueFromFieldType(field.getType(), value)
-                        );
+        if (newChild instanceof RoSGNode) {
+            if (child.children.length > 0) {
+                // we need to add the child's own children
+                addChildren(interpreter, newChild, child);
+            }
+
+            if (appendChild) {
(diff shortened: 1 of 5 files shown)


src/brsTypes/components/RoSGNode.ts
```diff
@@ -1443,31 +1443,30 @@ function addChildren(
...
+                            );
+                        }
                     }
                 }
             }
-        }
 
-        // If there is no type definition, that means it's a built-in component.
-        // In that case, use the child's ComponentNode information to keep recursing.
-        let newTypeDef = interpreter.environment.nodeDefMap.get(child.name) || child;
```
lkipke, line 1467: This gets called in `createNodeByType` anyway, so we don't need to call it here (and in fact calling it here is what caused this bug).

PR #492, opened 2020-08-05
fix(exts,mocks): differentiate between scoped and unscoped function mocks

lkipke:
# Change Summary

We currently conflate module-scoped mock functions (`mockComponentPartial`) with universally scoped mock functions (mocks that override any function of the same name, at every scope level above function-scope). This causes issues when the universal mocks aren't copied from environment to environment.

This PR separates scoped vs. unscoped function mocks.

### DIFF
--- a/test/e2e/BrsComponents.test.js
+++ b/test/e2e/BrsComponents.test.js
@@ -666,21 +666,4 @@ describe("end to end brightscript functions", () => {
             "Node",
         ]);
     });
-
-    test("components/scripts/MockFunctionsMain.brs", async () => {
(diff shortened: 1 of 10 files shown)


test/e2e/BrsComponents.test.js
```diff
@@ -666,21 +666,4 @@ describe("end to end brightscript functions", () => {
             "Node",
         ]);
     });
-
```
lkipke, line 669: This appeared to just be a duplicate of the test in `BrsMock.test.js`

PR #498, opened 2020-08-10
Add coverage report

lkipke:
# Change Summary

This adds the ability for `brs` to collect code coverage statistics while running. Note: this was written pretty quickly and without any testing -- think of this more as like a "first pass" or "what do y'all think of this approach?"

Essentially, the `FileCoverage` class implements `Visitor`, and then stores every statement and expression in the codebase. Then, whenever the interpreter encounters a statement or expression, it logs a hit to the coverage reporter. The coverage reporter finds the corresponding statement/expression in its own records, and logs a hit to the line and the statement or expression.

This hooks into [IstanbulJS](https://istanbul.js.org/) as well, so we can leverage any of their coverage reporters!


src/brsTypes/Callable.ts
```diff
@@ -161,6 +161,8 @@ export class Callable implements Brs.BrsValue {
 
         let mutableArgs = args.slice();
 
+        // interpreter.reportExpressionHit(this.)
```
alimnios72, line 164: do we need to report the expression here too?

src/interpreter/index.ts
```diff
@@ -40,20 +40,23 @@ import { isBoxable, isUnboxable } from "../brsTypes/Boxing";
...
     stderr: NodeJS.WriteStream;
+    generateCoverage: boolean;
 }
 
 /** The default set of execution options.  Includes the `stdout`/`stderr` pair from the process that invoked `brs`. */
 export const defaultExecutionOptions: ExecutionOptions = {
     root: process.cwd(),
     stdout: process.stdout,
     stderr: process.stderr,
+    generateCoverage: true
```
alimnios72, line 59: should we default this to false?

src/interpreter/index.ts
```diff
@@ -113,7 +135,6 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
 
         interpreter.environment.nodeDefMap = componentMap;
 
-        let componentScopeResolver = new ComponentScopeResolver(componentMap, parseFn);
```
alimnios72, line 116: Any reason why this was removed? We'd always have to pass an instance of `componentScopeResolver` with the same `componentMap`, right?

src/parser/ComponentScopeResolver.ts
```diff
@@ -38,7 +47,7 @@ export class ComponentScopeResolver {
      * @param statementMap Statement collections broken up by component.
      * @returns A collection of statements that have been flattened based on hierarchy.
      */
-    private flatten(statementMap: Stmt.Statement[][]): Stmt.Statement[] {
+    public flatten(statementMap: Stmt.Statement[][]): Stmt.Statement[] {
```
alimnios72, line 50: Do you need to change the visibility of this and below methods?

lkipke:
Just FYI, a "production-ready" version of this is coming soon -- it its current state, this PR isn't quite ready for review beyond the overarching design/usage of the `Visitor` pattern 😄

PR #507, opened 2020-08-17
feat: Make AST compatible with ESLint

heftyfunseeker:
ESLint is a popular pluggable linter that semi-relies on a standardized
AST format called ESTree. This change makes the ast brs produces compliant
with ESLint by exposing a few fields the linter needs to walk the tree
and evaluate rules.

* Statements and Expressions now extend a new class called AstNode


src/parser/AstNode.ts
```diff
@@ -0,0 +1,28 @@
+import { Location } from "../lexer";
+
+/**
+ * ABC (that's minimally compliant with ESLint/ESTree) for all nodes within
```
sjbarag, line 4: I assume "ABC" is ~"American Broadcasting Company"~  ~"easy as 1-2-3"~ "abstract base class"?

src/parser/AstNode.ts
```diff
@@ -0,0 +1,28 @@
...
+ * the AST generated by our parser.
+ */
+export abstract class AstNode {
+    _type!: string;
```
sjbarag, line 8: Any particular reason to call this `_type`?  I don't have a problem with it, I'm just curious since it doesn't seem to be required by ESTree, and the `type()` getter seems like it'll be used for everything anyway 😄

test/parser/controlFlow/__snapshots__/While.test.js.snap
```diff
@@ -3,7 +3,9 @@
 exports[`parser while statements while with exit 1`] = `
```
sjbarag, line 3: If I'm honest, I'm scrolling past all of these snapshots on the assumption you ran `yarn test -u` to generate the new versions and didn't hand-edit them.

…I _really_ hope you didn't hand-edit them, because that's a pain!

heftyfunseeker: Correct! `yarn test -u` did the trick. Thanks for the tip!

src/parser/AstNode.ts
heftyfunseeker: It's because I don't know how to Typescript/Javascript :p I'll get rid of the accessor and change this to a readonly member called `type`. I think I was just fumbling over things when I wrote this earlier. Thanks for pointing it out!

heftyfunseeker:
> Changes look good, can you elaborate a little bit on how ESLinter would use the AST? is it like an argument that you can pass to the linter cli?

Yeah, ESLint supports custom parsers. So you can tell ESLint to call your parser, and as long you provide a few of the properties the `Traverser` is looking for on each node, it will walk your tree and build up a queue of nodes to evaluate linter rules for. https://eslint.org/docs/developer-guide/working-with-custom-parsers, https://eslint.org/docs/developer-guide/working-with-rules.

PR #510, opened 2020-08-18
fix(parse): Allow multiple statements in single-line `if` branches

### LINKED ISSUE #481: Single-line if statements don't allow multiple statements in "then" block
The following code:
```brightscript
array = [ "a", "b", "c" ]
index = 2
found = false
for i = 0 to array.Count() - 1
    ?i
    if i = index then found = true : exit for
next
?found
```
has the following outputs on a Roku device:
```
 0
 1
 2
true
```

while on **brs**:
```
0
false
```

-----

Minimum working example:

```brightscript
sub main()
    if false then nop(): print "#481 still repro's": return
    print "#481 fixed"
end sub

sub nop()
end sub
```

sjbarag:
`:` appears to be a valid statement separator for nearly all cases -- including the branches of a single-line `if` statement!  This also solves an off-by-one error in some location tracking logic, likely caused by the fact that editors perfer one-indexing and our AST (in compliance with ESTree) uses zero-indexing.

As a side benefit, this includes a `test:debug` npm script.  I've been using it for a while and keep forgetting to publish it.  It starts `jest` and immediately pauses for a debugger to attach, allowing for interactive debugging without having to remember a bunch of `jest` arguments.  Pretty handy!

fixes #481

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -26,7 +26,9 @@ type BlockTerminator =
     | Lexeme.EndIf
     | Lexeme.EndWhile
     | Lexeme.EndSub
-    | Lexeme.EndFunction;
+    | Lexeme.EndFunction
+    | Lexeme.Newline // possible only in a single-line `if` statement
+    | Lexeme.Eof; // possible only in a single-line `if` statement
 
 /** The set of operators valid for use in assignment statements. */
 const assignmentOperators = [
@@ -261,10 +263,34 @@ export class Parser {
 
         function declaration(...additionalTerminators: BlockTerminator[]): Statement | undefined {
             try {
-                // consume any leading newlines
-                while (match(Lexeme.Newline));
+                let statementSeparators = [Lexeme.Colon];
 
-                if (check(Lexeme.Sub, Lexeme.Function)) {
+                if (additionalTerminators.includes(Lexeme.Newline)) {
+                    // if this declaration can be terminated with a newline, check for one
+                    // before other statement separators
+                    if (check(Lexeme.Newline)) {
+                        return;
+                    }
+                } else {
+                    statementSeparators.push(Lexeme.Newline);
+                }
+
+                while (match(...statementSeparators));
+
+                if (additionalTerminators.includes(Lexeme.Newline)) {
+                    // if this declaration can be terminated with a newline, check for one
+                    // _after_ a series of `:`s.
+                    if (check(Lexeme.Newline)) {
+                        return;
+                    }
+                }
+
+                // if we reached the end, don't attempt to do anything else
+                if (isAtEnd()) {
+                    return;
+                }
+
+                if (functionDeclarationLevel === 0 && check(Lexeme.Sub, Lexeme.Function)) {
                     return functionDeclaration(false);
                 }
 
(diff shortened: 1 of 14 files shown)


src/brsTypes/components/RoSGNode.ts
```diff
@@ -10,7 +10,7 @@ import {
 } from "../BrsType";
 import { RoSGNodeEvent } from "./RoSGNodeEvent";
 import { BrsComponent, BrsIterable } from "./BrsComponent";
-import { BrsType } from "..";
+import { BrsType, isBrsNumber } from "..";
```
sjbarag, line 13: whoops, this snuck in from some other WIP code — brb reverting

src/parser/Parser.ts
```diff
@@ -902,14 +897,14 @@ export class Parser {
                     match(Lexeme.Newline);
                 }
             } else {
-                let thenStatement = declaration(Lexeme.ElseIf, Lexeme.Else);
-                if (!thenStatement) {
+                let maybeThenBranch = block(Lexeme.Newline, Lexeme.Eof, Lexeme.ElseIf, Lexeme.Else);
```
sjbarag, line 900: This is the core of the change.  Even in single-line `if` statements, we need to look for a `block()` instead of a single statement (via `declaration()`).  The remainder of the diff adjusts where statement separators are parsed (moved from the various functions that return statements and consolidated in `block()` and `declaration()`)

test/e2e/resources/conditionals.brs
```diff
@@ -1,123 +1,135 @@
...
+    test_481()
+end sub
+
+' MWE from https://github.com/sjbarag/brs/issues/481
+sub test_481()
+    nop = sub() : end sub
+
+    if false then nop(): print "#481 still repro's": return
+    print "#481 fixed"
+end sub
```
sjbarag, line 135: Oooh an e2e test!

test/parser/controlFlow/If.test.js
```diff
@@ -165,6 +165,20 @@ describe("parser if statements", () => {
             expect(statements).not.toBeNull();
             expect(statements).toMatchSnapshot();
         });
+
+        it("allows multiple statements in 'then' block (#481)", () => {
```
sjbarag, line 169: aaaaand the unit test.  Everything else is a snapshot change caused by the corrected off-by-one in location tracking.

test/parser/controlFlow/__snapshots__/If.test.js.snap
```diff
@@ -64,7 +64,7 @@ Array [
         },
         "file": "",
         "start": Object {
-          "column": 21,
+          "column": 20,
```
alimnios72, line 67: What are these snapshots?

sjbarag: I've been using [jest snapshots](https://jestjs.io/docs/en/snapshot-testing#snapshot-testing-with-jest) for the parser tests (probably since the first few parser tests that existed) to serialize an AST from the parser and compare it to a previously-serialized copy.  This fails a test when the AST changes the same way explicit `expect()`s would, but allows us to quickly update the expected state with `yarn test -u`.  It's still a [change detector test](https://testing.googleblog.com/2015/01/testing-on-toilet-change-detector-tests.html) but it feels like an appropriate use for that 😄

PR #511, opened 2020-08-18
fix(rsg): Cast numbers to proper field types

### LINKED ISSUE #508: Setting a width or height for a poster as an integer is not allowed
I have found that `BRS` works differently from `Roku`, when we try update width or height of `Poster` with an integer it will not be updated.

strattonbrazil:
In issue #508 setting numbers on fields that didn't match the field type were ignored because the check to compare types didn't take numbers into account. 

This updates that check to allow assigning numbers to any number-based field by converting it to the proper type. 

Note, I've updated all number constructors to accept long to make the conversion simpler (so _getValue()_ from a number can be applied to any other number). 

fixes #508

### DIFF
--- a/src/brsTypes/Double.ts
+++ b/src/brsTypes/Double.ts
@@ -20,8 +20,8 @@ export class Double implements Numeric, Comparable, Boxable {
      * @param value the value to store in the BrightScript double, rounded to 64-bit (double)
      *              precision.
      */
-    constructor(value: number) {
-        this.value = value;
+    constructor(value: number | Long) {
+        this.value = Number(value);
     }
 
     /**
--- a/src/brsTypes/Double.ts
+++ b/src/brsTypes/Double.ts
@@ -6,6 +6,7 @@ import { Int64 } from "./Int64";
 import { Float } from "./Float";
 import { roDouble } from "./components/RoDouble";
 import { Boxable } from "./Boxing";
+import Long from "long";
 
 export class Double implements Numeric, Comparable, Boxable {
     readonly kind = ValueKind.Double;
@@ -21,7 +22,7 @@ export class Double implements Numeric, Comparable, Boxable {
      *              precision.
      */
     constructor(value: number | Long) {
-        this.value = Number(value);
+        this.value = value instanceof Long ? value.toNumber() : value;
     }
 
     /**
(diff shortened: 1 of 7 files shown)


src/brsTypes/Double.ts
```diff
@@ -20,8 +20,8 @@ export class Double implements Numeric, Comparable, Boxable {
      * @param value the value to store in the BrightScript double, rounded to 64-bit (double)
      *              precision.
      */
-    constructor(value: number) {
-        this.value = value;
+    constructor(value: number | Long) {
+        this.value = Number(value);
```
sjbarag, line 24: Is calling `Number(value)` safe on an instance of `Long`?  I see that `Long` has a `.toNumber` method, so perhaps something like this would be safer:

```suggestion
        this.value = value instanceof Long ? value.toNumber() : value;
```

src/brsTypes/components/RoSGNode.ts
```diff
@@ -151,6 +154,18 @@ export class Field {
         // Once a field is set, it is no longer hidden.
         this.hidden = false;
 
+        if (isBrsNumber(value)) {
```
lkipke, line 157: 💭 Would it be any cleaner to use `getValueKindFromFieldType` here?
```brightscript
if (isBrsNumber(value) && value.kind !== getValueKindFromFieldType(this.type)) {
  switch (value.kind) {
    case ValueKind.Float:
      value = new Float(value.getValue());
      break;
    case ValueKind.Int32:
...
```

strattonbrazil: The switch adds a lot of break boilerplate IMO, but your suggestion of moving the field check out definitely makes things cleaner. Will change.

src/brsTypes/Double.ts
strattonbrazil: Will change.

src/brsTypes/Float.ts
```diff
@@ -26,7 +27,8 @@ export class Float implements Numeric, Comparable, Boxable {
      * @param value the value to store in the BrightScript float, rounded to 32-bit floating point
      *              precision and maintaining only seven significant digits of accuracy.
      */
-    constructor(value: number) {
+    constructor(value: number | Long) {
+        if (value instanceof Long) value = value.toNumber();
```
sjbarag, line 31: oh wow I'm surprised `prettier` allowed this 😉

strattonbrazil: I'd never do that on purpose. Will fix.

PR #512, opened 2020-08-19
feat(perf): further memoize lexing and parsing

lkipke:
# Change Summary

This pulls the statement memoization into the lexer-parser function instead of the component scope resolver class. After we resolve component scopes, [we then lex and parse](https://github.com/sjbarag/brs/blob/master/src/index.ts#L66) what ends up being a lot of the same files, but without the memoization benefits. Also, this sets us up for the incoming code coverage work, part of which will be lexing and parsing all the same files.

### DIFF
--- a/src/LexerParser.ts
+++ b/src/LexerParser.ts
@@ -19,47 +19,66 @@ export function getLexerParserFn(
     options: Partial<ExecutionOptions>
 ) {
     const executionOptions = Object.assign(defaultExecutionOptions, options);
+    /**
+     * Map file URIs to promises. The promises resolve to an array of that file's statements.
+     * This allows us to only parse each file once.
+     */
+    let memoizedStatements = new Map<string, Promise<Stmt.Statement[]>>();
     return async function parse(filenames: string[]): Promise<Stmt.Statement[]> {
+        async function lexAndParseFile(filename: string) {
+            let contents;
+            try {
+                contents = await readFile(filename, "utf-8");
+            } catch (err) {
+                return Promise.reject({
+                    message: `brs: can't open file '${filename}': [Errno ${err.errno}]`,
+                });
+            }
+
+            let lexer = new Lexer();
+            let preprocessor = new PP.Preprocessor();
+            let parser = new Parser();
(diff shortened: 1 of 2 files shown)


lkipke:
> This looks great. Just curious, did you see any improvement in the parsing time when running tests?

🤔 How do we normally measure that? Is there debug output or a perf command?

alimnios72:
> > This looks great. Just curious, did you see any improvement in the parsing time when running tests?
> 
> 🤔 How do we normally measure that? Is there debug output or a perf command?

Well there isn't a formal process to measure it, I was just thinking of checking if there is a perceptible speed improvement from when you start running the tests to the point you see the first assertion printed out, nothing too fancy 😅

sjbarag:
> > > This looks great. Just curious, did you see any improvement in the parsing time when running tests?
> > 
> > 
> > 🤔 How do we normally measure that? Is there debug output or a perf command?
> 
> Well there isn't a formal process to measure it, I was just thinking of checking if there is a perceptible speed improvement from when you start running the tests to the point you see the first assertion printed out, nothing too fancy 😅

I'm lazy and just run `time npm test` in a repo with `roca` set up, using the before and after versions of `brs` installed via `npm install /path/to/that/repo` 🤷

lkipke:
> > > > This looks great. Just curious, did you see any improvement in the parsing time when running tests?
> > > 
> > > 
> > > 🤔 How do we normally measure that? Is there debug output or a perf command?
> > 
> > 
> > Well there isn't a formal process to measure it, I was just thinking of checking if there is a perceptible speed improvement from when you start running the tests to the point you see the first assertion printed out, nothing too fancy 😅
> 
> I'm lazy and just run `time npm test` in a repo with `roca` set up, using the before and after versions of `brs` installed via `npm install /path/to/that/repo` 🤷

Well, it's not _that_ impressive of a difference, but I'll still take it 😉 
```
Before
8.76s user 0.55s system 131% cpu 7.098 total

After
8.62s user 0.58s system 125% cpu 7.337 total
```

PR #514, opened 2020-08-22
fix(parse): Restore nested block parsing

sjbarag:
I'd accidentally broken nested-block parsing in #510!  This fixes that and ensures that `block()` always consumes the block-closing token.

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -290,25 +290,29 @@ export class Parser {
                     return;
                 }
 
-                if (functionDeclarationLevel === 0 && check(Lexeme.Sub, Lexeme.Function)) {
-                    return functionDeclaration(false);
-                }
+                try {
+                    if (functionDeclarationLevel === 0 && check(Lexeme.Sub, Lexeme.Function)) {
+                        return functionDeclaration(false);
+                    }
 
-                if (checkLibrary()) {
-                    return libraryStatement();
-                }
+                    if (checkLibrary()) {
+                        return libraryStatement();
+                    }
 
-                // BrightScript is like python, in that variables can be declared without a `var`,
-                // `let`, (...) keyword. As such, we must check the token *after* an identifier to figure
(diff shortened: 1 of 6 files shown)


src/parser/Parser.ts
```diff
@@ -832,45 +836,49 @@ export class Parser {
...
-                if (match(Lexeme.Else)) {
-                    elseBranch = block(Lexeme.EndIf);
-                    endIfToken = advance();
+                if (blockEnd.kind === Lexeme.Else) {
+                    let maybeElseBranch = block(Lexeme.EndIf);
+                    if (!maybeElseBranch) {
+                        throw addError(peek(), "Expected 'end if' to terminate 'else' block");
+                    }
+                    elseBranch = maybeElseBranch.body;
+                    endIfToken = maybeElseBranch.closingToken;
```
lkipke, line 881: Ooh nice does this mean the `endIfToken` is populated now? I ran into this when I was writing the coverage tool

src/parser/Parser.ts
```diff
@@ -927,27 +939,28 @@ export class Parser {
...
                     if (!maybeElseBranch) {
                         throw addError(peek(), `Expected a statement to follow 'else'`);
                     }
-                    elseBranch = maybeElseBranch;
+                    elseBranch = maybeElseBranch.body;
                 }
 
                 // move past the first trailing newline to cleanly exit the single-line if
-                match(Lexeme.Newline);
+                // match(Lexeme.Newline);
```
lkipke, line 963: should we just remove?

src/parser/Parser.ts
sjbarag: It does!  That's where a few of the snapshot updates came into play 😃

src/parser/Parser.ts
sjbarag: Whoooooops yes :)

PR #518, opened 2020-08-25
feat(rsg): Add Scene Node 

philmein23:
This PR includes the ability to create a Scene node/component.


src/brsTypes/components/Scene.ts
```diff
@@ -0,0 +1,20 @@
+import { FieldModel } from "./RoSGNode";
+import { Group } from "./Group";
+import { AAMember } from "./RoAssociativeArray";
+
+export class Scene extends Group {
+    readonly defaultFields: FieldModel[] = [
+        { name: "backgroundURI", type: "uri" },
+        { name: "backgroundColor", type: "color", value: "0x000000FF" },
```
sjbarag, line 8: Would you mind adding an e2e test that sets new values for these?  The `uri` and `color` types tend to have special rules involved, and I'm curious how those are handled within `brs` today 😃

Issue #503, opened 2020-08-13
Having a 'return' statement on a sub that is called by an onChange event is raising an error:


adheus:
Given the following node:
**SomeNode.xml**
```
<component name="SomeNode" extends="Group">
    <interface>
        <field id="someText" type="string" onChange="onSomeTextChanged" />
    </interface>
    <script type="text/brightscript" uri="SomeNode.brs" />
</component>
```
**SomeNode.brs**
```
sub init()
    ?"Initing SomeNode..."
    onSomeTextChanged({fakeEvent: ""})
end sub

sub onSomeTextChanged(event=invalid)
    ?"onSomeTextChanged called"
    if event <> invalid
        ?"Event is not invalid"
        return
    end if
end sub
```

When executing the following steps on **main.brs**:
```
node = CreateObject("roSGNode", "SomeNode")
node.someText = "some"
```

We have the following output:
>Initing SomeNode...
onSomeTextChanged called
Event is not invalid
onSomeTextChanged called
Event is not invalid
source/main.brs(3,7-15): `return` encountered

Notice that the first call to `onSomeTextChanged`(outside of the `onChange` scope) does not raise an error but the `onChange` one does.

vbuchii:
I was going to fill out the ticket about it but it is already here :)

sjbarag:
Fixed as part of #522 (which I incorrectly attributed to #508, I'm sorry!) — should be included in our next release 😃

Issue #486, opened 2020-08-03
Reorganize components test directory


lkipke:
# Summary

Right now, our [e2e components directory](https://github.com/sjbarag/brs/tree/master/test/e2e/resources/components) is kind of a mess. We currently keep some scripts inside `scripts`, some out of it, have all sorts of file names, etc. We should consider doing something along the lines of:
```
|-- components
  |-- test-feature-group
    |-- scripts
      Foo.brs
      Bar.brs
    Foo.xml
    Bar.xml
```

Issue #505, opened 2020-08-14
ifSGNodeField#observeFieldScoped not implemented


sjbarag:
Looks like we forgot to implement `observeFieldScoped`, which is available in RBI [as documented here](https://developer.roku.com/docs/references/brightscript/interfaces/ifsgnodefield.md#observefieldscopedfieldname-as-string-functionname-as-string-infofields-as-object-as-boolean).

We'll want to also implement `unobserveFieldScoped`, since those go hand-in-hand 😃

PR #521, opened 2020-08-28
feat(rsg): Require a host node for observeField

sjbarag:
The reference implementation doesn't allow `observeField` (and friends) to be called without a "host node": the SceneGraph node that holds the callback function whos name is passed to `observeField`.  In other words, this is not allowed from the main entry point of a channel:

```brightscript
sub main()
    p = createObject("roSGNode", "Poster") ' works!
    p.observeField("loadStatus", "onLoadStatusChanged")
    ' => BRIGHTSCRIPT: ERROR: roSGNode.ObserveField: no active host node: pkg:/source/main.brs(3)
end sub

sub onLoadStatusChanged(event as object)
    ' does something with new field
end sub
```

The concept of a Host Node appears to be the key to making `unobserveFieldScoped` work differently than `unobserveField`.  While this commit doesn't include support for either of the `unobserveField*` functions, it does make it significantly easier to build them.

supports #505


test/interpreter/Environment.test.js
```diff
@@ -144,6 +145,12 @@ describe("Environment", () => {
         expect(env.getRootM()).toBe(env.createSubEnvironment().getRootM());
     });
 
+    it("maintains host node between sub-nvironments", () => {
```
alimnios72, line 148: typo: `nvironments `

src/brsTypes/components/RoSGNode.ts
```diff
@@ -25,6 +25,7 @@ import { ComponentFactory, BrsComponentName } from "./ComponentFactory";
 import { Environment } from "../../interpreter/Environment";
 import { roInvalid } from "./RoInvalid";
 import type * as MockNodeModule from "../../extensions/MockNode";
+import { BrsError } from "../../Error";
```
alimnios72, line 28: Is this import used?

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1442,10 +1456,12 @@ function addFields(interpreter: Interpreter, node: RoSGNode, typeDef: ComponentD
             }
 
             // Add the onChange callback if it exists.
-            let observeField = node.getMethod("observeField");
```
alimnios72, line 1445: So this block was wrong, wasn't it? We weren't supposed to call the callback when adding fileds

src/interpreter/index.ts
```diff
@@ -1526,10 +1533,16 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
     }
 
     evaluate(this: Interpreter, expression: Expr.Expression): BrsType {
+        try {
```
lkipke, line 1536: Why do we need to surround this with `try-catch`? Isn't `location` required on the interface for `Statement` and `Expression`?

sjbarag:
> @lkipke Also, I'm curious, do you know if this will fix #503 as well? Seems like it's related.

#503 was actually solved by #522 , and I misattributed that PR! 😅

src/interpreter/index.ts
sjbarag: This is admittedly because of my laziness — a lot of unit tests for the `interpreter/` tree use raw AST nodes as input, but attempting to access `location` winds up throwing an error 🤢 

I'll see if there's a way to fix that without blowing up the diff!

sjbarag:
> @lkipke I think this PR closes #486.

I'd say it supports #486 but doesn't completely solve it. I'd love to have a separate subdirectory for each Jest `test()` as a goal!

src/brsTypes/components/RoSGNode.ts
sjbarag: I don't think we were ever actually calling the callback when adding those fields, but we were running through `Callable#call`, which has its own quirks with `m`.  I switched to manually implementing the small amount of `observeField` that we need here, since the logic is pretty simple 😄

src/interpreter/index.ts
sjbarag: Fixed!

src/brsTypes/components/RoSGNode.ts
sjbarag: Nope!  Removed — it probably snuck in while I was working on #522

PR #522, opened 2020-09-01
fix(interp): Allow return statements in observeField callbacks

### LINKED ISSUE #508: Setting a width or height for a poster as an integer is not allowed

sjbarag:
This exposed a circular dependency between `brsTypes/` `interpreter/index.ts`, and `Error.ts`, resolved by extracting `TypeMismatch.ts` to the `interpreter/` tree since it's currently only used at BrightScript runtime.

fixes #508

### DIFF
--- a/test/e2e/resources/components/roSGNode.brs
+++ b/test/e2e/resources/components/roSGNode.brs
@@ -225,6 +225,7 @@ end sub
 
 sub onCB1Called()
     print "callback 1 called"
+    return
 end sub
 
 sub onCB2Called()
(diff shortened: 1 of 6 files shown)


test/e2e/resources/components/roSGNode.brs
```diff
@@ -225,6 +225,7 @@ end sub
 
 sub onCB1Called()
     print "callback 1 called"
+    return
```
sjbarag, line 228: Now with a test 🎉

PR #525, opened 2020-09-01
feat(coverage): add code coverage collection

lkipke:
# Change Summary

This adds the ability for `brs` to collect code coverage statistics while running.

Essentially, the `FileCoverage` class implements `Visitor`, and then stores every statement and expression in the codebase. Then, whenever the interpreter encounters a statement or expression, it logs a hit to the coverage reporter. The coverage reporter finds the corresponding statement/expression in its own records, and logs a hit to the line and the statement or expression.

This exports a POJO that can be consumed easily by `roca` (which will then hook into [IstanbulJS](https://istanbul.js.org/), see https://github.com/hulu/roca/pull/56).


src/parser/Statement.ts
```diff
@@ -126,7 +152,7 @@ export class ExitWhile extends AstNode implements Statement {
 
 export class Function extends AstNode implements Statement {
     constructor(readonly name: Identifier, readonly func: Expr.Function) {
-        super("Function");
+        super("NamedFunction");
```
lkipke, line 155: Needed to change this because `Statement` also has a class called `"Function"`

sjbarag:
@lkipke looks like builds are now failing after I landed #531, which introduced a new `visitDim` function for all `Statement.Visitor`s.  Sorry about that!

test/coverageLib/FileCoverage.test.js
```diff
@@ -0,0 +1,414 @@
...
+            it("1 hit", () => {
+                checkSimpleStatement(new Stmt.Expression(new Expr.Literal(new Int32(1))), 10);
+            });
+
+            it("10 hits", () => {
+                checkSimpleStatement(new Stmt.Expression(new Expr.Literal(new Int32(1))), 10);
+            });
+        });
+
+        it("Assignment", () => {
```
sjbarag, line 119: super nit: consider using `test()` for these since the names of the tests don't form sentences/statements/behaviors/etc.

test/e2e/Coverage.test.js
```diff
@@ -0,0 +1,29 @@
...
+    });
+
+    test("running test", async () => {
+        await execute([resourceFile("components", "scripts", "Coverage.brs")], {
+            generateCoverage: true,
+            ...outputStreams,
+        });
+
+        // expect(allArgs(outputStreams.stdout.write).filter((arg) => arg !== "\n")).toEqual([]);
+        // expect(true).toEqual(true);
```
sjbarag, line 27: I assume you're still working out the assertions here?

src/interpreter/index.ts
```diff
@@ -77,6 +81,16 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
         return this._environment;
     }
 
+    public setCoverageCollector(collector: CoverageCollector) {
+        this.coverageCollector = collector;
+    }
+
+    public reportCoverageHit(statement: Expr.Expression | Stmt.Statement) {
```
sjbarag, line 88: I may be missing something, but I suspect this doesn't strictly need to be `public` right?

sjbarag (changes requested):
Looking great so far!  I think there's a few missing assertions towards the bottom (~the e2e test area) so I'm requesting changes for the time being.  I really like the way it's trending though!

src/parser/Statement.ts
sjbarag: Great catch!

src/coverageLib/FileCoverage.ts
```diff
@@ -0,0 +1,368 @@
+import { Stmt, Expr } from "../parser";
```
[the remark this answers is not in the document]
lkipke: The reason I didn't name this `coverage` is because it's often included in the `.gitignore`. For example, if you run `jest --coverage` in a javascript project such as this one, it will create a `coverage/` directory, which you wouldn't want to commit. However, I could rename it to `coverageCollector` or something maybe?

src/coverageLib/FileCoverage.ts
```diff
@@ -0,0 +1,368 @@
...
+        [key: string]: {
+            hits: number[];
+            locations: Location[];
+            /** Branch type, i.e. "if", "or", "and", etc. */
+            type: string;
+        };
+    };
+}
+
+export class FileCoverage {
```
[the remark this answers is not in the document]
lkipke: Yuuuup definitely should do that

src/parser/Statement.ts
```diff
@@ -38,6 +64,9 @@ export interface Statement {
 
     /** The starting and ending location of the expression. */
     location: Location;
+
+    /** The type of statement. */
+    type: string;
```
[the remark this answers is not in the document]
lkipke: Didn't think about that -- great idea 😄

test/coverageLib/FileCoverage.test.js
```diff
@@ -0,0 +1,414 @@
...
+        text: text,
+        isReserved: brs.lexer.ReservedWords.has((text || "").toLowerCase()),
+        literal: literal,
+        location: generateLocation(locationNum || -1),
+    };
+}
+
+function identifier(value, locationNum = 1) {
+    return token(Lexeme.Identifier, value, locationNum);
+}
```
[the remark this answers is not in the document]
lkipke: Yeah, that's where I got the inspiration! I can combine them.

src/coverageLib/FileCoverage.ts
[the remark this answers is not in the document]
lkipke: This didn't _quite_ work out as cleanly as I hoped, because `AstNode` and `Statement`/`Expression` have no knowledge of each other. So Typescript didn't like it when I tried to cast one to the other. But I was able to do `statement as unknown as AstNode`, if that's better 🤷‍♂️

src/parser/Statement.ts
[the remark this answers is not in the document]
lkipke: This didn't _quite_ work out as cleanly as I hoped, because `AstNode` and `Statement`/`Expression` have no knowledge of each other. So Typescript didn't like it when I tried to cast one to the other. But I was able to do `statement as unknown as AstNode`, if that's better 🤷‍♂️

src/parser/Statement.ts
```diff
@@ -27,6 +28,41 @@ export interface Visitor<T> {
...
+]);
+
+/**
+ * Returns a boolean of whether or not the given object is a Statement.
+ * @param obj object to check
+ */
+export function isStatement(obj: Expr.Expression | Stmt.Statement) {
+    // This is to play nice with Typescript. We know that implementations of Statements
+    // and Expressions also extend AstNode, but Typescript doesn't know that.
+    let ast = (obj as unknown) as AstNode;
```
sjbarag, line 58: 🤔 I'm not a huge fan of the `(x as unknown) as y` pattern.

> We know that implementations of Statements and Expressions also extend AstNode, but Typescript doesn't know that.

Let's tell TypeScript that then!  We can modify `interface Expression` to `interface Expression extends AstNode` and `interface Statement` -> `interface Statement extends AstNode`.  At that point, `obj` is guaranteed to be an instance of `AstNode`, so there's no casting necessary here 😃

sjbarag (approved):
Looks great!  I've got one suggestion for that type coercion we're doing in `isStatement` but otherwise this looks perfect.  Approved on the assumption that you'll do something about it (or let me know if you can't) before merging 😄

src/parser/Statement.ts
lkipke: Great idea! Just made that change 👍

alimnios72 (commented):
This looks great and clean, kudos for all the hard work you did here. The only thing that I didn't see was the new cli option to generate the coverage, is that how it works?

alimnios72 (approved):
Just saw the roca PR, looks like the cli option isn't needed

PR #527, opened 2020-09-02
feat(rsg): Implement observeFieldScoped/unobserveFieldScoped

### LINKED ISSUE #505: ifSGNodeField#observeFieldScoped not implemented

sjbarag:
This implements `observeFieldScoped` and `unobserveFieldScoped` on top of #526.  They're implemented pretty similarly to the previous `observeField` and `observeFieldScoped` now that we have support for a host node within each environment.

closes #505


alimnios72 (approved):
Looks great to me!

PR #532, opened 2020-09-04
feat(rsg): Added setString method of ifString interface

### LINKED ISSUE #515: setString only supports one overload
The [roString](https://developer.roku.com/docs/references/brightscript/components/rostring.md) object has the `setString` method defined in two of its interfaces ([ifString](https://developer.roku.com/docs/references/brightscript/interfaces/ifstring.md#setstringvalue-as-string-as-void), [ifStringOps](https://developer.roku.com/docs/references/brightscript/interfaces/ifstringops.md#setstrings-as-string-len-as-integer-as-void)) and they take 2 different method overloading but `brs` only supports one of them, so the following will fail whereas in a real Roku device won't:

```
st = createObject("roString")
st.setString("myString")
```

```
Provided arguments don't match SetString's signature.
    function SetString(s as String, len as Integer) as Void:
        * SetString requires at least 2 arguments, but received 1.
```

alimnios72:
The `roString` object only supported the `setString` with one overload, I added a second overload so the method is supported on both `ifString` and `ifStringOps` interfaces

Closes #515

### DIFF
--- a/src/brsTypes/components/RoString.ts
+++ b/src/brsTypes/components/RoString.ts
@@ -34,6 +34,7 @@ export class RoString extends BrsComponent implements BrsValue, Comparable, Unbo
                 this.toInt,
                 this.toFloat,
                 this.tokenize,
+                this.setString,
                 this.split,
                 this.getEntityEncode,
                 this.escape,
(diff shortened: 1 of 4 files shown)


sjbarag (approved):
Great catch!!

PR #533, opened 2020-09-05
fix(rsg): Adds optional arg to RoString constructor

### LINKED ISSUE #513: AppendString() not working
The following function generates a 22 digits number on Roku native brightscript:
```
function generateSessionId()
    MAX_SIZE = 22
    index = 0
    sessionId = box("")
    while index <= MAX_SIZE
        index = index + 1
        sessionId.appendString(Rnd(9).toStr(), 1)
    end while
    return sessionId
end function
```

On **brs**, it generates an empty string.

NOTE: The `box()`function is not available but I'm using a mock that just returns the same value passed as argument. 
See [Box(x as Dynamic) as Object](https://developer.roku.com/en-gb/docs/references/brightscript/language/runtime-functions.md#boxx-as-dynamic-as-object)

alimnios72:
I found that when creating an string like this:
```
str = createObject("roString")
```
The intrinsic string doesn't have initial value because we were always expecting a second argument in the `createObject` syntax, then when we tried to use methods like `appendString` to the object they won't work because there is no initial string.
I made optional the argument in the `RoString` class constructor so it doesn't break the string boxing functionality but at the same time makes things consistent with the Roku implementation of `roString` which doesn't take a second argument.

Thanks to @adheus who find the bug

Fixes #513

### DIFF
--- a/src/brsTypes/components/RoString.ts
+++ b/src/brsTypes/components/RoString.ts
@@ -10,16 +10,19 @@ import { Float } from "../Float";
 
 export class RoString extends BrsComponent implements BrsValue, Comparable, Unboxable {
     readonly kind = ValueKind.Object;
-    private intrinsic: BrsString;
+    private intrinsic: BrsString = new BrsString("");
 
     public getValue(): string {
         return this.intrinsic.value;
     }
 
-    constructor(initialValue: BrsString) {
+    constructor(initialValue?: BrsString) {
         super("roString");
 
-        this.intrinsic = initialValue;
+        if (initialValue instanceof BrsString) {
+            this.intrinsic = initialValue;
+        }
+
         this.registerMethods({
             ifString: [this.setString, this.getString],
             ifStringOps: [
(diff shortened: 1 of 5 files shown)


src/brsTypes/components/RoString.ts
```diff
@@ -10,16 +10,19 @@ import { Float } from "../Float";
...
     public getValue(): string {
         return this.intrinsic.value;
     }
 
-    constructor(initialValue: BrsString) {
+    constructor(initialValue?: BrsString) {
         super("roString");
 
-        this.intrinsic = initialValue;
+        if (initialValue instanceof BrsString) {
```
sjbarag, line 22: We probably don't need an `instanceof` check here since we've declared that parameter as `BrsString | undefined` 😃 

```suggestion
        if (initialValue) {
        // or if you want to be explicit:
        // if (initialValue != null) {
```

sjbarag (approved):
Well that seemed easy — thanks for taking care of it!!

src/brsTypes/components/RoString.ts
alimnios72: Good call!

PR #536, opened 2020-09-08
fix(parser): add else token

lkipke:
# Change Summary

This adds support for the parser to record the `else` token location.


sjbarag (approved):
That'll do it!

PR #544, opened 2020-09-24
fix: minor naming collision with eslint/estree spec

heftyfunseeker:
- Rename Function.end to Function.endKeyword

AST nodes should not have a start or end property.
https://github.com/eslint/eslint/pull/8984


sjbarag (approved):
Good catch!  Thanks for sending this over @heftyfunseeker 🙂

alimnios72 (approved):
Sweet deal

PR #547, opened 2020-09-24
fix(error-handling): Use addError instead of throw

lkipke:
# Change Summary

This calls `interpreter.addError` instead of `throw new BrsError` in several spots so that we get these runtime errors printed out to the console when they occur. Also, it removes a `finally` block so that we are explicitly swallowing errors, rather than doing it implicitly.


test/e2e/Errors.test.js
```diff
@@ -0,0 +1,37 @@
...
+    });
+
+    test("components/errors/indexed-get.brs", async () => {
+        await execute([resourceFile("components", "errors", "indexed-get.brs")], outputStreams);
+
+        let errOutput = allArgs(outputStreams.stderr.write).filter((arg) => arg !== "\n");
+        expect(
+            errOutput[0].includes("Attempting to retrieve property from non-iterable value")
+        ).toBeTruthy();
+    });
```
alimnios72, line 36: Can we add a test case for `"Attempting to retrieve property from iterable with illegal index type"` error?

alimnios72 (commented):
Looks good, just one small request

alimnios72 (approved):
Thanks, this looks great

sjbarag (approved):
Thanks for fixing this!🎉

PR #549, opened 2020-09-25
feat(components): configurable directories for component search

### LINKED ISSUE #501: Utility functions don't have access to the components scope
Let's say that we have `SomeComponent.brs` component and `SomeComponent.test.brs` for testing it. If for some reason we'd like to create utility function in `SomeComponent.test.brs` file and need access to `m` of `SomeComponent` it doesn't look possible.

lkipke:
# Change Summary

Allows directories to search for `.xml` files to be configurable. Closes #501.

### DIFF
--- a/src/componentprocessor/index.ts
+++ b/src/componentprocessor/index.ts
@@ -82,7 +82,10 @@ export class ComponentDefinition {
     }
 }
 
-export async function getComponentDefinitionMap(rootDir: string = "", additionalDirs: string[] = []) {
+export async function getComponentDefinitionMap(
+    rootDir: string = "",
+    additionalDirs: string[] = []
+) {
     let searchString = `{components,${additionalDirs.join(",")}}`;
     const componentsPattern = path.join(rootDir, searchString, "**", "*.xml");
     const xmlFiles: string[] = fg.sync(componentsPattern, {});
--- a/src/componentprocessor/index.ts
+++ b/src/componentprocessor/index.ts
@@ -82,9 +82,9 @@ export class ComponentDefinition {
     }
 }
 
-export async function getComponentDefinitionMap(rootDir: string, componentDirs: string[] = []) {
-    let componentDirsString = ["components"].concat(componentDirs).join("|");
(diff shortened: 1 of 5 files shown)


src/componentprocessor/index.ts
```diff
@@ -82,8 +82,12 @@ export class ComponentDefinition {
...
 
-export async function getComponentDefinitionMap(rootDir: string) {
-    const componentsPattern = rootDir + "/components/**/*.xml";
+export async function getComponentDefinitionMap(
+    rootDir: string = "",
+    additionalDirs: string[] = []
+) {
+    let searchString = `{components,${additionalDirs.join(",")}}`;
+    const componentsPattern = path.join(rootDir, searchString, "**", "*.xml");
     const xmlFiles: string[] = fg.sync(componentsPattern, {});
```
sjbarag, line 91: One day we'll move this to an `await promisify(fg)(…)` pattern or something.  One day!

src/componentprocessor/index.ts
```diff
@@ -82,8 +82,12 @@ export class ComponentDefinition {
     }
 }
 
-export async function getComponentDefinitionMap(rootDir: string) {
-    const componentsPattern = rootDir + "/components/**/*.xml";
+export async function getComponentDefinitionMap(
+    rootDir: string = "",
+    additionalDirs: string[] = []
+) {
+    let searchString = `{components,${additionalDirs.join(",")}}`;
```
sjbarag, line 89: I assume the trailing comma doesn't cause issues with `fastglob`?

sjbarag (approved):
This all makes sense! Do you think it'd be worth exposing this via the `brs` commandline as well?  It doesn't get a ton of use but it is handy for debugging things in `roca` sometimes.

heftyfunseeker (approved):
Sweet! 🚀

alimnios72 (approved):
LG

src/componentprocessor/index.ts
lkipke: It does not! In fact, if you _don't_ have a trailing comma with one-item list, then it _does_ cause issues with `fastglob`, strangely enough.

Issue #535, opened 2020-09-08
callFunc() is not supported for functions which interface specified in the parent component


vbuchii:
**BaseNode.xml**
```
<?xml version="1.0" encoding="utf-8" ?>
<component name="BaseNode" extends="Node">
    <interface>
        <function name="myFunction" />
    </interface>
.....
```
**BaseNode.brs**
```
sub init()
end sub

sub myFunction(myField as object)
    'Child nodes must implement this method
end sub
```

**SomeNode.xml**
```
<?xml version="1.0" encoding="utf-8" ?>
<component name="SomeNode" extends="BaseNode">
    <interface>
    </interface>
.....
```
**SomeNode.brs**
```
sub init()
end sub

sub myFunction(myField as object)
   ' some implementation here
end sub
```

**SomeNode.test.brs**
```
....
m.it("myFunction")", sub()
                node = createObject("RoSGNode", "SomeNode")
                node.callFunc("myFunction", {})
....
```

`Warning calling function in SomeNode: no function interface specified for myFunction Warning calling function in SomeNode: no function interface specified for myFunction`

lkipke:
Closing -- this was fixed as part of #551.

PR #551, opened 2020-09-25
fix(stdlib): fix callFunc inheritance and arguments bugs

### LINKED ISSUE #483: callFunc() only supports a single parameter
Calling a `node.callFunc("funcName", param1, param2)` with more than one parameter raises the folllowing error:

```
Provided arguments don't match callfunc's signature.
    function callfunc(functionname as String, [functionargs as Dynamic]) as Dynamic:
        * callfunc accepts at most 2 arguments, but received 3.Runtime error encountered in BRS implementation:  Error: Something terrible happened and we didn't throw a `BlockEnd` instance.
```
This should be supported, accordingly to Roku docs:

> Any context from the caller can be passed via one or more method parameters, which may be of any type (previously, callFunc() only supported a single associative array parameter).

lkipke:
# Change Summary

Closes #483 and #535.


src/brsTypes/components/RoSGNode.ts
```diff
@@ -526,48 +526,58 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+                            } else {
+                                return functionToCall.call(subInterpreter);
+                            }
+                        }, componentDef.environment);
                     }
 
-                    subInterpreter.environment.setM(this.m);
-                    subInterpreter.environment.setRootM(this.m);
+                    // Try the parent node if we can't find an implementation here
+                    componentDef = interpreter.environment.nodeDefMap.get(componentDef.extends);
```
sjbarag, line 571: Did we not merge those functions down into child classes? 🤔

sjbarag (approved):
Makes sense to me!  I'm a little surprised that we didn't flatten all the `<function>` definitions when we flatted the rest of the `<interface>` — do you happen to know if there's a reason we didn't?

Either way, thank you for moving the `callfunc` tests into their own component subtree.  That makes things so much easier to understand ❤️

src/brsTypes/components/RoSGNode.ts
lkipke: > Makes sense to me! I'm a little surprised that we didn't flatten all the <function> definitions when we flatted the rest of the <interface> — do you happen to know if there's a reason we didn't?

Great question. I remember having this conversation before. I think what we landed on was that we are actually _not_ flattening our component definitions. Instead, when we instantiate them in `createNodeByType`, we traverse the inheritance tree. I know @alimnios72 filed #408 a while ago that touched on this.

Actually, if you look at the [component definition flattening code](https://github.com/sjbarag/brs/blob/master/src/componentprocessor/index.ts#L128-L151), we're just ignoring `inheritedFields` completely, which was referenced [in this PR for on-change callbacks](https://github.com/sjbarag/brs/pull/424#discussion_r432173107). I know there was a reason for doing that; I am struggling to remember exactly why. It was something to do with the on-change callbacks not existing in the scope of a child component when the definitions were flattened maybe?

alimnios72: I think we only for `init` methods but never for functions available in the component interface. We should probably do in the future to avoid searching the correct function at runtime

alimnios72 (commented):
Changes look great, thanks for fixing this. Just one request: can you add a test case for a function that gets overriden by the child's function?

PR #555, opened 2020-10-02
Add MiniKeyboard

philmein23:
Implementation for MiniKeyboard.

### DIFF
--- a/src/brsTypes/components/MiniKeyboard.ts
+++ b/src/brsTypes/components/MiniKeyboard.ts
@@ -1,0 +1,23 @@
+import { FieldModel } from "./RoSGNode";
+import { Group } from "./Group";
+import { AAMember } from "./RoAssociativeArray";
+
+export class MiniKeyboard extends Group {
+    readonly defaultFields: FieldModel[] = [
+        { name: "text", type: "string", value: "" },
+        { name: "keyColor", type: "string", value: "0x000000FF" },
+        { name: "focusedKeyColor", type: "string", value: "0x000000FF" },
+        { name: "keyboardBitmapUri", type: "string", value: "" },
+        { name: "focusBitmapUri", type: "string", value: "" },
+        { name: "textEditBox", type: "node" },
+        { name: "showTextEditBox", type: "boolean", value: "true" },
+        { name: "lowerCase", type: "boolean", value: "true" },
+    ];
+
+    constructor(initializedFields: AAMember[] = [], readonly name: string = "MiniKeyboard") {
+        super([], name);
+
+        this.registerDefaultFields(this.defaultFields);
+        this.registerInitializedFields(initializedFields);
+    }
+}
--- a/test/e2e/resources/components/MiniKeyboard.brs
+++ b/test/e2e/resources/components/MiniKeyboard.brs
@@ -1,3 +1,18 @@
 sub main()
+    miniKeyboard = createObject("roSGNode", "MiniKeyboard") 
+    print "miniKeyboard node type:" type(miniKeyboard) ' => Node
+    print "miniKeyboard node subtype:"miniKeyboard.subtype() ' => MiniKeyboard
 
+    miniKeyboard.text = "hello"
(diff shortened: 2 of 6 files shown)


src/brsTypes/components/MiniKeyboard.ts
```diff
@@ -0,0 +1,23 @@
...
+import { AAMember } from "./RoAssociativeArray";
+
+export class MiniKeyboard extends Group {
+    readonly defaultFields: FieldModel[] = [
+        { name: "text", type: "string", value: "" },
+        { name: "keyColor", type: "string", value: "0x000000FF" },
+        { name: "focusedKeyColor", type: "string", value: "0x000000FF" },
+        { name: "keyboardBitmapUri", type: "string", value: "" },
+        { name: "focusBitmapUri", type: "string", value: "" },
+        { name: "textEditBox", type: "node" },
```
lkipke, line 12: Hmmm i'm wondering if we should maybe implement the `TextEditBox` node as well (or we can file an issue to do it later).

test/e2e/resources/components/MiniKeyboard.brs
```diff
@@ -0,0 +1,18 @@
...
+    print "miniKeyboard focusedKeyColor:" miniKeyboard.focusedKeyColor
+
+    miniKeyboard.keyboardBitmapUri = "/images/somebitmap.bmp"
+    miniKeyboard.focusBitmapUri = "/images/somebitmap.bmp"
+    print "miniKeyboard keyBitmapUri:" miniKeyboard.keyboardBitmapUri
+    print "miniKeyboard focusBitmapUri:" miniKeyboard.focusBitmapUri
+
+    print "miniKeyboard showTextEditBox:" miniKeyboard.showTextEditBox
+    print "miniKeyboard lowerCase:" miniKeyboard.lowerCase
+end sub
```
lkipke, line 18: nit: newline

lkipke (approved):
Makes sense! Most of the logic in the keyboard is based on remote key presses anyway, so until we have some sort of simulator for that (if ever) then it doesn't seem like there's much more to do here 👍

src/brsTypes/components/MiniKeyboard.ts
philmein23: For sure. It seems like `TextEditBox` node only applies itself in the context of MiniKeyboard or Keyboard. I don't mind adding it to this PR, if that is okay with you or @sjbarag .

sjbarag: Yep, that's fine with me!

lkipke, line 12: Currently, when we have a `node` type field in a component, we initialize it to `invalid`. In this case, should we add some logic after `registerDefaultFields` to initialize `textEditBox` to an actual valid `TextEditBox` node, since that's the behavior in RBI (I assume)?

lkipke (approved):
Awesome! 👍

philmein23:
@sjbarag - Whenever you have a chance, can you give you this PR a review? Thanks again! 👍

alimnios72 (approved):
LGTM

src/brsTypes/components/TextEditBox.ts
```diff
@@ -0,0 +1,25 @@
...
+
+export class TextEditBox extends Group {
+    readonly defaultFields: FieldModel[] = [
+        { name: "text", type: "string", value: "" },
+        { name: "hintText", type: "string", value: "" },
+        { name: "maxTextLength", type: "integer", value: "15" },
+        { name: "cursorPosition", type: "integer", value: "0" },
+        { name: "clearOnDownKey", type: "boolean", value: "true" },
+        { name: "active", type: "boolean", value: "false" },
+        { name: "textColor", type: "string", value: "0x000000FF" },
```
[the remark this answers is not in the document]
philmein23: Should be the right format according to -https://developer.roku.com/docs/references/scenegraph/widget-nodes/texteditbox.md

I got the wrong default set though, so I need to fix that.

PR #556, opened 2020-10-07
feat(extensions): allow resetting single mock components and functions 

### LINKED ISSUE #517: Add support for reseting a specific component mock/partial mock
There are a few cases where we want to mock components to induce behaviors for testing purposes but there's no way to clear a specific component mock without clearing all other mocked components. I would suggest something similar to:

`_brs_.resetMockComponent("MyMockedNode")` - which would reset either `mockComponent` or `mockComponentPartial` calls

`_brs_.resetMockFunction("myMockedFunctionName")`

lkipke:
# Change Summary

Closes #517 

This adds `_brs_.resetMockFunction` and `_brs_.resetMockComponent` so that you can reset a single mocked function or component. Also, it comes with a couple of bug fixes and improvements in that area:
- `_brs_.resetMockComponents` wasn't actually resetting the "scoped" mocks for any specific environment (which is how `mockComponentPartial` works behind the scenes. Now, it does.
- We weren't plumbing through the subtype for a mocked component, which would cause lookups to fail when we tried to do something like `mockedNode.callFunc("func")`. Now, this works.
- Our component type definition map was case-sensitive, which was causing some headaches for differentiating between a node mock and a built-in object mock (the object map was _not_ case sensitive). Also, you can have a component called `name="CamelCase"` and still run `createObject("rosgnode", "camelcase")` successfully, so there was no need to to store it case-sensitive.
- Re-organized the test directory for our mock extension unit tests **(sorry, it's a lot of file changes! most of the 47 changed files are in the `tests/` directory)**


alimnios72 (approved):
Looks great

PR #559, opened 2020-10-08
fix(stdlib): roSGNode#delete and roSGNode#removeField ignore case

lkipke:
# Change Summary

Fix `roSGNode#delete` and `roSGNode#removeField` so that they ignore case


sjbarag (approved):
Good catch!  Was that only on `roSGNode` and not `roAssociativeArray`?

lkipke:
> Good catch! Was that only on `roSGNode` and not `roAssociativeArray`?

Yep, [roAssociativeArray#delete](https://github.com/sjbarag/brs/blob/master/src/brsTypes/components/RoAssociativeArray.ts#L128) handles this correctly already 👍

PR #564, opened 2020-10-22
fix(mocks, stdlib): mocked component callFunc works now

lkipke:
# Change summary

Yep, you guessed it... more `callFunc` updates! This time, we need to account for mocked components in `callFunc` calls.


alimnios72 (approved):
Sweet

sjbarag (approved):

PR #568, opened 2020-10-27
fix(stdlib): only append flag once to RoRegex

lkipke:
# Change summary

Previously, if you tried to call `RoRegex#matchAll` or `RoRegex:replaceAll` multiple times using the same `RoRegex` instance, a Javascript error would be thrown because the `g` flag appeared multiple times in the Javascript regex constructor. This PR fixes that.


sjbarag (approved):
Great catch!

PR #571, opened 2020-10-28
feat(ext): implement brs#triggerKeyEvent

lkipke:
# Change Summary

This implements a `_brs_.triggerKeyEvent`, which will send key events up the focus chain until it's handled.

### DIFF
--- a/src/extensions/triggerKeyEvent.ts
+++ b/src/extensions/triggerKeyEvent.ts
@@ -3,28 +3,12 @@ import {
     Callable,
     StdlibArgument,
     BrsInvalid,
-    RoAssociativeArray,
     BrsString,
     BrsBoolean,
     RoSGNode,
 } from "../brsTypes";
 import { Interpreter } from "../interpreter";
 import { Stmt } from "../parser";
-
-let ButtonNames: Set<string> = new Set([
-    "back",
-    "up",
-    "down",
-    "left",
-    "right",
-    "OK",
-    "replay",
-    "play",
-    "playonly",
-    "rewind",
-    "fastforward",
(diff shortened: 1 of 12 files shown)


src/extensions/triggerKeyEvent.ts
```diff
@@ -0,0 +1,78 @@
...
+export const triggerKeyEvent = new Callable("triggerKeyEvent", {
+    signature: {
+        args: [
+            new StdlibArgument("key", ValueKind.String),
+            new StdlibArgument("press", ValueKind.Boolean),
+        ],
+        returns: ValueKind.Void,
+    },
+    impl: (interpreter: Interpreter, key: BrsString, press: BrsBoolean) => {
+        // Arbitrary key presses are supported, but they're prefixed by "Lit_"
```
lkipke, line 38: We don't necessarily need to do this -- it might be more clear to leave it up to the person writing the test case. I could go either way 🤷‍♂️

lkipke: [Yep, they are!](https://developer.roku.com/docs/references/scenegraph/component-functions/onkeyevent.md)

> Starting from Roku OS version 8.1, literal key keypress events (such as keyboard letters, and so forth) that are sent to the channels via the mobile app or ECP keydown/keyup commands, now go to the onKeyEvent() handler. Previously, only keys that corresponded to remote keys went to the onKeyEvent handler.
> 
> Such keys are now sent at "Lit_" followed by the actual keypress ("A," "B," "C," etc.).

I think I'll remove this logic, it seems like the behavior would be more clear that way.

PR #576, opened 2020-11-10
fix(manifest): BREAKING-CHANGE Prevent precision loss in parsed manifest

### LINKED ISSUE #279: brs.preprocessor.getManifest() does not allow raw values without type conversions
I was working on adding support for auto naming of component library out files in the brightscript language extension and was moving over to use this to parse the manifest.

Due to the fact that the getManifest() function does not allow you to specify if you want automatic boolean and integer type conversions I was not able to use it. See the following discussion: 
https://github.com/TwitchBronBron/vscode-brightscript-language/pull/181#discussion_r308821477

I envision the function to de expanded to support this like so: 
```
export async function getManifest(rootDir: string, smartTypeConversion: boolean = true): Promise<Manifest> {
  ...
}
```

sjbarag:
Numbers are allowed in RBI's `manifest` file, but `brs` used to lose precision by parsing them as numbers.  This caused some subtle differences when a `manifest` was re-serialized after parsing, in which `1.0` was represented only as `1`.  It's unclear whether or not this caused issues with applications run in RBI using that rewritten `manifest`, but this was originally reported back in July 2019.  Stop parsing `manifest` values as numbers to prevent precision loss.

fixes #279

### DIFF
--- a/src/preprocessor/Manifest.ts
+++ b/src/preprocessor/Manifest.ts
@@ -2,11 +2,14 @@ import * as fs from "fs";
 import * as path from "path";
 import { promisify } from "util";
 
-const access = promisify(fs.access);
 const readFile = promisify(fs.readFile);
 
-/** The set of possible value types in a `manifest` file's `key=value` pair. */
-export type ManifestValue = number | string | boolean;
+/**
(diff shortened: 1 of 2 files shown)


alimnios72 (approved):
Thanks for fixing this

PR #583, opened 2020-11-20
feat(rsg): Support adding new BrsObjects at Runtime

markwpearce:
This is a series of changes that allow other software using BRS as a library to add new BrsComponents or Scenegraph components (PinPad, BusySpinner, etc) at Runtime.

This is useful, for example, if another piece of software wanted to implement video playback or Draw2d functionality, and then it could do so by implementing those components locally and still leverage the BRS project.

While the BRS project is very useful as a parser/lexer/interpreter, it does not do any emulation or real-world display of Scenegraph nodes. If there were another piece of software the wanted to leverage BRS, but do that display piece, then being able to override/extend the built in implementation would be a good thing.

### DIFF
--- a/src/brsTypes/components/BrsObjects.ts
+++ b/src/brsTypes/components/BrsObjects.ts
@@ -16,6 +16,7 @@ import { Float } from "../Float";
 import { Int32 } from "../Int32";
 import { Interpreter } from "../../interpreter";
 import { roInvalid } from "./RoInvalid";
+import { BrsComponent } from "./BrsComponent";
 
 /** Map containing a list of brightscript components that can be created. */
 export const BrsObjects = new Map<string, Function>([
@@ -39,3 +40,26 @@ export const BrsObjects = new Map<string, Function>([
     ["roint", (_: Interpreter, literal: Int32) => new roInt(literal)],
     ["roinvalid", (_: Interpreter) => new roInvalid()],
 ]);
+
+/**
+ * Lets another software using BRS as a library to add/overwrite an implementation of a BrsObject.
+ *
+ * This is useful, for example, if another piece of software wanted to implement video playback or Draw2d functionality.
+ *
+ * In each element of the objectList param, it is pair:
+ * [name of the BrsObject (e.g. "roScreen", etc.), function (interpreter, ...additionalArgs) that returns a new object]
+ *
+ * @example
+ *
+ * AddAdditionalBrsObjects([
+ *   ["roScreen", (_interpreter) => {return new roScreen();}]
+ * ])
+ *
+ * @export
+ * @param {[string, Function][]} objectList - array of pairs: [name, constructor function]
+ */
+export function AddAdditionalBrsObjects(objectList: [string, () => BrsComponent][]): void {
+    objectList.forEach(([name, ctor]) => {
+        BrsObjects.set(name.toLowerCase(), ctor);
+    });
+}
--- a/src/brsTypes/components/BrsObjects.ts
+++ b/src/brsTypes/components/BrsObjects.ts
@@ -51,14 +51,13 @@ export const BrsObjects = new Map<string, Function>([
  *
  * @example
  *
- * AddAdditionalBrsObjects([
+ * extendBrsObjects([
  *   ["roScreen", (_interpreter) => {return new roScreen();}]
  * ])
  *
- * @export
- * @param {[string, Function][]} objectList - array of pairs: [name, constructor function]
+ * @param objectList array of pairs: [name, constructor function]
  */
-export function AddAdditionalBrsObjects(objectList: [string, () => BrsComponent][]): void {
+export function extendBrsObjects(objectList: [string, () => BrsComponent][]): void {
     objectList.forEach(([name, ctor]) => {
         BrsObjects.set(name.toLowerCase(), ctor);
     });
--- a/src/brsTypes/components/ComponentFactory.ts
+++ b/src/brsTypes/components/ComponentFactory.ts
@@ -13,6 +13,7 @@ import {
     Scene,
     MiniKeyboard,
     TextEditBox,
+    BrsComponent,
 } from "..";
 
 export enum BrsComponentName {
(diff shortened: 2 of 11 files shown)


src/brsTypes/index.ts
```diff
@@ -32,6 +32,7 @@ export * from "./Float";
 export * from "./Double";
 export * from "./BrsInterface";
 export * from "./Callable";
+export * from "./components/BrsComponent";
```
markwpearce, line 35: Export this so it can be extended by other software

src/stdlib/CreateObject.ts
```diff
@@ -11,6 +11,20 @@ import { BrsObjects } from "../brsTypes/components/BrsObjects";
 import { Interpreter } from "../interpreter";
 import { MockNode } from "../extensions/MockNode";
 
+const AdditionalBrsObjects = new Map<string, Function>();
```
markwpearce, line 14: Simple function to add in additional objects

src/stdlib/CreateObject.ts
```diff
@@ -39,8 +53,9 @@ export const CreateObject = new Callable("CreateObject", {
         if (possibleMock instanceof RoAssociativeArray) {
             return new MockNode(possibleMock, objToMock);
         }
+        const objNameLookup = objName.value.toLowerCase();
 
-        let ctor = BrsObjects.get(objName.value.toLowerCase());
+        let ctor = AdditionalBrsObjects.get(objNameLookup) || BrsObjects.get(objNameLookup);
```
markwpearce, line 58: prefer the additional objects list so the defaults can be overwritten

src/parser/ComponentScopeResolver.ts
```diff
@@ -68,7 +68,7 @@ export class ComponentScopeResolver {
         let currentComponent: ComponentDefinition | undefined = component;
         while (currentComponent.extends) {
             // If this is a built-in component, then no work is needed and we can return.
-            if (currentComponent.extends in BrsComponentName) {
+            if (ComponentFactory.canResolveComponentType(currentComponent.extends)) {
```
markwpearce, line 71: the component factory can do more than `BrsComponentName` types now

src/brsTypes/components/ComponentFactory.ts
```diff
@@ -34,11 +35,33 @@ export enum BrsComponentName {
...
+
+    /**
+     * Adds additional components types to the factory, so other software can extend brs if necessary.
+     * This would allow other software using this to add other node/component types at runtime
+     * For example, adding custom implementations of the built-in types, or
+     * adding additional types (PinPad, BusySpinner, etc) that aren't here yet
+     *
+     * @static
+     * @param {[componentType: string, ctor: (name: string) => RoSGNode][]} types
+     * @memberof ComponentFactory
```
sjbarag, line 48: Since this is a typescript file, `@memberof` and the `{[componentType: string, ctor: (name: stirng) => RoSGNode][]}` parts of these JSDoc lines aren't necessary — the same information shows up in `public static` and the signature of the function on line 50 😄 

```suggestion
     * @param types - the types to add
```

src/brsTypes/components/BrsObjects.ts
```diff
@@ -39,3 +40,26 @@ export const BrsObjects = new Map<string, Function>([
...
+ * [name of the BrsObject (e.g. "roScreen", etc.), function (interpreter, ...additionalArgs) that returns a new object]
+ *
+ * @example
+ *
+ * AddAdditionalBrsObjects([
+ *   ["roScreen", (_interpreter) => {return new roScreen();}]
+ * ])
+ *
+ * @export
+ * @param {[string, Function][]} objectList - array of pairs: [name, constructor function]
```
sjbarag, line 59: Since this is a typescript file, the `@export` and `{[string, Function][]}` parts of these JSDoc comments aren't necessary — that same information is available on the function's declaration!  Please remove those parts:

```suggestion
 * @param objectList - array of pairs: [name, constructor function]
```

src/brsTypes/components/BrsObjects.ts
```diff
@@ -39,3 +40,26 @@ export const BrsObjects = new Map<string, Function>([
...
+ */
+export function AddAdditionalBrsObjects(objectList: [string, () => BrsComponent][]): void {
```
sjbarag, line 61: Most functions in this project are `camelCase`d, and "additional" is pretty redundant here since we're already "add"ing some :) would you mind changing the name to `addBrsObjects` or `extendBrsObjects` or something similar??

```suggestion
export function addBrsObjects(objectList: [string, () => BrsComponent][]): void {
```

src/brsTypes/components/ComponentFactory.ts
```diff
@@ -72,4 +95,20 @@ export class ComponentFactory {
...
     }
+
+    /**
+     * Checks to see if teh given component type can be resolved by the Factory
+     * That is, if it is a built in type or has been added at run time.
+     *
+     * @static
+     * @param {(BrsComponentName | string)} componentType
+     * @returns {boolean}
+     * @memberof ComponentFactory
```
sjbarag, line 106: Like the others, some of these properties aren't necessary.  At the same time, I'd really appreciate seeing what those parameters and the return value represent!  Consider something like:

```suggestion
     * @param componentType - the component that might be resolvable
     * @returns - `true` if `componentType` can be resolved, otherwise `false`
```

src/brsTypes/components/ComponentFactory.ts
```diff
@@ -72,4 +95,20 @@ export class ComponentFactory {
                 return;
         }
     }
+
+    /**
+     * Checks to see if teh given component type can be resolved by the Factory
```
sjbarag, line 100: Uh oh, a typo!

```suggestion
     * Checks to see if the given component type can be resolved by the Factory
```

test/e2e/CustomComponents.test.js
```diff
@@ -0,0 +1,68 @@
...
+        });
+    }
+}
+
+describe("Extending custom components", () => {
+    let outputStreams;
+    let clock;
+
+    beforeAll(() => {
+        clock = lolex.install({ now: 1547072370937 });
```
sjbarag, line 37: I suspect we don't actually need to mock out the clock for these tests, since none of this appears to depend on the current time. Would you mind removing this?

test/e2e/CustomComponents.test.js
```diff
@@ -0,0 +1,68 @@
...
+        jest.resetAllMocks();
+    });
+
+    afterAll(() => {
+        clock.uninstall();
+        jest.restoreAllMocks();
+    });
+
+    test("components/ExtendCustom.brs", async () => {
+        outputStreams.root = __dirname + "/customResources";
```
sjbarag, line 54: Please use `path.join` so this test doesn't fail on Windows.  Windows support isn't great right now, but I'd prefer to not make it worse 😅 

```suggestion
        outputStreams.root = path.join(__dirname, "customResources");
```

Thanks for putting these resources in a separate directory though — it's extremely helpful!

src/brsTypes/components/BrsObjects.ts
```diff
@@ -39,3 +40,26 @@ export const BrsObjects = new Map<string, Function>([
...
+ * This is useful, for example, if another piece of software wanted to implement video playback or Draw2d functionality.
+ *
+ * In each element of the objectList param, it is pair:
+ * [name of the BrsObject (e.g. "roScreen", etc.), function (interpreter, ...additionalArgs) that returns a new object]
+ *
+ * @example
+ *
+ * AddAdditionalBrsObjects([
+ *   ["roScreen", (_interpreter) => {return new roScreen();}]
+ * ])
```
sjbarag, line 56: Ooooh thanks for the example!

sjbarag (changes requested):
This is an excellent idea, @markwpearce !  There's a little bit of cleanup we'll need to do before this lands, but the your overall approach is exactly what I'd expect.  Great start!

markwpearce:
I updated the PR with fixes from all your comments. Thanks for being so detailed!

markwpearce:
Am I missing something? I think I did all the requested changes...

sjbarag (approved):
This looks perfect, @markwpearce !  I'm so sorry for the awkward periods of silence — I need to start triaging PRs more often 😓

Thanks a ton for the PR.  We really do appreciate it! ❤️

Issue #582, opened 2020-11-19
RoSGNode#isSubtype not implemented


lkipke:
# Summary

`RoSGNode#isSubtype` is not implemented. See [the docs for requirements](https://developer.roku.com/docs/references/brightscript/interfaces/ifsgnodedict.md#issubtypenodetype-as-string-as-boolean).

PR #584, opened 2020-11-23
feat(rsg): Add isSubtype() and parentSubtype() functions to RoSGNode

markwpearce:
Solves https://github.com/sjbarag/brs/issues/582

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -269,7 +269,7 @@ export class Field {
 }
 
 /* Hierarchy of all node Types. Used to discover is a current node is a subtype of another node */
-const subTypeHierarchy = new Map<string, string>();
+const subtypeHierarchy = new Map<string, string>();
 
 /**
  *  Checks the node sub type hierarchy to see if the current node is a sub component of the given node type
@@ -287,7 +287,7 @@ function isSubtypeCheck(currentNodeType: string, checkType: string): boolean {
     if (currentNodeType === checkType) {
         return true;
     }
-    return isSubtypeCheck(subTypeHierarchy.get(currentNodeType)!, checkType);
+    return isSubtypeCheck(subtypeHierarchy.get(currentNodeType)!, checkType);
 }
 
 export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
@@ -584,8 +584,8 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
                     // RoSGNode is referenced as "Node"
                     parentType = "Node";
                 }
-                if (!subTypeHierarchy.has(currentNodeType)) {
-                    subTypeHierarchy.set(currentNodeType, parentType);
+                if (!subtypeHierarchy.has(currentNodeType)) {
+                    subtypeHierarchy.set(currentNodeType, parentType);
                 }
             } else {
                 break;
@@ -1636,7 +1636,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             returns: ValueKind.String | ValueKind.Invalid,
         },
         impl: (interpreter: Interpreter, nodeType: BrsString) => {
-            const parentType = subTypeHierarchy.get(nodeType.value.toLowerCase());
+            const parentType = subtypeHierarchy.get(nodeType.value.toLowerCase());
             if (parentType) {
                 return new BrsString(parentType);
             }
@@ -1728,7 +1728,7 @@ export function createNodeByType(interpreter: Interpreter, type: BrsString): RoS
         typeDefStack.push(typeDef);
         while (typeDef) {
             // Add the current typedef to the subtypeHierarchy
-            subTypeHierarchy.set(typeDef.name!.toLowerCase(), typeDef.extends || "Node");
+            subtypeHierarchy.set(typeDef.name!.toLowerCase(), typeDef.extends || "Node");
 
             typeDef = interpreter.environment.nodeDefMap.get(typeDef.extends?.toLowerCase());
             if (typeDef) typeDefStack.push(typeDef);
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -26,6 +26,7 @@ import { Environment } from "../../interpreter/Environment";
 import { roInvalid } from "./RoInvalid";
 import type * as MockNodeModule from "../../extensions/MockNode";
 import { BlockEnd } from "../../parser/Statement";
+import { lstat } from "fs";
 
 interface BrsCallback {
     interpreter: Interpreter;
@@ -268,11 +269,15 @@ export class Field {
     }
 }
 
+/* Hierarchy of all node Types. Used to discover is a current nods is a subtype of another node */
+const SubTypeHierarchy = new Map<string, string>();
+
 export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
     readonly kind = ValueKind.Object;
     private fields = new Map<string, Field>();
     private children: RoSGNode[] = [];
     private parent: RoSGNode | BrsInvalid = BrsInvalid.Instance;
+
     readonly defaultFields: FieldModel[] = [
         { name: "change", type: "roAssociativeArray" },
         { name: "focusable", type: "boolean" },
@@ -283,6 +288,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
 
     constructor(initializedFields: AAMember[], readonly nodeSubtype: string = "Node") {
         super("Node");
+        this.setExtendsType();
 
         // All nodes start have some built-in fields when created.
         this.registerDefaultFields(this.defaultFields);
@@ -337,7 +343,14 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
                 this.insertchildren,
             ],
             ifSGNodeFocus: [this.hasfocus, this.setfocus, this.isinfocuschain],
-            ifSGNodeDict: [this.findnode, this.issamenode, this.subtype, this.callfunc],
+            ifSGNodeDict: [
+                this.findnode,
+                this.issamenode,
+                this.subtype,
+                this.callfunc,
+                this.issubtype,
+                this.parentsubtype,
+            ],
         });
     }
 
@@ -531,6 +544,31 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
             return true;
         }
         return false;
+    }
+
+    /* used for isSubtype */
+    protected setExtendsType() {
+        let baseClass = this.constructor;
+        let currentNodeType: string, parentType: string;
+        while (baseClass) {
+            currentNodeType = baseClass.name.toLowerCase();
+
+            const parentClass = Object.getPrototypeOf(baseClass);
+
+            if (parentClass && parentClass !== Object && parentClass.name) {
+                baseClass = parentClass;
+                parentType = parentClass.name;
+                if (parentType === "BrsComponent") {
+                    // Only care about RoSgNode and above
+                    break;
+                }
+                if (!SubTypeHierarchy.has(currentNodeType)) {
+                    SubTypeHierarchy.set(currentNodeType, parentType);
+                }
+            } else {
+                break;
+            }
+        }
     }
 
     /**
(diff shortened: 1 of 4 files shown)


src/brsTypes/components/RoSGNode.ts
```diff
@@ -268,11 +268,15 @@ export class Field {
     }
 }
 
+/* Hierarchy of all node Types. Used to discover is a current node is a subtype of another node */
+const SubTypeHierarchy = new Map<string, string>();
```
strattonbrazil, line 272: Nit on casing, should be *subTypeHierarchy*.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -533,6 +545,31 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
     }
 
+    /* used for isSubtype */
+    protected setExtendsType() {
+        let baseClass = this.constructor;
+        let currentNodeType: string, parentType: string;
+        while (baseClass) {
+            currentNodeType = baseClass.name.toLowerCase();
+
+            const parentClass = Object.getPrototypeOf(baseClass);
```
strattonbrazil, line 555: I don't think the method prototypes is the best way to do this since we might for whatever reason wrap these objects. The _createNodeByType_ function has context on the node's inheritance chain, and passes its subtype into the constructor. I suggest also passing in the superclass name as well, so you don't need to do this lookup.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1604,6 +1675,24 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
         });
     }
+
+    /**
+     *  Checks the node sub type hierarchy to see if the current node is a sub compnent of the given node type
+     *
+     * @param {string} checkType
+     * @returns {boolean}
+     */
+    protected isSubtypeCheck(currentNodeType?: string, checkType?: string): boolean {
```
strattonbrazil, line 1685: In what cases are these values optional?

test/brsTypes/components/RoSGNode.test.js
```diff
@@ -2436,6 +2437,51 @@ describe("RoSGNode", () => {
...
+                    }
+                }
+
+                class ChildComponent extends ParentComponent {
+                    constructor(name = "ChildComponent") {
+                        super(name);
+                    }
+                }
+
+                it("acurately checks isSubtype", () => {
```
strattonbrazil, line 2454: *accurately*

strattonbrazil (changes requested):
Thanks for the changes, Mark! I added few suggestions and questions. 

I also believe you need to handle multi-level extension, which I don't believe is handled in isSubtype(). For example, if C extends B, and B extends A, C is a subtype of A. Or with MarkupGrid, it is a subtype of ArrayGrid, Group, Node, and apparently even MarkupGrid itself! 

```
    ?"mg: ";m.searchResults.isSubtype("MarkupGrid") -- surprisingly true
    ?"group: ";m.searchResults.isSubtype("Group") -- true
    ?"node: ";m.searchResults.isSubtype("Node") -- true
    ?"font: ";m.searchResults.isSubtype("Font") -- false
```

If that is handled and I missed it let's just add a test for it.

src/brsTypes/components/RoSGNode.ts
lkipke: I'd say you could also use `interpreter.environment.nodeDefMap` here for non-builtin components. If it's builtin (i.e. `Group`), then the simplest is probably `Object.getPrototypeOf`. If it's custom (i.e. `FooComponent`), then the simplest is probably `interpreter.environment.nodeDefMap.get(subtype).extends`.

[We do a similar recursion in the component processor.](https://github.com/sjbarag/brs/blob/main/src/componentprocessor/index.ts#L123)

markwpearce: @lkipke I'll update to use the nodeDefMap when needed.

src/brsTypes/components/RoSGNode.ts
markwpearce: Mostly set as optional in the case where where SubTypeHierarchy.get("someType") returns undefined. :) 
Not really "optional" as a parameter, but more as a convenience.

markwpearce:
Made some Changes so `subtypeHierarchy` is updated both in builtin types and in components that ar e loaded through XML - see the tests

src/brsTypes/components/RoSGNode.ts
markwpearce: I elected to update the subTypeHierarchy on build/creation, as it should never change based on the current state of the interpreter

test/e2e/BrsComponents.test.js
```diff
@@ -229,6 +229,10 @@ describe("end to end brightscript functions", () => {
             "ExtendedChild init",
             "ExtendedComponent init",
             "ExtendedComponent start",
+            "true",
```
markwpearce, line 232: print m.top.isSubtype("ExtendedComponent")
    print m.top.isSubtype("BaseComponent")
    print m.top.isSubtype("Node")
    print m.top.isSubtype("OtherComponent")

test/e2e/resources/components/scripts/ExtendedComponent.brs
```diff
@@ -1,4 +1,9 @@
 sub init()
     print "ExtendedComponent init"
     start()
+
+    print m.top.isSubtype("ExtendedComponent")
```
lkipke, line 5: Can we also add `m.top.parentSubtype` in here?

lkipke (approved):
Cool, makes sense to me!

test/e2e/resources/components/scripts/ExtendedComponent.brs
markwpearce: sure

src/brsTypes/components/RoSGNode.ts
```diff
@@ -268,11 +268,34 @@ export class Field {
     }
 }
 
+/* Hierarchy of all node Types. Used to discover is a current node is a subtype of another node */
+const subTypeHierarchy = new Map<string, string>();
```
strattonbrazil, line 272: Nitpick on casing, but _sub_ here is a prefix and not a distinct word so it should be _subtypeHierarchy_ just like you used _subtype_ in the comment above or _isSubtypeCheck_ below.

strattonbrazil (approved):
Looks nice.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -268,11 +268,34 @@ export class Field {
...
+    if (!currentNodeType || !checkType) {
+        return false;
+    }
+    checkType = checkType.toLowerCase();
+    currentNodeType = currentNodeType.toLowerCase();
+    if (currentNodeType === checkType) {
+        return true;
+    }
+    return isSubtypeCheck(subtypeHierarchy.get(currentNodeType)!, checkType);
+}
```
sjbarag, line 291: The coercion to a non-null type with `!` feels strange in combination with the checks in line 282 and the `: string` typing for the `currentNodeType` and `checkType` parameters.  We force the type system to think the next value of `currentNodeType` is truthy, then in that next call to `isSubtypeCheck` we check for falsy values immediately. Consider checking checking the result of `subtypeHierarchy.get()` before `isSubtypeCheck` calls itself recursively:

```suggestion
function isSubtypeCheck(currentNodeType: string, checkType: string): boolean {
    checkType = checkType.toLowerCase();
    currentNodeType = currentNodeType.toLowerCase();
    if (currentNodeType === checkType) {
        return true;
    }
    let nextNodeType = subtypeHierarchy.get(currentNodeType);
    if (nextNodeType == null) {
        return false;
    }
    return isSubtypeCheck(nextNodeType, checkType);
}
```

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1550,6 +1610,40 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
+
+    /* Checks whether the subtype of the subject node is a descendant of the subtype nodeType
+     * in the SceneGraph node class hierarchy.
+     *
+     *
+     */
+    private parentsubtype = new Callable("parentsubtype", {
+        signature: {
+            args: [new StdlibArgument("nodeType", ValueKind.String)],
+            returns: ValueKind.String | ValueKind.Invalid,
```
sjbarag, line 1636: Unfortunately the type system in this project isn't as expressive as TypeScript's, so this likely doesn't do what you expect.  If this function can return a string or `invalid`, the output type will have to be `ValuKind.Object`.  The reference docs for `parentSubtype` probably have a typo since they don't mention returning `invalid` at all.

```suggestion
            returns: ValueKind.Object,
```

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1550,6 +1610,40 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         },
     });
 
+    /* Checks whether the subtype of the subject node is a descendant of the subtype nodeType
+     * in the SceneGraph node class hierarchy.
+     *
+     *
```
sjbarag, line 1616: Looks like a few empty lines snuck in! 😃 Please remove these when you get a moment.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1633,6 +1727,9 @@ export function createNodeByType(interpreter: Interpreter, type: BrsString): RoS
         // in the correct order.
         typeDefStack.push(typeDef);
         while (typeDef) {
+            // Add the current typedef to the subtypeHierarchy
+            subtypeHierarchy.set(typeDef.name!.toLowerCase(), typeDef.extends || "Node");
```
sjbarag, line 1731: It looks like the `subtypeHierarchy` map tends to contain lowercase keys and values — should `"Node"` be lowercase here as well?

test/brsTypes/components/RoSGNode.test.js
```diff
@@ -2436,6 +2437,83 @@ describe("RoSGNode", () => {
...
+                });
+
+                it("is case-insensitive", () => {
+                    let childNode = new ChildComponent();
+                    let issubtype = childNode.getMethod("issubtype");
+
+                    expect(issubtype.call(interpreter, new BrsString("Node")).value).toBe(true);
+                    expect(issubtype.call(interpreter, new BrsString("node")).value).toBe(true);
+                    expect(issubtype.call(interpreter, new BrsString("NODE")).value).toBe(true);
+                });
```
sjbarag, line 2484: Thanks for including this test!

test/brsTypes/components/RoSGNode.test.js
```diff
@@ -2436,6 +2437,83 @@ describe("RoSGNode", () => {
...
+                    let result = parentsubtype.call(interpreter, new BrsString("MarkUpGrid"));
+                    expect(result.value).toBe("ArrayGrid");
+                });
+
+                it("returns invalid if it does not exist", () => {
+                    let node = new RoSGNode([]);
+                    let parentsubtype = node.getMethod("parentsubtype");
+
+                    let result = parentsubtype.call(interpreter, new BrsString("UnknownNode"));
+                    expect(result.value).toBe(undefined);
```
sjbarag, line 2514: Rather than comparing `result.value` to `undefined`, consider comparing `result` directly to `BrsInvalid.Instance`.  It's a global singleton value so `.toBe` will work just fine, and will save us from having to change this test if `BrsInvalid` ever needs to store a different value.

```suggestion
                    expect(result).toBe(BrsInvalid.Instance);
```

sjbarag (changes requested):
These are really helpful functions to have implemented.  Thanks for tackling them @markwpearce !  My main concern here is with the type coercion happening in `isSubtypeCheck`, but once that's sorted out I'd be happy to merge this 😃

src/brsTypes/components/RoSGNode.ts
markwpearce: @sjbarag keys are lowercase, but values aren't.

The values aren't lowercase, so we can get the "true" names of the parent types (for display, logging, etc).
However, when checking "isSubType()" we must at that point perform the isLowerCase() on both the input and the value.

Does that make sense?

sjbarag (approved):
Thanks for your updates @markwpearce — this looks fantastic! !  Sorry for the delay on my end, things got pretty busy around here during the winter holidays 😅

PR #586, opened 2020-11-24
fix(interp): Remove unnecessary warnings

lkipke:
# Change summary

This removes unnecessary warnings so that test output isn't polluted with them.


sjbarag (approved):
Thanks for the cleanup! 🧹

PR #587, opened 2020-11-24
fix(stdlib): wait for async translations

lkipke:
# Change summary

Apparently I forgot how async code works. This fixes how we load the translation files, so that we actually _wait_ for them to load before executing.

### DIFF
--- a/src/stdlib/Localization.ts
+++ b/src/stdlib/Localization.ts
@@ -61,12 +61,10 @@ export async function loadTranslationFiles(interpreter: Interpreter, rootDir: st
                     let contents = await readFile(filePath, "utf-8");
                     let xmlStr = contents.toString().replace(/\r?\n|\r/g, "");
                     xmlNode = new XmlDocument(xmlStr);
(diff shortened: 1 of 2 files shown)


src/stdlib/Localization.ts
```diff
@@ -51,22 +52,24 @@ function parseTranslations(xmlNode: XmlDocument, locale: string) {
...
+                    xmlNode = new XmlDocument(xmlStr);
+                } catch (err) {
+                    interpreter.stderr.write(`Error reading translations file ${filePath}: ${err}`);
+                    return;
+                }
 
-            parseTranslations(xmlNode, locale);
-        }
-    });
+                parseTranslations(xmlNode, locale);
```
sjbarag, line 69: We could probably move this `parseTranslations()` call into the `try` branch since `catch` returns early anyway?

sjbarag (approved):
Whooooops.  Thanks for catching that :)

alimnios72 (approved):

PR #590, opened 2020-12-07
feat(scenegraph,lex,parse): Add support for <ComponentLibrary/>

### LINKED ISSUE #589: Add support for <ComponentLibrary> with package-local URIs
RBI's SceneGraph implementation supports a [ComponentLibrary](https://developer.roku.com/docs/references/scenegraph/control-nodes/componentlibrary.md) component, used to load components from a separately-packaged bundle of code.  These can be loaded either via HTTP or via `pkg:/`-style local URIs (basically the same rules as `<Poster/>`) and add all `.xml` components to the list of `createObject()`-able components in the main channel.  Those components _must_ be prefixed by the  namespace of the component library.

Useful Resources:
* A sample component library: https://github.com/rokudev/samples/blob/0b92a931ff5f5959e8d8f9361c9bfda012dbb143/utilities/ComponentLibrary.zip
* A sample channel using that component library: https://github.com/rokudev/samples/blob/0b92a931ff5f5959e8d8f9361c9bfda012dbb143/utilities/ComponentLibraryTestChannel.zip
* RBI documentation: https://developer.roku.com/docs/references/scenegraph/control-nodes/componentlibrary.md

sjbarag:
The `<ComponentLibrary/>` component loads a set of namespaced components from a location declared in its `uri` field.  Some remote resources can be loaded, but they require signing and packaging -- both of which are out of scope for this project.  Load `pkg:/`-local component libraries, which rely only on `.zip` compression for packaging.

fixes #589


sjbarag:
Draft PR because I think I forgot to reject non-`pkg:/` URIs

src/index.ts
```diff
@@ -24,6 +30,10 @@ import * as _parser from "./parser";
 export { _parser as parser };
 import { URL } from "url";
 import * as path from "path";
+import { Return } from "./parser/Statement";
+import pSettle from "p-settle";
+import crypto from "crypto";
```
gincwang, line 35: is this used?

gincwang (approved):
my review doesn't say much, but looking great!! thanks again Sean for jumping on this, can't wait to start using it soon 🎊

src/index.ts
sjbarag: Heck, nope!  I was originally using `"brs" + md5(someComponentLibraryUri)` as the prefix for the temp directory, but I switched towards `sanitize-filename` since that more accurately describes what we want.

src/interpreter/index.ts
```diff
@@ -157,6 +160,14 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
         return interpreter;
     }
 
+    public mergeWith(interpreters: Interpreter[]): void {
```
alimnios72, line 163: Can you add some docs please?

test/e2e/resources/component-library/library/manifest
```diff
@@ -0,0 +1,7 @@
+title=FooLibrary
+subtitle=Tests brs support for ComponentLibrary instances
+major_version=1
+minor_version=0
+build_version=0
+sg_component_libs_provided=Foo
+hidden=1
```
alimnios72, line 7: > The component library's manifest must include the rsg_version flag (rsg_version=1.2) to optimize channel performance.

It would be good to include the flag to make this more realistic

alimnios72 (commented):
This looks great I just have one question: do we need to version the zip package if you're already including the shell script?

src/interpreter/index.ts
sjbarag: D'oh, yes absolutely!

test/e2e/resources/component-library/library/manifest
sjbarag: Can do!  that's easy.

sjbarag:
> This looks great I just have one question: do we need to version the zip package if you're already including the shell script?

I'd rather avoid adding that script to the build pipeline.  It's fastish and pretty simple, but that `.zip` is going to change so rarely that doesn't feel terribly worth it.  I'm happy to add it in anyway if you disagree though 😄

alimnios72 (approved):
> I'd rather avoid adding that script to the build pipeline. It's fastish and pretty simple, but that .zip is going to change so rarely that doesn't feel terribly worth it. I'm happy to add it in anyway if you disagree though 😄

I don't have a strong opinion to delete it so let's keep it. Thanks for the changes btw

PR #593, opened 2020-12-10
feat(interp,exts): implement stack trace

lkipke:
# Summary

This implements an extension `_brs_.getStackTrace`, which, as you might expect, gets the stack trace.

Also, I cleaned up the `README` a little -- reorganized the extensions alphabetically.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,23 +18,21 @@
       - [Cleaning](#cleaning)
       - [All Together](#all-together)
   - [Extensions](#extensions)
-    - [`_brs_.getStackTrace(numEntries = 10, excludePatterns = [])`](#_brs_getstacktracenumentries--10-excludepatterns--)
+    - [`_brs_.process`](#_brs_process)
     - [`_brs_.global`](#_brs_global)
+    - [`_brs_.runInScope`](#_brs_runinscope)
     - [`_brs_.mockComponent`](#_brs_mockcomponent)
-    - [`_brs_.mockComponentPartial`](#_brs_mockcomponentpartial)
     - [`_brs_.mockFunction`](#_brs_mockfunction)
       - [Mock Function API](#mock-function-api)
         - [`mock.calls`](#mockcalls)
         - [`mock.results`](#mockresults)
         - [`mock.clearMock()`](#mockclearmock)
         - [`mock.getMockName()`](#mockgetmockname)
-    - [`_brs_.process`](#_brs_process)
-    - [`_brs_.resetMockComponent`](#_brs_resetmockcomponent)
-    - [`_brs_.resetMockComponents`](#_brs_resetmockcomponents)
-    - [`_brs_.resetMockFunction`](#_brs_resetmockfunction)
-    - [`_brs_.resetMockFunctions`](#_brs_resetmockfunctions)
     - [`_brs_.resetMocks`](#_brs_resetmocks)
-    - [`_brs_.runInScope`](#_brs_runinscope)
+    - [`_brs_.resetMockComponents`](#_brs_resetmockcomponents)
+    - [`_brs_.resetMockComponent`](#_brs_resetmockcomponent)
+    - [`_brs_.resetMockFunctions`](#_brs_resetmockfunctions)
+    - [`_brs_.resetMockFunction`](#_brs_resetmockfunction)
     - [`_brs_.triggerKeyEvent(key as string, press as boolean)`](#_brs_triggerkeyeventkey-as-string-press-as-boolean)
 
 <!-- END doctoc generated TOC please keep comment here to allow auto update -->
@@ -172,7 +170,7 @@ For the most part, `brs` attempts to emulate BrightScript as closely as possible
 
 ### `_brs_.getStackTrace(numEntries = 10, excludePatterns = [])`
 
-Prints out the stack trace. You can configure how many lines to print, and also add filters to exclude files/folders from being included in the trace.
+Prints out the stack trace. You can configure how many lines to print, and also use the `excludePatterns` arg, which is an **array of strings**, to exclude files/folders from being included in the trace. *Note: this function internally uses the [Javascript regex engine](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/RegExp) to match the strings, not [Brightscript PCRE](https://developer.roku.com/en-ca/docs/references/brightscript/components/roregex.md).*
 
 ```brs
 print _brs_.getStackTrace()
@@ -194,6 +192,23 @@ print _brs_.getStackTrace(10, ["lib"])
 ' ]
 ```
 
+### `_brs_.process`
+
+Allows you to access the command line arguments and locale. Locale changes will be reflected in related `RoDeviceInfo` functions and the standard library `Tr` function. Usage:
+
+```brightscript
+print _brs_.process
+' {
+'   argv: [ "some", "arg" ],
+'   getLocale: [Function getLocale]
+'   setLocale: [Function setLocale]
+' }
+
+_brs_.process.setLocale("fr_CA")
+print _brs_.process.getLocale() ' => "fr_CA"
+print createObject("roDeviceInfo").getCurrentLocale() ' => "fr_CA"
+```
+
 ### `_brs_.global`
 
 Allows you to access `m.global` from inside your unit tests. Usage:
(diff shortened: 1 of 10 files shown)


src/extensions/GetStackTrace.ts
```diff
@@ -0,0 +1,66 @@
...
+        returns: ValueKind.Object,
+    },
+    impl: (interpreter: Interpreter, numEntries: Int32, excludePatterns: BrsType) => {
+        let stack = interpreter.stack;
+
+        // Filter out any files that the consumer doesn't want to include
+        if (excludePatterns instanceof RoArray) {
+            excludePatterns.getValue().forEach((pattern) => {
+                if (isBrsString(pattern)) {
+                    let regexFilter = new RegExp(pattern.value);
```
alimnios72, line 43: I think we should mention in the documentation that the Javascript regex engine is used for exclusions. [Brightscript uses PCRE ](https://developer.roku.com/en-ca/docs/references/brightscript/components/roregex.md)which works different in some expressions.

src/interpreter/index.ts
```diff
@@ -1614,14 +1615,24 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
 
     evaluate(this: Interpreter, expression: Expr.Expression): BrsType {
         this.location = expression.location;
+        this.stack.push(this.location);
         this.reportCoverageHit(expression);
-        return expression.accept<BrsType>(this);
+
+        let value = expression.accept<BrsType>(this);
+
+        this.stack.pop();
```
alimnios72, line 1623: Why is the location immediately popped out from the stack? Wouldn't the stack be empty in all cases when calling getStackTrace?

src/extensions/GetStackTrace.ts
lkipke: Hmmm great point. Do you think it'd make more sense to pass in an array of `roRegex` objects instead of strings? Or keep it as-is, but update the docs?

src/interpreter/index.ts
lkipke: Great question! `expression.accept` and `statement.accept` work recursively -- they will evaluate the current expression/statement, which will then require evaluating whatever expressions/statements it contains, so on and so forth. For example:
```brs
sub main()          '  ==> 1. `accept`, which then evaluates the inside of this function (`Block`)
    a = [1]         '  ==> 2. `accept` the `Block`, which then evaluates/executes line-by-line, starting with this one.
    print a[0] + 1  '  ==> 3. `accept` the statement, which then `accept`s (evaluates) `a[0]`, steps out, then executes the entire statement
...
```

At each step, we haven't yet exited the relevant `execute`/`evaluate` call for that `statement`/`expression`. When we're done with a statement/expression and have moved on to the next one, then we exit the `accept` call and pop the location from the stack.

src/extensions/GetStackTrace.ts
alimnios72: Given that our `roRegex` implementation isn't PCRE compatible either 😅 I'd  just update the docs and keep it as it is

src/interpreter/index.ts
alimnios72: Got it!, thanks for the detailed explanation.

alimnios72 (approved):
Other than the update in the documentation this looks good to me

README.md
```diff
@@ -196,14 +205,6 @@ print _brs_.global
...
 ```
 
-### `_brs_.runInScope`
-
-Runs a file (or set of files) **in the current global + module scope** with the provided arguments, returning either the value returned by those files' `main` function or `invalid` if an error occurs.
-
-```brightscript
-_brs_.runInScope("/path/to/myFile.brs", { foo: "bar" })
-```
-
```
sjbarag, line 206: This kind of reordering is fine for now, but next time let's shoot for a separate commit to reorder.  Or even better, a separate PR!  It'll help separate "things I've changed because feature X needs it that way" from "things I changed because they're better this way" 😃

src/extensions/GetStackTrace.ts
sjbarag: > Given that our `roRegex` implementation isn't PCRE compatible either 😅 I'd just update the docs and keep it as it is

Agreed!  As long as the docs mention that this is an array of strings and not an array of `roRegex` objects I'm onboard 😃

src/extensions/GetStackTrace.ts
```diff
@@ -0,0 +1,66 @@
...
+            stack
+                // Filter out any internal stack traces.
+                .filter((location) => !INTERNAL_REGEX_FILTER.test(location.file))
+                // Remove any duplicate entries that appear next to each other in the stack.
+                .filter((location, index, locations) => {
+                    if (index === 0) return true;
+                    let prevLocation = locations[index - 1];
+                    return stringifyLocation(prevLocation) !== stringifyLocation(location);
+                })
+                .map((location) => new BrsString(stringifyLocation(location)))
```
sjbarag, line 61: Watch out!  It looks like this'll stringify every location three times (except for the first and last ones) — once as `stringifyLocation(location)`, again in `new BrsString(stringifyLocation(location))` and once as the "previous location".  If formatting all locations to a string is necessary to remove duplicate entries, consider reordering this so that we stringify everything in one pass then filter the already-converted elements:

```typescript
stack
     // Filter out any internal stack traces.
     .filter((location) => !INTERNAL_REGEX_FILTER.test(location.file))
      .map((location) => new BrsString(stringifyLocation(location)))
     // Remove any duplicate entries that appear next to each other in the stack.
     .filter((location, index, locations) => {
         if (index === 0) return true;
         let prevLocation = locations[index - 1];
         return prevLocation !== location;
      });
```

Alternatively, I suppose you could implement a comparison function on `Location` and call that before converting to strings?  Something like:

```typescript
class Location {
    equalTo(other: Location): boolean {
        return this.file === other.file && this.start.line === other.start.line && this.start.column === other.start.column && …
    }
}
```

But it looks like we're only concerned with the starting locations here anyway, so perhaps that's overkill.

src/interpreter/index.ts
sjbarag, line 1623: Do we need to wrap this in a `try/finally` block?  I can't remember if `expression.accept` / `statement.execute` ever throw errors, but if they do it'll be really helpful to ensure the expression/statement that threw is removed from the stack.

sjbarag: > > Great question! `expression.accept` and `statement.accept` work recursively -- they will evaluate the current expression/statement, which will then require evaluating whatever expressions/statements it contains, so on and so forth. For example:
> > 
> > ```brightscript
> > sub main()          '  ==> 1. `accept`, which then evaluates the inside of this function (`Block`)
> >     a = [1]         '  ==> 2. `accept` the `Block`, which then evaluates/executes line-by-line, starting with this one.
> >     print a[0] + 1  '  ==> 3. `accept` the statement, which then `accept`s (evaluates) `a[0]`, steps out, then executes the entire statement
> > ...
> > ```
> > 
> > At each step, we haven't yet exited the relevant `execute`/`evaluate` call for that `statement`/`expression`. When we're done with a statement/expression and have moved on to the next one, then we exit the `accept` call and pop the location from the stack.
> 
> Got it!, thanks for the detailed explanation. 

Excellent description!

sjbarag (approved):
Great work @lkipke !  This is a very clean implementation, and is well documented as well.  Now that component libraries landed (#590), would you mind checking `_brs_.getStackTrace()` from inside one of those?  Everything should be fine, but it seems worth checking before this merges 😃

src/extensions/GetStackTrace.ts
lkipke: Great callout! Yeah, I considered implementing a comparison function on `Location` as well -- that might be the cleanest way to do it.

README.md
lkipke: 🙈 yep, great call -- i might actually just do that now!

lkipke:
> Great work @lkipke ! This is a very clean implementation, and is well documented as well. Now that component libraries landed (#590), would you mind checking `_brs_.getStackTrace()` from inside one of those? Everything should be fine, but it seems worth checking before this merges 😃

Yep, it all looks good!
```
      Array [
        "[Foo:Bar::init]",
    +   "<Component: roArray> =
    + [
    +     /Users/levi.kipke/Projects/brs/test/e2e/resources/component-library/source/main.brs:2:4
    +     /var/folders/sr/j_zsg28s4tqgc3g0kw74hs1rbz38fw/T/brs-pkgcomponentslibrary.zip2bz65m/components/Bar.brs:3:4
    +     /var/folders/sr/j_zsg28s4tqgc3g0kw74hs1rbz38fw/T/brs-pkgcomponentslibrary.zip2bz65m/components/Bar.brs:3:10
    +     /var/folders/sr/j_zsg28s4tqgc3g0kw74hs1rbz38fw/T/brs-pkgcomponentslibrary.zip2bz65m/source/util.brs:2:4
    +     /var/folders/sr/j_zsg28s4tqgc3g0kw74hs1rbz38fw/T/brs-pkgcomponentslibrary.zip2bz65m/source/util.brs:2:10
    + ]",
        "[Foo:util::someUtil]",
```

src/interpreter/index.ts
lkipke: Great catch!

PR #596, opened 2020-12-17
Reverse stack trace

lkipke:
# Change summary

It probably makes more sense for the stack trace to be formatted so that the 0th element is the most recent. It allows for easier traversal for consumers. This also lines up with [Javascript error stack](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/stack), for example.


sjbarag (approved):
~Just rename the function to `_brs_.getStackQueue()` or `_brs_.getEcartKcats()`~

Whoops!  Thanks for the quick fix 😃

PR #597, opened 2020-12-17
fix(interp): Don't modify defaultExecutionOptions after creation

sjbarag:
Using `Object.assign()` without an initial `{}` overwrote values in `defaultExecutionOptions` in a "last overwrite wins" pattern.  This caused the produced interpreter to execute functions like `ListFiles()` relative to the temporary directory the final component library is unzipped into, rather than relative to the process's current working directory.

Use the object spread syntax instead of `Object.assign()` since it's harder to get wrong, and freeze the `defaultExecutionOptions` variable after creation to prevent accidental modification.


lkipke (approved):
🚢

PR #598, opened 2020-12-23
fix(stdlib): fix uninitialized type casing

### LINKED ISSUE #482: Calling type() on a roSGNode instance returns a different type name
The following code has a different output when compared to the one of a Roku device:

```
node =  CreateObject("roSGNode", "SomeNode")
?type(node)
```

**Roku:**
```
roSGNode
```

**brs:**
```
Node
```

Puritanic:
This MR fixes the casing of the `<uninitialized>` brightscript type, and is related to this issue #482 

Apparently, when we try to compare the type of some value to `"<UNINITIALIZED>"` in Brightscript running on Roku devices, it's causing the channel executing that code to crash, as it seems that this should be in all lowercase, which is working as expected.

The problem I'm having with this is that we have some tests which are running in `brs` and in those tests we're testing a util function which is checking if the value is valid and defined:

```vb
function isValidVal(val) as boolean
	if ((type(val) = "<uninitialized>") or (val = invalid) or (type(val) = "roInvalid"))
		return false
	end if
	return true
end function
```

If I test this function in brs like this:

```vb
result = isValidVal(not_defined_anywhere)
```

This returns `true`, which is not correct, but If I change the above function to check for `"<UNINITIALIZED>"` instead it returns the correct `result` which is `false`. The problem is that `"<UNINITIALIZED>"` is not the correct type and I can't refactor the function above to check for it as it's gonna cause a crash for my app.

That being said, I'm not sure if there is a particular reason why is this named `"<UNINITIALIZED>"`, first time contributing to this project.

fixes #482

### DIFF
--- a/src/brsTypes/BrsType.ts
+++ b/src/brsTypes/BrsType.ts
@@ -398,6 +398,6 @@ export class Uninitialized implements BrsValue, Comparable {
     }
 
     toString(parent?: BrsType) {
-        return "<uninitialized>";
+        return "<UNINITIALIZED>";
     }
 }
--- a/src/brsTypes/BrsType.ts
+++ b/src/brsTypes/BrsType.ts
@@ -55,7 +55,7 @@ export namespace ValueKind {
             case ValueKind.Void:
                 return "Void";
             case ValueKind.Uninitialized:
-                return "<UNINITIALIZED>";
+                return "<uninitialized>";
             case ValueKind.Object:
                 return "Object";
         }
@@ -389,7 +389,7 @@ export class Uninitialized implements BrsValue, Comparable {
 
     equalTo(other: BrsType): BrsBoolean {
         if (other.kind === ValueKind.String) {
-            // Allow variables to be compared to the string "<UNINITIALIZED>" to test if they've
+            // Allow variables to be compared to the string "<uninitialized>" to test if they've
             // been initialized
             return BrsBoolean.from(other.value === this.toString());
         }
@@ -398,6 +398,6 @@ export class Uninitialized implements BrsValue, Comparable {
     }
 
     toString(parent?: BrsType) {
-        return "<UNINITIALIZED>";
+        return "<uninitialized>";
     }
 }
--- a/src/brsTypes/BrsType.ts
+++ b/src/brsTypes/BrsType.ts
@@ -391,7 +391,7 @@ export class Uninitialized implements BrsValue, Comparable {
         if (other.kind === ValueKind.String) {
             // Allow variables to be compared to the string "<uninitialized>" to test if they've
             // been initialized
-            return BrsBoolean.from(other.value === this.toString());
+            return BrsBoolean.from(other.value === this.toString().toLowerCase());
         }
 
         return BrsBoolean.False;
--- a/README.md
+++ b/README.md
@@ -17,6 +17,7 @@
       - [Testing](#testing)
       - [Cleaning](#cleaning)
       - [All Together](#all-together)
+  - [Gaps](#gaps)
   - [Extensions](#extensions)
     - [`_brs_.getStackTrace(numEntries = 10, excludePatterns = [])`](#_brs_getstacktracenumentries--10-excludepatterns--)
     - [`_brs_.global`](#_brs_global)
@@ -86,7 +87,7 @@ Dennis Ritchie said "Hello, World!"
 The [Roku](https://roku.com) series of media streaming devices are wildly popular amongst consumers, and several [very](https://netflix.com) [popular](https://hulu.com) [streaming](https://amazon.com/primevideo) [services](https://crackle.com) offer Channels for the Roku platform.  Unfortunately, Roku chanels *must* be written in a language called BrightScript, which is only executable directly on a Roku device.  BRS hopes to change that by allowing Roku developers to test their code on their own machines, thus improving the quality of their channels and the end-user's experience as a whole.
 
 ## So can I use this to watch TV without a Roku?
-Nope!  The BRS project currently has no intention of emulating the Roku user interface, integrating with the Roku store, or emulating content playback.  In addition to likely getting this project in legal trouble, that sort of emulation is a ton of work.  BRS isn't mature enough to be able to sustain that yet.
+Nope!  The BRS project currently has no intention of emulating the Roku user interface, integrating with the Roku store, or emulating content playback.  In addition to likely getting this project in legal trouble, that sort of emulation is a ton of work.
 
 ## Building from source
 The BRS project follows pretty standard `node` development patterns, with the caveat that it uses `yarn` for dependency management.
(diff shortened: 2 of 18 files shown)


src/brsTypes/BrsType.ts
```diff
@@ -389,7 +389,7 @@ export class Uninitialized implements BrsValue, Comparable {
 
     equalTo(other: BrsType): BrsBoolean {
         if (other.kind === ValueKind.String) {
-            // Allow variables to be compared to the string "<UNINITIALIZED>" to test if they've
+            // Allow variables to be compared to the string "<uninitialized>" to test if they've
             // been initialized
             return BrsBoolean.from(other.value === this.toString());
```
sjbarag, line 394: It looks like `print someVariableThatDoesntExist` will print capitalized `<UNINITIALIZED>`, but `print type(someVariableThatDoesntExist)` will print lowercase `<uninitialized>`.  Similarly, `print doesNotExist = "<UNINITIALIZED>"` is `false`, but `print doesNotExist = "<uninitialized>"` is `true`.  We'll probably need to make this compare against `.toString().toLowerCase()` and revert the `toString()` change below!

```suggestion
            return BrsBoolean.from(other.value === this.toString().toLowerCase());
```

src/brsTypes/BrsType.ts
```diff
@@ -398,6 +398,6 @@ export class Uninitialized implements BrsValue, Comparable {
     }
 
     toString(parent?: BrsType) {
-        return "<UNINITIALIZED>";
+        return "<uninitialized>";
```
sjbarag, line 401: Printing an uninitialized variable does print the capitalized version in the reference BrightScript implementation from what I can tell — would you mind reverting this portion?

```suggestion
        return "<UNINITIALIZED>";
```

sjbarag (changes requested):
Hey @Puritanic — thanks for the PR!  You're right, something's definitely not correct in `brs`.  Looks like there's a little more work to align with the `print variableThatDoesNotExist` output in the reference implementation, but this is a fantastic start.  Thanks for all your effort so far!

Puritanic:
> Hey @Puritanic — thanks for the PR! You're right, something's definitely not correct in `brs`. Looks like there's a little more work to align with the `print variableThatDoesNotExist` output in the reference implementation, but this is a fantastic start. Thanks for all your effort so far!

I'll try to fix this today (if I have time) 👍

Puritanic:
@sjbarag I've fixed tests and refactored code based on suggestions you've provided, I hope that everything is okay now.

FYI, I'm getting some weird issues when running tests locally:

```js
Summary of all failing tests
 FAIL  test/brsTypes/components/RoDateTime.test.js
  ● RoDateTime › methods › getLastDayOfMonth › returns the date/time value's last day of the month

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

      Int32 {
        "kind": 4,
    -   "value": 31,
    +   "value": 30,
      }

      210 |                 let result = getLastDayOfMonth.call(interpreter);
      211 |                 expect(getLastDayOfMonth).toBeTruthy();
    > 212 |                 expect(result).toEqual(new Int32(31));
          |                                ^
      213 |             });
      214 |         });
      215 |

      at Object.toEqual (test/brsTypes/components/RoDateTime.test.js:212:32)
      at processTicksAndRejections (node:internal/process/task_queues:93:5)

 FAIL  test/e2e/BrsComponents.test.js (5.228 s)
  ● end to end brightscript functions › components/roDateTime.brs

    expect(received).toEqual(expected) // deep equality

    - Expected  - 1
    + Received  + 1

    @@ -20,11 +20,11 @@
        "Minutes: ",
        "14",
        "Seconds: ",
        "15",
        "Last Day of Month: ",
    -   "30",
    +   "29",
        "Milliseconds: ",
        "160",
        "ISO String UTC: ",
        "2010-11-12T13:14:15Z",
      ]

      81 |         await execute([resourceFile("components", "roDateTime.brs")], outputStreams);
      82 |
    > 83 |         expect(allArgs(outputStreams.stdout.write).filter((arg) => arg !== "\n")).toEqual([
         |                                                                                   ^
      84 |             "Full Date: ",
      85 |             "Friday November 12, 2010",
      86 |             "No Week Day: ",

      at Object.toEqual (test/e2e/BrsComponents.test.js:83:83)
```

These two tests are failing when I run the `yarn test` locally, but seems to be passing in CI 🤔

sjbarag:
> @sjbarag I've fixed tests and refactored code based on suggestions you've provided, I hope that everything is okay now.
> 
> FYI, I'm getting some weird issues when running tests locally:
> 
> ```js
> Summary of all failing tests
>  FAIL  test/brsTypes/components/RoDateTime.test.js
>   ● RoDateTime › methods › getLastDayOfMonth › returns the date/time value's last day of the month
> 
>     expect(received).toEqual(expected) // deep equality
> 
>     - Expected  - 1
>     + Received  + 1
> 
>       Int32 {
>         "kind": 4,
>     -   "value": 31,
>     +   "value": 30,
>       }
> 
>       210 |                 let result = getLastDayOfMonth.call(interpreter);
>       211 |                 expect(getLastDayOfMonth).toBeTruthy();
>     > 212 |                 expect(result).toEqual(new Int32(31));
>           |                                ^
>       213 |             });
>       214 |         });
>       215 |
> 
>       at Object.toEqual (test/brsTypes/components/RoDateTime.test.js:212:32)
>       at processTicksAndRejections (node:internal/process/task_queues:93:5)
> 
>  FAIL  test/e2e/BrsComponents.test.js (5.228 s)
>   ● end to end brightscript functions › components/roDateTime.brs
> 
>     expect(received).toEqual(expected) // deep equality
> 
>     - Expected  - 1
>     + Received  + 1
> 
>     @@ -20,11 +20,11 @@
>         "Minutes: ",
>         "14",
>         "Seconds: ",
>         "15",
>         "Last Day of Month: ",
>     -   "30",
>     +   "29",
>         "Milliseconds: ",
>         "160",
>         "ISO String UTC: ",
>         "2010-11-12T13:14:15Z",
>       ]
> 
>       81 |         await execute([resourceFile("components", "roDateTime.brs")], outputStreams);
>       82 |
>     > 83 |         expect(allArgs(outputStreams.stdout.write).filter((arg) => arg !== "\n")).toEqual([
>          |                                                                                   ^
>       84 |             "Full Date: ",
>       85 |             "Friday November 12, 2010",
>       86 |             "No Week Day: ",
> 
>       at Object.toEqual (test/e2e/BrsComponents.test.js:83:83)
> ```
> 
> These two tests are failing when I run the `yarn test` locally, but seems to be passing in CI 🤔

Those are both date/time-based issues!  Github reports your location as Serbia, so I wouldn't be surprised if our tests are a little flaky — most contributors are in UTC-7 / UTC-4, and the GitHub Actions runners probably use UTC as their timezone.

In fact, I'm able to reproduce that failure by running tests with` TZ=Asia/Krasnoyarsk yarn test`!  You may want to run those tests locally with `TZ=UTC yarn test`, but I'll file an issue to figure out why those are failing with `TZ=Asia/Krasnoyarsk`.  Sorry about that!

sjbarag (approved):
Thanks for the updates, @Puritanic!  I really appreciate the effort you've put in 😄

PR #600, opened 2020-12-23
Implement RoLongInteger, cast int to LongInt, float to Double

vbuchii:
This implements the [RoLongInteger](https://developer.roku.com/en-ca/docs/references/brightscript/components/rolonginteger.md), also adds support of auto casting `roInt` to `roLongInt`, `roFloat` to `roDouble`.

fixes #364


test/brsTypes/components/RoLongInteger.test.js
```diff
@@ -0,0 +1,81 @@
+const brs = require("brs");
+const { roLongInteger, Int64, BrsBoolean, BrsString, Callable } = brs.types;
+const { Interpreter } = require("../../../lib/interpreter");
+
+describe("roLongInteger", () => {
+    describe("equality", () => {
+        it("compares to intrinsic Int64", () => {
+            let a = new roLongInteger(new Int64(10));
```
alimnios72, line 8: Can you actually use long integers here? I think javascript should be able to handle integers greater than 32 bits

test/e2e/resources/components/roIntrinsics.brs
```diff
@@ -17,5 +18,7 @@ sub main()
     print "Double to string "doubleObject.toStr()
     print "Float to string "floatObject.toStr()
     print "Integer to string "integerObject.toStr()
+    print "LongInteger object "longIntegerObject
```
alimnios72, line 21: this test look pretty similar to the one below, can you check that the type is correct instead here?

test/interpreter/Variables.test.js
```diff
@@ -100,11 +100,11 @@ describe("interpreter variables", () => {
             },
             {
                 lhs: "double#",
-                values: [str, int32, float, int64],
+                values: [str, int32, int64, ""],
```
alimnios72, line 103: what are the empty strings for? will strings be casted to doubles?

src/interpreter/index.ts
```diff
@@ -446,19 +448,25 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
...
-                        location: statement.name.location,
-                    },
-                    right: {
-                        type: value,
-                        location: statement.value.location,
-                    },
-                })
-            );
+            if (requiredType === ValueKind.Int64 && value.kind === ValueKind.Int32) {
+                value = new Int64(value.getValue());
```
alimnios72, line 452: Will this cast all 32 bit integers to 64 bits even small numbers as `1`?

alimnios72 (commented):
This looks pretty good to me, I only have a few comments but this is going in the right track

test/brsTypes/components/RoLongInteger.test.js
vbuchii: It is done in the same way as in [RoDouble.test.js](https://github.com/sjbarag/brs/blob/c8067f2c9f11a6de969a744ab26cacec37a97a4e/test/brsTypes/components/RoDouble.test.js#L8). Can you please add more context to what you are expecting here?

test/interpreter/Variables.test.js
vbuchii: Here are described types that are not allowed for `Double`. So, I removed `float` cuz we made changes to cast a `float` to `Double`. So, we expect that the test should not fail if we get `float`  or `Double` value for a variable with `Double` type.

src/interpreter/index.ts
vbuchii: No, it shouldn't because we have above if statement:`if (requiredType && requiredType !== value.kind) {`. 
So, if we try to write an `Integer` value into a variable with an `Integer` type then we get `Integer`:
`someInteger = 1` ' will be integer (`requiredType` and `value.kind` are the same)

The problem was that when we tried to write into `LongInteger` variable, value with an `Integer` type, it crashed due to type mismatch:
`someLongInt& = 1 `' crashes while running tests but works on Roku
By these changes, we allow setting `Integer` value into a variable with `LongInteger` type, the same with `float` and `Double`.

test/brsTypes/components/RoLongInteger.test.js
vbuchii: done

test/e2e/resources/components/roIntrinsics.brs
vbuchii: done

test/interpreter/Variables.test.js
alimnios72: I get the part of removing `float` from the list but what about the last empty string in the array? I guess we already check for string types in the beginning of the array so why is `""` needed? Also `""` is not a BrsType, it would be `new BrsString("")`

src/interpreter/index.ts
alimnios72: Oh I see, what you explain makes sense but I wonder if this should be right place to do that check. Maybe @sjbarag can help us here

test/interpreter/Variables.test.js
vbuchii: My bad, removing it.

src/interpreter/index.ts
sjbarag: This is probably the only reasonable place to do this kind of conversion for assignments, but I suspect we're missing a few cases (do we need `foo# = 1`?).

We currently do this sort of type promotion in `Callable.ts` (if we expected a Double but received a Float, convert the received Float to a Double) — in fact, the next line basically duplicates the logic from `Callable#getSignatureMismatches()`:

https://github.com/sjbarag/brs/blob/6613a3ee34014c38160560af70f9c822a61420b6/src/brsTypes/Callable.ts#L261-L283

Given that these are separate use cases and _may_ have different rules (we'll need to check all combinations against RBI), it's probably fine to duplicate this logic.  If we find that it's exactly the same, let's open a github issue to implement something like:

```typescript
function tryCoerceType(into: ValueKind, value: BrsType): BrsType {
}
```

that covers all combinations.

src/interpreter/index.ts
```diff
@@ -446,19 +448,25 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
...
-                        type: requiredType,
-                        location: statement.name.location,
-                    },
-                    right: {
-                        type: value,
-                        location: statement.value.location,
-                    },
-                })
-            );
+            if (requiredType === ValueKind.Int64 && value.kind === ValueKind.Int32) {
```
sjbarag, line 451: I suspect we'll need to add an extra branch in `Callable#getSignatureMismatches()` to cover a function that accepts a 64-bit integer and receives a 32-bit:

```brightscript
function addOne(a as LongInteger) as LongInteger
    return a + 1
end function

three% = 3
addOne(three%)
```

I'm not sure how RBI handles the other cases, so please check them all:

| Declared Argument Type (`as ___`) | Passed Argument Type | Converts? |
| ----- | ----- | ----- |
| integer | integer | no |
| integer | float | |
| integer | double | |
| integer | longinteger | |
| float | integer | |
| float | float | no |
| float | double | |
| float | longinteger | |
| double | integer | |
| double | float | yes |
| double | double | no |
| double | longinteger | |
| longinteger | integer | yes |
| longinteger | float | |
| longinteger | double | |
| longinteger | longinteger | no |

test/interpreter/Variables.test.js
```diff
@@ -100,11 +100,11 @@ describe("interpreter variables", () => {
             },
             {
                 lhs: "double#",
-                values: [str, int32, float, int64],
+                values: [str, int32, int64],
             },
             {
                 lhs: "int64&",
-                values: [str, int32, float, double],
+                values: [str, float, double],
```
sjbarag, line 107: Can you add a case that proves that we can now safely assign float values to `double#` variables, and int32 values to `int64&` variables?  We're no longer testing that those assignments cause errors, but there's no test to ensure those assignments do work.

sjbarag (changes requested):
This is off to a great start!  Looks like there's a few extra type pairs we'll need to check against RBI to ensure we're handling those properly, but you're definitely off to a great start @vbuchii . Thanks for the PR!

src/interpreter/index.ts
vbuchii: I've made changes and now it works as described below:

Declared Argument Type (as ___) | Passed Argument Type | Converts?
-- | -- | --
integer | integer | no
integer | float |  yes
integer | double |  fails
integer | longinteger |  fails
float | integer | yes
float | float | no
float | double |  fails
float | longinteger |  fails
double | integer |  fails
double | float | yes
double | double | no
double | longinteger |  fails
longinteger | integer | yes
longinteger | float |  fails
longinteger | double |  fails
longinteger | longinteger | no

On Roku we have next:
Screenshot 2021-01-13 at 16 26 37


As you can see, on Roku we have a bit weird casting, so probably it would be difficult to do the same here. @sjbarag should we do the same in `BRS` interpreter?

test/interpreter/Variables.test.js
vbuchii: done

src/interpreter/index.ts
sjbarag: Yes!  `brs` should match the behavior seen in the reference implementation found on Roku devices as much as possible 😃

vbuchii: Created a separate ticket (#609), as was discussed in private.

sjbarag (approved):
Considering that we don't have support for `roLongInteger` at all right now, it's probably safe to land this.  #609 will need to be fixed pretty quickly though, or we'll open ourselves up to lots of other bugs. @vbuchii can you try to resolve that soon?

Thanks for taking this on!

vbuchii:
> Considering that we don't have support for `roLongInteger` at all right now, it's probably safe to land this. #609 will need to be fixed pretty quickly though, or we'll open ourselves up to lots of other bugs. @vbuchii can you try to resolve that soon?
> 
> Thanks for taking this on!

Sure, I'm planning to do that soon.

Issue #463, opened 2020-07-21
Make a list of missing components/functions/functionality


lkipke:
# Summary

Right now when a test breaks, it's pretty difficult for a developer to tell whether something went wrong in the unit test, or something went wrong in `brs`. Particularly, not every component is implemented, and some components (i.e. `MarkupGrid`) are just "skeletons" that don't actually behave like they do in BrightScript.

We should keep a list of all of the components and global functions that are not yet implemented so that it's easily discoverable for developers.

(Also, it'd be really cool if we were able to somehow leverage that in the code, so we could print messages to the user like "Error: Tried to use FooComponent, but it is not implemented")

sjbarag:
d'oh, that wasn't the full list — that's what I get for manually linking.

PR #601, opened 2021-01-04
doc: List unsupported BrightScript components, functions, and statements

sjbarag:
The maintainers of this `brs` often receive questions that are answered with "X isn't implemented", but determining that X isn't yet implemented is a daunting task for someone not familiar with the project.  Add a list of unsupported BrightScript statements, functions, and components that we can reference.

part of #463


sjbarag:
This is a pretty big list, so I've kept SceneGraph components out for now.  That'll be a separate PR.

lkipke (approved):
🔥

PR #602, opened 2021-01-05
fix(interpreter): allow dynamic functions to return void

lkipke:
# Change Summary

Currently, when a function uses `return` without a value, we treat it as returning `void`. This is a problem for functions with a `dynamic` return type, because void is currently not accepted as a valid return type for dynamic functions. This PR enables `dynamic` functions to accept `void` as a return type.

Alternatively, we could instead evaluate `return` as `return invalid`; but this would require additional logic to handle this case for `sub`s, which (because they're void functions) are _only_ supposed to accept void return types.

### DIFF
--- a/src/interpreter/index.ts
+++ b/src/interpreter/index.ts
@@ -1114,7 +1114,7 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
                     );
                 }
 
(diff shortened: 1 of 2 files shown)


strattonbrazil (approved):
Looks good to me!

lkipke:
It turns out that RBI doesn't allow an empty `return` in a function using the `function` keyword -- it causes a compile error. So this is not a bug, but a feature 😉

PR #604, opened 2021-01-05
fix(stdlib): callFunc should always return a value

lkipke:
# Change Summary

This fixes a bug where the interpreter throws a type mismatch when `CallFunc` doesn't return a value. Because `CallFunc` has a dynamic return type, the interpreter does not allow void.


sjbarag (approved):
Great catch!

alimnios72 (approved):

sjbarag (approved):
Riiiiiiiight we're using `throw` to handle `return`.

PR #607, opened 2021-01-09
doc: Add table of scenegraph node statuses

### LINKED ISSUE #463: Make a list of missing components/functions/functionality

sjbarag:
Like #601 but for scenegraph!  Add table of scenegraph nodes, their implementation status in `brs`, and additional notes.

fixes #463


alimnios72 (approved):
❤️ it

Issue #554, opened 2020-09-29
Add support for try/catch


sjbarag:
The Reference BrightScript Implementation (RBI) added support for `try`/`catch`/`throw` in version 9.4!  We should do the same.

This work will involve the lexer, parser, and interpreter, and may require some extra changes in the standard library to accommodate.

PR #611, opened 2021-01-19
feat(lex,parse): Add stub try/catch implementation

sjbarag:
Implementing `try`/`catch`/`throw` takes a good bit of work in the interpreter, so I'm splitting that into two pull requests.  This one includes parsing `try`/`catch`/`end try` (without `throw`), and always executes _only_ the `try` block to maintain backwards compatibility.  It's a weird partial state, but the alternative is a _massive_ pull request that can't be reasonably reviewed or understood.

see #554

### DIFF
--- a/src/lexer/ReservedWords.ts
+++ b/src/lexer/ReservedWords.ts
@@ -58,6 +58,7 @@ export const ReservedWords = new Set([
  */
 export const KeyWords: { [key: string]: L } = {
     and: L.And,
+    catch: L.Catch,
     dim: L.Dim,
     else: L.Else,
     elseif: L.ElseIf,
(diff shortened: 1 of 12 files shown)


src/lexer/ReservedWords.ts
```diff
@@ -94,6 +98,8 @@ export const KeyWords: { [key: string]: L } = {
     stop: L.Stop,
     sub: L.Sub,
     to: L.To,
+    try: L.Try,
```
alimnios72, line 101: Noob question: why are `try` and `endtry` keywords but not reserved words while `throw` is both?

alimnios72 (approved):
Looks great, thanks for keeping this PR simple. I just have a noob question below

lvcabral:
Try..Catch and Throw implementation was finished in the Roku Community fork:
https://github.com/rokucommunity/brs/releases/tag/v0.46.0

PR #614, opened 2021-01-21
fix(lex): Allow JS object properties as identifiers

### LINKED ISSUE #613: `constructor` identifier causes parse issues in `brs`
Attempting to use the word `constructor` as an identifier in `brs` causes parser errors, despite the fact that this is valid in RBI.

```brightscript
sub main()
    constructor = function()
        return {
            foo: "bar"
        }
    end function

    instance = constructor()
    print instance.foo ' => "bar"
end sub
```

output:

```
$ bin/cli.js /tmp/mwe.brs 
/tmp/derp.brs(2,4-15): Found unexpected token 'constructor'
/tmp/derp.brs(6,4-16): Expected 'end sub' to terminate sub block
/tmp/derp.brs(8,15-26): Found unexpected token 'constructor'
/tmp/derp.brs(10,0-7): Found unexpected token 'end sub'
Error occurred parsing
```

sjbarag:
By using an identifier in a BrightScript file as an index into that object, we accidentally leaked non-Token properties on the `KeyWords` JS object from the lexer, e.g. `return KeyWords["constructor"]`.  Use `hasOwnProperty()` to check if an identifier is a BrightScript keyword to avoid returning JavaScript internals as Tokens.

fixes #613

### DIFF
--- a/src/lexer/Lexer.ts
+++ b/src/lexer/Lexer.ts
@@ -617,12 +617,10 @@ export class Lexer {
                     advance();
                 } // read the next word
 
-                let twoWords = source.slice(start, current);
-                //replace all of the whitespace with a single space character so we can properly match keyword token types
-                twoWords = twoWords.replace(whitespace, " ");
-                let maybeTokenType = KeyWords[twoWords.toLowerCase()];
-                if (maybeTokenType) {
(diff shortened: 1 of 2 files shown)


alimnios72 (approved):
Nice catch

PR #615, opened 2021-01-22
fix(parse,interp): Support unary + operator for numbers

### LINKED ISSUE #227: Allow positive character in for step param
The following is valid brightscript 
`for i = 0 to arr.count() - 1 step +1`
but currently throws out `Found unexpected token '+'`

sjbarag:
BrightScript, like JavaScript and several other languages, supports a leading `+` sign before numeric literals (and variables, incidentally). Check for `+` tokens when parsing unary expressions, and evaluate them at runtime (no-op).

fixes #227


alimnios72 (approved):
+nice

PR #617, opened 2021-01-22
fix(lex): Allow explicit double designator on literals > 10 characters

### LINKED ISSUE #616: Very large double-precision literals with a trailing # are treated like conditional compilation code
In RBI it's perfectly valid to use a numeric literal of more than ten characters and still use the `#` operator to explicitly mark it as a double.  In `brs`, that causes a lexing error:

```brightscript
sub main()
    if (3 >= 9999999999999999999#)
        print "oh no"
    else
        print "phew"
    end if
end sub
```

```
hash-in-conditional.brs(2,32-34): Found unexpected conditional-compilation string '#)'
Error occurred during lexing
```

sjbarag:
RBI forces all literal numbers to double-precision if they're longer than characters, and also allows such a number to be explicitly designated a double with a trailing `#`.  `brs` didn't.  Look for a trailing `#` when lexing long numeric literals, to avoid treating it like a conditional compilation string.

fixes #616

### DIFF
--- a/src/lexer/Lexer.ts
+++ b/src/lexer/Lexer.ts
@@ -474,18 +474,13 @@ export class Lexer {
             let numberOfDigits = containsDecimal ? asString.length - 1 : asString.length;
             let designator = peek().toLowerCase();
 
-            if (numberOfDigits >= 10 && designator !== "&") {
-                // numeric literals over 10 digits with no type designator are implicitly Doubles
-                if (designator === "#") {
-                    // but it's legal to explicitly designate it a double anyway
-                    advance();
-                }
-                addToken(Lexeme.Double, Double.fromString(asString));
-                return;
-            } else if (designator === "#") {
+            if (designator === "#") {
                 // numeric literals ending with "#" are forced to Doubles
                 advance();
-                asString = source.slice(start, current);
+                addToken(Lexeme.Double, Double.fromString(asString));
+                return;
+            } else if (numberOfDigits >= 10 && designator !== "&") {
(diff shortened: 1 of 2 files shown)


sjbarag:
> allow-explicit-double-designation-with-octothorp

ok i admit it i just wanted to use "octothorp" in a sentence.  i don't get many chances.

src/lexer/Lexer.ts
```diff
@@ -476,6 +476,10 @@ export class Lexer {
 
             if (numberOfDigits >= 10 && designator !== "&") {
                 // numeric literals over 10 digits with no type designator are implicitly Doubles
+                if (designator === "#") {
+                    // but it's legal to explicitly designate it a double anyway
```
alimnios72, line 480: Does swapping below `else if` with `if` on line 477 fixes the issue without needing to add this extra `if` block?
Also, not part of this change but line 488 looks like a dupe of 473

sjbarag: Oooh it might!  Let me try.

sjbarag: Good catch re: the duplicate line.  `advance()` changes the value of `current` so _technically_ we should be grabbing the text between `start` and the new `current`.  In practice it doesn't matter, because we pass that to JavaScript's `Float.fromString()` which ignores the trailing `#` that the second `source.slice()` returns anyway 🤷 Removing that too!

alimnios72 (approved):
All those "octothorps" look great

PR #618, opened 2021-01-25
fix(rsg): Ensure hostNode set in <function> implementation

### LINKED ISSUE #585: `no active host node: (internal):(-1)` occurs when attempting to observe a node that was passed in via callFunc
```
Assuming nodeA has field `field1`, and nodeB has function field `func1`

if we attempt to do this inside nodeB's func1:

' // some other component
nodeB.callFunc("func1", nodeA)

' // nodeB.brs
sub func1(nodeA as object)
    m.nodeA = nodeA
    m.nodeA.observeField("field1", "onField1Changed")       ' this emits "BRIGHTSCRIPT: ERROR: roSGNode.ObserveField: no active host node: (internal):(-1)" error in the log
end func
```

sjbarag:
When a function is executed via `callFunc(name, ...args)`, we were accidentally losing the hostNode because we use interpreter.inSubEnv with the optional second parameter to pass in a custom environment. That environment is based on the prototype-like environment created for each component's definition, which can't include an accurate hostnode
based on runtime control flow.  Manually set a hostNode when calling functions via `callFunc`.

fixes #585

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -636,6 +636,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
 
                         subInterpreter.environment.setM(this.m);
                         subInterpreter.environment.setRootM(this.m);
+                        subInterpreter.environment.hostNode = this;
 
                         try {
                             // Determine whether the function should get arguments or not.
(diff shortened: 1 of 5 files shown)


gincwang (approved):
🚀  thanks so much for getting this fixed!!

alimnios72 (approved):
Look great

PR #622, opened 2021-03-01
feat(mocks): mockFunction improvements

lkipke:
# Change summary

This improves `_brs_.mockFunction` in a couple of ways:
1. Allow calling without passing a substitute/mock (i.e. implicit mock)
1. Implement `_brs_.mockReturnValue` (named like [jest](https://jestjs.io/docs/en/mock-function-api#mockfnmockreturnvaluevalue))

### DIFF
--- a/test/e2e/resources/components/mocks/functions/main.brs
+++ b/test/e2e/resources/components/mocks/functions/main.brs
@@ -61,6 +61,27 @@ sub main()
     mock.clearMock()
     print mock.calls.count() ' => 0
     print mock.results.count() ' => 0
+
+    mock.mockReturnValue("foo")
+    spyOnMe(1234)
+    print mock.calls[0][0] ' => 1234
+    print mock.results[0] ' => "foo"
+
+    ' Should be able to call with any number of args
+    spyOnMe()
+    print mock.calls[1].count() ' => 0
+    print mock.results[1] ' => "foo"
+
+    ' Implicit mock
+    mock = _brs_.mockFunction("originalHasArgs")
+    ' should work with no args
+    print originalHasArgs() ' => invalid
+
+    ' should work with args that don't match the original signature
+    print originalHasArgs(1, 2, 3, 4, 5) ' => invalid
+
+    mock.mockReturnValue("mocked!")
+    print originalHasArgs(1, 2, 3, 4, 5) ' => mocked!
 end sub
 
 function mockMe()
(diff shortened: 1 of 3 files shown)


test/e2e/resources/components/mocks/functions/main.brs
```diff
@@ -61,6 +61,27 @@ sub main()
     mock.clearMock()
     print mock.calls.count() ' => 0
     print mock.results.count() ' => 0
+
+    mock.mockReturnValue("foo")
+    spyOnMe(1234)
+    print mock.calls[0][0] ' => 1234
+    print mock.results[0] ' => "foo"
```
sjbarag, line 68: I was going to ask if this matched the `jest.fn()` API and clearly you already thought of that — this is near-identical!  I think the only thing we might want to consider is making `results` entries associative arrays of `{ type: "return" | "throw", value: dynamic }`.  That'll match Jest closer but also makes supporting `try/catch` a bit easier when that happens 😃

sjbarag (changes requested):
I'm sorry I took so long to review this!  This looks great @lkipke!  I've got one small request w.r.t. the types of `mock.results[]` entries, but I'm not sure how you feel about it (nor how much work it'd be).  If it winds up being a ton of work I'm fine with pulling that into a separate PR 😃

test/e2e/resources/components/mocks/functions/main.brs
lkipke: Great idea! When we first implemented the mocks, `try/catch` was not yet a Brightscript feature, which is why we didn't include a `type`. This will be a breaking change, though, and is not strictly related to the improvements in this PR -- can we do that in a separate PR?

sjbarag: > @lkipke Great idea! When we first implemented the mocks, `try/catch` was not yet a Brightscript feature, which is why we didn't include a `type`. This will be a breaking change, though, and is not strictly related to the improvements in this PR -- can we do that in a separate PR?

Sure thing!  I saw you opened #624 so that works for me 😃

alimnios72 (approved):
Change looks great. Is this new API going to be documented in `roca`?

lkipke:
> Change looks great. Is this new API going to be documented in `roca`?

Yep, definitely will, filed this https://github.com/hulu/roca/issues/81

PR #625, opened 2021-03-17
feat(docs): link to docs site

lkipke:
# Summary

This removes all API documentation in favor of redirecting to our docs site, https://hulu.github.io/roca.

PR #626, opened 2021-03-17
fix(docs): fix link to docs site

lkipke:
# Summary

Whoops, in #625 I wrote a link backwards in markdown. This fixes it.


sjbarag (approved):
I saw that and assumed you just wanted the full URL to display inside the `<a/>` and didn't read any further 😅

PR #628, opened 2021-03-18
add double support to Str() (#581)

### LINKED ISSUE #581: Global str function only supports float arguments
Even when the Roku docs say that `str` takes a float argument, in reality it can take double values as well:

```
Brightscript Debugger> myvar = 1605743966.521#
Brightscript Debugger> print type(myvar)
Double
Brightscript Debugger> print str(myvar)
 1.605744e+09
```

The current implementation tries to enforce that float numbers are taken as arguments and it fails if a double is provided
```
 function Str(value as Float) as String:
        * Argument 'value' must be of type Float, but received Double.
```

We should match Roku implementation

strattonbrazil:
This addresses #581, where only floats are accepted by the Str() function. 

**Change**
The implementation for floats and doubles is identical where they need to be converted to a string, and if positive add a space beforehand. The simplest fix was to change the input type to double, so both floats and doubles would be supported. 

Fixes #581

### DIFF
--- a/src/stdlib/String.ts
+++ b/src/stdlib/String.ts
@@ -144,11 +144,11 @@ export const Mid = new Callable(
 );
 
 /**
- * Return a string from a float. If it is positive, prefix it with a space.
+ * Return a string from a float/double. If it is positive, prefix it with a space.
  */
 export const Str = new Callable("Str", {
     signature: {
-        args: [new StdlibArgument("value", ValueKind.Float)],
+        args: [new StdlibArgument("value", ValueKind.Double)],
         returns: ValueKind.String,
     },
     impl: (interpreter: Interpreter, value: Float): BrsString => {
--- a/test/e2e/StdLib.test.js
+++ b/test/e2e/StdLib.test.js
@@ -50,6 +50,7 @@ describe("end to end standard libary", () => {
             "7",
             "10",
             " 3.4",
+            " 9.7",
             "-3",
             "12.34",
             "Mary and Bob",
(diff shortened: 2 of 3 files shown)


src/stdlib/String.ts
```diff
@@ -144,11 +144,11 @@ export const Mid = new Callable(
 );
 
 /**
- * Return a string from a float. If it is positive, prefix it with a space.
+ * Return a string from a float/double. If it is positive, prefix it with a space.
  */
 export const Str = new Callable("Str", {
     signature: {
-        args: [new StdlibArgument("value", ValueKind.Float)],
+        args: [new StdlibArgument("value", ValueKind.Double)],
```
lkipke, line 151: So this function will accept both doubles and floats? Do you happen to know where the logic is where we accept `ValueKind.Float` for signatures that require `ValueKind.Double`?

lkipke (approved):
Looks good, just the one question

src/stdlib/String.ts
strattonbrazil: Good question. Could be wrong, but it's either this code or very similar to it:
https://github.com/sjbarag/brs/blob/main/src/interpreter/index.ts#L453
```
        ... before doing assignment
        if (requiredType && requiredType !== value.kind) {
            if (requiredType === ValueKind.Int64 && value.kind === ValueKind.Int32) {
                value = new Int64(value.getValue());
            } else if (requiredType === ValueKind.Double && value.kind === ValueKind.Float) {
                value = new Double(value.getValue());
       ... now do assignment
```

alimnios72 (approved):
Thanks for fixing this

PR #620, opened 2021-01-27
fix(interp): Coerce function arguments and variable assignments into desired types

vbuchii:
#609
## While running on interpreter:
Screenshot 2021-01-27 at 19 08 40

## On Roku:
Screenshot 2021-01-27 at 19 11 57

PR #630, opened 2021-03-26
fix(int): Match RBI result when coercing Int64 to Int32

sjbarag:
RBI appears to ignore the upper 32 bits when converting a 64-bit integer ("LongInteger") to a 32-bit integer ("integer"), which `brs` didn't previously do.  Use only the least-significant 32 bits when creating a 32-bit integer from an instance of `Long`.

### DIFF
--- a/test/brsTypes/Int32.test.js
+++ b/test/brsTypes/Int32.test.js
@@ -10,23 +10,37 @@ const {
 } = require("../../lib/brsTypes");
 
 describe("Int32", () => {
-    it("truncates initial values to integers", () => {
-        let three = new Int32(3.4);
-        let four = new Int32(3.5);
-        expect(three.getValue()).toBe(3);
-        expect(four.getValue()).toBe(3);
-    });
+    describe("construction", () => {
+        it("truncates initial values to integers", () => {
+            let three = new Int32(3.4);
+            let four = new Int32(3.5);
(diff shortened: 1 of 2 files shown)


sjbarag:
This came up in #620 as one of the issues found while writing tests, but it seems to warrant a separate PR.

test/brsTypes/Int32.test.js
```diff
@@ -10,23 +10,37 @@ const {
...
-        expect(twoFiftyFive.getValue()).toBe(255);
+        it("creates base-16 integers from strings", () => {
+            let twoFiftyFive = Int32.fromString("&hFF");
+            expect(twoFiftyFive.getValue()).toBe(255);
+        });
+
+        it("ignores upper 32 bits of 64-bit input", () => {
+            let large = new Int64(2147483647119);
+            let truncated = new Int32(large.getValue());
+            expect(truncated.getValue()).toBe(-881);
```
strattonbrazil, line 36: Math checks out. :)
```
int main(int argc, char **argv)
{
    int foo = 2147483647119;
    std::cout << foo << std::endl; // prints -881
}
```

strattonbrazil (approved):
Looks great!

lkipke (approved):
what a catch 🕵️

PR #631, opened 2021-03-26
fix(interp): don't mutate function call arguments while type-checking

sjbarag:
While it's a bit of an edge-case with "in the wild BrightScript", it was previously possible for `brs` to accidentally mutate the arguments passed to a function while type-checking those arguments.  Make a shallow copy of the provided arguments before performing type checking and type coercion.


sjbarag:
This might break the existing type coercion — I've gotta do more digging to figure that out 😬

sjbarag:
Combining with #620

PR #633, opened 2021-03-30
chore(infra): Don't publish new nightly version if nothing changed

sjbarag:
We previously published a new nightly build every night, but there was no difference between many of those builds.  It's pretty silly to make npmjs.com keep storing the same package with a different version all the time.  Stop publishing nightly builds when there are no commits since last night's build.


sjbarag:
There's no easy way to test this other than by waiting until the job runs again so I'm going to merge this as soon as the existing checks pass.

PR #634, opened 2021-03-31
fix(infra): build and publish if there are commits since last night

sjbarag:
I negated all of those `if` properties.  Whooooops


sjbarag:
Fixes my goof from #633 😅

alimnios72 (approved):
Nice catch 😅

PR #637, opened 2021-04-07
fix(lexer): Add support for explicit integer literals

### LINKED ISSUE #627: Support "%" type declaration character
In Brightscript numbers with 10 or more digits are treated with the [following logic](https://developer.roku.com/docs/references/brightscript/language/expressions-variables-types.md#literals-constants):

> 1. If a constant contains 10 or more digits, or if D is used in the exponent, that number is double precision. Adding a # declaration character also forces a constant to be double precision.
> 2.  If the number is not double-precision, and if it contains a decimal point, then the number is float. If the number is expressed in exponential notation with E preceding the exponent, the number is float.
> 3. If neither of the above is true of the constant, then it is an integer.

but you can also force long numbers to be casted as integers (as long as they fit in 32 bits) by using the `%` character. 

```
Brightscript Debugger> myvar = 1234567890%
Brightscript Debugger> print getInterface(myvar, "ifDouble")
invalid
Brightscript Debugger> print getinterface(myvar, "ifInt")
<Interface: ifInt>
```

However `%` isn't supported in brs
```
brs> myvar = 1234567890%
REPL(1,18-19): Unexpected character '%'
brs> myvar = 1234567890
brs> print getInterface(myvar, "ifDouble")
<Interface: ifDouble>
brs> print getinterface(myvar, "ifInt")
invalid
```

Vasya-M:
this PR resolves #627 issue
what were done:

- added support for `%` suffix in integer literals
- added unit test for imlementation(lexer)
- added e2e test.

### DIFF
--- a/src/lexer/Lexer.ts
+++ b/src/lexer/Lexer.ts
@@ -543,6 +543,9 @@ export class Lexer {
                 addToken(Lexeme.LongInteger, Int64.fromString(asString));
                 return;
             } else {
+                if (designator === "%") {
+                    advance();
+                } // consume the "%"
                 // otherwise, it's a regular integer
                 addToken(Lexeme.Integer, Int32.fromString(asString));
                 return;
--- a/test/e2e/Syntax.test.js
+++ b/test/e2e/Syntax.test.js
@@ -30,6 +30,7 @@ describe("end to end syntax", () => {
             "true",
             "false",
             "5",
+            "6",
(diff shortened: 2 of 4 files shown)


sjbarag (approved):
I could have sworn we fixed this already, but perhaps I'm mistaken. Thanks for clean patch, @Vasya-M !

Issue #638, opened 2021-04-07
`CreateObject("roSGNode", "node")` returns `invalid`


Vasya-M:
`CreateObject("roSGNode", "node")` should return`"Node"` object, but it returns invalid
works as expected for `CreateObject("roSGNode", "Node")`

PR #639, opened 2021-04-08
feat(stdlib): Add lookupCI for assocarray and Node

### LINKED ISSUE #629: lookupCI() is not supported
`Error: 'lookupCI' is not a function and cannot be called.`

From [Roku SDK](https://developer.roku.com/en-gb/docs/references/brightscript/interfaces/ifassociativearray.md): 
_"LookupCI(key as String) as Dynamic
Description
Same as the Lookup() method except that the key comparison is always case insensitive, regardless of the case mode."_

Vasya-M:
solution $629 issue
what were done:
- fixed `CreateObject("roSGNode", "node")`
- added support for `lookupCI` for AA and Nodes
- added UT and e2e tests

resolves #629  

UPD:
 - Fixed logic of using/saving keys in AA (saving last key used by `[]`, or lowercase key if accessed by `.`)

### DIFF
--- a/src/brsTypes/components/BrsComponent.ts
+++ b/src/brsTypes/components/BrsComponent.ts
@@ -58,9 +58,10 @@ export interface BrsIterable {
     /**
      * Retrieves an element from this component at the provided `index`.
      * @param index the index in this component from which to retrieve the desired element.
+     * @param isCaseSensitiveGet determinate whether operation of getting should be case sensitive or not.
      * @returns the element at `index` if one exists, otherwise throws an Error.
      */
-    get(index: BrsType): BrsType;
+    get(index: BrsType, isCaseSensitiveGet?: boolean): BrsType;
 
-    set(index: BrsType, value: BrsType): BrsInvalid;
+    set(index: BrsType, value: BrsType, isCaseSensitiveSet?: boolean): BrsInvalid;
 }
--- a/src/brsTypes/components/BrsComponent.ts
+++ b/src/brsTypes/components/BrsComponent.ts
@@ -58,10 +58,10 @@ export interface BrsIterable {
     /**
      * Retrieves an element from this component at the provided `index`.
      * @param index the index in this component from which to retrieve the desired element.
-     * @param isCaseSensitiveGet determinate whether operation of getting should be case sensitive or not.
+     * @param isCaseSensitive determinate whether operation of getting should be case sensitive or not.
      * @returns the element at `index` if one exists, otherwise throws an Error.
      */
-    get(index: BrsType, isCaseSensitiveGet?: boolean): BrsType;
+    get(index: BrsType, isCaseSensitive?: boolean): BrsType;
 
-    set(index: BrsType, value: BrsType, isCaseSensitiveSet?: boolean): BrsInvalid;
+    set(index: BrsType, value: BrsType, isCaseSensitive?: boolean): BrsInvalid;
 }
--- a/src/brsTypes/components/ComponentFactory.ts
+++ b/src/brsTypes/components/ComponentFactory.ts
@@ -18,6 +18,7 @@ import {
 
 export enum BrsComponentName {
     Node = "Node",
+    Node_lowerCase = "node",
     Group = "Group",
     LayoutGroup = "LayoutGroup",
     Rectangle = "Rectangle",
@@ -67,6 +68,8 @@ export class ComponentFactory {
             case BrsComponentName.LayoutGroup:
                 return new LayoutGroup([], name);
             case BrsComponentName.Node:
+                return new RoSGNode([], name);
+            case BrsComponentName.Node_lowerCase:
                 return new RoSGNode([], name);
             case BrsComponentName.Rectangle:
                 return new Rectangle([], name);
--- a/src/brsTypes/components/ComponentFactory.ts
+++ b/src/brsTypes/components/ComponentFactory.ts
@@ -18,7 +18,6 @@ import {
 
 export enum BrsComponentName {
     Node = "Node",
-    Node_lowerCase = "node",
     Group = "Group",
     LayoutGroup = "LayoutGroup",
     Rectangle = "Rectangle",
@@ -68,8 +67,6 @@ export class ComponentFactory {
             case BrsComponentName.LayoutGroup:
                 return new LayoutGroup([], name);
             case BrsComponentName.Node:
-                return new RoSGNode([], name);
-            case BrsComponentName.Node_lowerCase:
                 return new RoSGNode([], name);
             case BrsComponentName.Rectangle:
                 return new Rectangle([], name);
--- a/src/brsTypes/components/RoAssociativeArray.ts
+++ b/src/brsTypes/components/RoAssociativeArray.ts
@@ -17,6 +17,10 @@ export interface AAMember {
 export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIterable {
     readonly kind = ValueKind.Object;
     elements = new Map<string, BrsType>();
+    /** keyMap - maps lowercased key to original one which used in this.elements to store value
+     * Main benefit of it is fast case insensitive access
+     */
+    keyMap = new Map<string, string>();
     private modeCaseSensitive: boolean = false;
 
     constructor(elements: AAMember[]) {
@@ -111,25 +115,18 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
 
         let indexValue = isCaseSensitive ? index.value : index.value.toLowerCase();
         this.elements.set(indexValue, value);
+        this.keyMap.set(index.value.toLowerCase(), indexValue);
+
         return BrsInvalid.Instance;
     }
 
     /** if AA is in insensitive mode, it means that we should do insensitive search of real key */
     private findElementKey(elementKey: string, isCaseSensitiveFind = false) {
-        let useCaseSensitiveSearch = this.modeCaseSensitive && isCaseSensitiveFind;
-        elementKey = useCaseSensitiveSearch ? elementKey : elementKey.toLowerCase();
-        if (this.elements.has(elementKey)) {
+        if (this.modeCaseSensitive && isCaseSensitiveFind) {
             return elementKey;
+        } else {
+            return this.keyMap.get(elementKey.toLowerCase());
         }
-
-        let realElementKey = undefined;
-        for (let key of this.elements.keys()) {
-            if ((useCaseSensitiveSearch ? key : key.toLowerCase()) === elementKey) {
-                realElementKey = key;
-                break;
-            }
-        }
-        return realElementKey;
     }
 
     private findElement(elementKey: string, isCaseSensitiveFind = false) {
@@ -145,6 +142,7 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
         },
         impl: (interpreter: Interpreter) => {
             this.elements.clear();
+            this.keyMap.clear();
             return BrsInvalid.Instance;
         },
     });
@@ -158,6 +156,20 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
         impl: (interpreter: Interpreter, str: BrsString) => {
             let key = this.findElementKey(str.value, this.modeCaseSensitive);
             let deleted = key ? this.elements.delete(key) : false;
+            this.keyMap.delete(str.value.toLowerCase());
+
+            // if we have {"DD": 0, "dD": 0},  keyMap["dd"] is pointed to "DD",
+            // and we delets it, then we should find another value for case insensitive accessing ("dD")
+            if (this.modeCaseSensitive) {
+                let lKey = str.value.toLowerCase();
+                for (let key of this.elements.keys()) {
+                    if (key.toLowerCase() === lKey) {
+                        this.keyMap.set(lKey, key);
+                        break;
+                    }
+                }
+            }
+
             return BrsBoolean.from(deleted);
         },
     });
@@ -267,8 +279,7 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
             returns: ValueKind.Dynamic,
         },
         impl: (interpreter: Interpreter, key: BrsString) => {
-            let lKey = key.value;
-            return this.get(new BrsString(lKey), true);
+            return this.get(key, true);
         },
     });
 
@@ -279,8 +290,7 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
             returns: ValueKind.Dynamic,
         },
         impl: (interpreter: Interpreter, key: BrsString) => {
-            let lKey = key.value;
-            return this.get(new BrsString(lKey));
+            return this.get(key);
         },
     });
 
--- a/src/brsTypes/components/RoAssociativeArray.ts
+++ b/src/brsTypes/components/RoAssociativeArray.ts
@@ -259,7 +259,7 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
 
     /** Given a key, returns the value associated with that key.
      * This method is case insensitive either-or case sensitive, depends on whether `setModeCasesensitive` was called or not.
-     * */
+     */
     private lookup = new Callable("lookup", {
         signature: {
             args: [new StdlibArgument("key", ValueKind.String)],
--- a/src/brsTypes/components/RoAssociativeArray.ts
+++ b/src/brsTypes/components/RoAssociativeArray.ts
@@ -76,7 +76,7 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
             .map((value: BrsType) => value);
     }
 
-    get(index: BrsType, isCaseSensitiveGet = false) {
+    get(index: BrsType, isCaseSensitive = false) {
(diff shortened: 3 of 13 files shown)


src/brsTypes/components/ComponentFactory.ts
```diff
@@ -18,6 +18,7 @@ import {
 
 export enum BrsComponentName {
     Node = "Node",
+    Node_lowerCase = "node",
```
lkipke, line 21: Is this a problem with _all_ components? What if I did `createObject("roSGNode", "group")` -- I'm assuming I'd encounter the same issue.

Also I think we should probably do this in a separate PR, just because it's not strictly related to `lookupCI`.

lkipke (changes requested):
I think the `lookupCI` code looks great, but let's pull out the case-sensitive `createObject` code into its own PR 🙂

src/brsTypes/components/ComponentFactory.ts
lkipke: Also, what about cases like `createObject("RoSGNode", "nOdE")`? Should those also create a Node object?

Vasya-M: yep, you are right, all naming for type of node is case insensitive (works for user components too).

Vasya-M: reverted fix for #638 issue

lkipke (approved):
This looks great, thanks for the updates!

src/brsTypes/components/RoAssociativeArray.ts
```diff
@@ -235,6 +239,17 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
         },
     });
 
+    private lookupCI = new Callable("lookupCI", {
```
alimnios72, line 242: Nit, can you add docs to this function please?

src/brsTypes/components/RoSGNode.ts
```diff
@@ -801,6 +802,8 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         },
     });
 
+    private lookupCI = new Callable("lookupCI", this.lookup.signatures[0]);
```
alimnios72, line 805: please add docs as well. Also, is this function expected to do the same as `lookup` or what is the difference between the 2?

test/brsTypes/components/RoAssociativeArray.test.js
```diff
@@ -372,6 +372,20 @@ describe("RoAssociativeArray", () => {
 
                 result2 = aa.get(new BrsString("KeY2"));
                 expect(result2).toBe(v2);
+
+                // check lookupCI method
+                let lookupCI = aa.getMethod("lookupCI");
```
alimnios72, line 377: can you please add a new test suite for these test cases? Following the same template as we did for other functions, i.e.:
```
describe("lookupCI", () => {
...
}
```
Also it would be interesting to test this function when `SetModeCaseSensitive` is set and when it's not

alimnios72 (commented):
overall looks good, a few comments below

test/brsTypes/components/RoAssociativeArray.test.js
Vasya-M: ok, but this method is always is case insensitive, so there is no need test it under different states of `SetModeCaseSensitive`, just need to test if CaseSensitive is enabled
just now tested it and noticed that there is bug
```
brs> aa = {}
brs> aa.SetModeCaseSensitive()
brs> aa["AA"] = 0
brs> ?aa.aa
invalid ' will be 0 in roku
brs> ?aa.lookupCI("aa")
invalid ' will be 0 in roku
brs> ?aa
<Component: roAssociativeArray> =
{
    AA: 0
}

```

src/brsTypes/components/RoSGNode.ts
alimnios72: You had one job brightscript, LOL 😆

test/brsTypes/components/RoAssociativeArray.test.js
alimnios72: Thanks for checking, maybe for posterity is worth having its own different test case hoping that one day Roku fixes the bug and we can update the tests appropriately

src/brsTypes/components/RoSGNode.ts
```diff
@@ -435,30 +436,53 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
...
-        let fieldType = kind || FieldKind.fromBrsType(value);
         let field = this.fields.get(mapKey);
 
         if (!field) {
-            // RBI does not create a new field if the value isn't valid.
-            if (fieldType) {
-                field = new Field(value, fieldType, alwaysNotify);
-                this.fields.set(mapKey, field);
-            }
+            // TODO: change to using interpreter.stderr
```
alimnios72, line 448: If having this creates issues int the current PR (which I think it's breaking unit tests right now) I'm ok with opening a github issue and tackling this work in a separate PR

Vasya-M: yep, did it

src/brsTypes/components/RoAssociativeArray.ts
```diff
@@ -93,18 +92,49 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
         // method with the desired name separately? That last bit would work but it's pretty gross.
         // That'd allow roArrays to have methods with the methods not accessible via `arr["count"]`.
         // Same with RoAssociativeArrays I guess.
-        let indexValue = this.modeCaseSensitive ? index.value : index.value.toLowerCase();
-        return this.elements.get(indexValue) || this.getMethod(index.value) || BrsInvalid.Instance;
+        return (
+            this.findElement(index.value, isCaseSensitiveGet) ||
```
alimnios72, line 96: I think you can get rid of `findElement` function and just do the following:
```
this.elements.get(this.findElementKey(index.value), isCaseSensitiveGet)
```
because if the key is `undefined` the map should return that as well

alimnios72 (commented):
A few last comments below

src/brsTypes/components/RoAssociativeArray.ts
Vasya-M: image
looks like no, it can not accept `undefined`

Vasya-M: image

src/brsTypes/components/RoAssociativeArray.ts
```diff
@@ -93,18 +92,49 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
...
+        let oldKey = this.findElementKey(index.value);
+        if (!this.modeCaseSensitive && oldKey) {
+            this.elements.delete(oldKey);
+        }
+
+        let indexValue = isCaseSensitiveSet ? index.value : index.value.toLowerCase();
         this.elements.set(indexValue, value);
         return BrsInvalid.Instance;
     }
+    /** if AA is in insensitive mode, it means that we should do insensitive search of real key */
```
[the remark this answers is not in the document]
Vasya-M: done

src/brsTypes/components/BrsComponent.ts
```diff
@@ -58,9 +58,10 @@ export interface BrsIterable {
     /**
      * Retrieves an element from this component at the provided `index`.
      * @param index the index in this component from which to retrieve the desired element.
+     * @param isCaseSensitiveGet determinate whether operation of getting should be case sensitive or not.
      * @returns the element at `index` if one exists, otherwise throws an Error.
      */
-    get(index: BrsType): BrsType;
+    get(index: BrsType, isCaseSensitiveGet?: boolean): BrsType;
```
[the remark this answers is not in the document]
Vasya-M: done

test/brsTypes/components/RoAssociativeArray.test.js
```diff
@@ -374,5 +374,40 @@ describe("RoAssociativeArray", () => {
                 expect(result2).toBe(v2);
             });
         });
+        describe("lookupCI", () => {
```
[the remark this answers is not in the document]
Vasya-M: done

src/brsTypes/components/RoAssociativeArray.ts
```diff
@@ -77,7 +76,7 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
             .map((value: BrsType) => value);
     }
 
-    get(index: BrsType) {
+    get(index: BrsType, isCaseSensitiveGet = false) {
```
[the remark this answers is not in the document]
Vasya-M: done

src/brsTypes/components/RoAssociativeArray.ts
```diff
@@ -93,19 +92,51 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
...
+            return elementKey;
+        }
+
+        let realElementKey = undefined;
+        for (let key of this.elements.keys()) {
+            if ((useCaseSensitiveSearch ? key : key.toLowerCase()) === elementKey) {
+                realElementKey = key;
+                break;
+            }
+        }
```
sjbarag, line 131: This seems less than ideal, as it'll cause an `O(n)` lookup for every access once `setCaseInsensitive` is called.  I'm curious if there's an alternative solution we can take that won't require us to iterate across every element every time.  Could we maintain a case-insensitive `Map` next to the case-sensitive one, and write to both for every `set()`? It'd of course require twice as much memory, but would maintain constant time lookups.

src/brsTypes/components/RoAssociativeArray.ts
```diff
@@ -223,8 +258,22 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
...
+    /** Given a key, returns the value associated with that key.
+     * This method is case insensitive either-or case sensitive, depends on whether `setModeCasesensitive` was called or not.
+     */
     private lookup = new Callable("lookup", {
+        signature: {
+            args: [new StdlibArgument("key", ValueKind.String)],
+            returns: ValueKind.Dynamic,
+        },
+        impl: (interpreter: Interpreter, key: BrsString) => {
+            let lKey = key.value;
```
sjbarag, line 270: Perhaps I've missed something — why's this variable `lKey` while the rest are `key`?

sjbarag (changes requested):
This is a good start, @Vasya-M!  I'm a bit worried by the `O(n)` lookups this introduces in case-insensitive mode, so let's see what we can do to keep that from being an exhaustive search 😃

src/brsTypes/components/RoAssociativeArray.ts
Vasya-M: idk why, but it was as is before my changes

src/brsTypes/components/RoAssociativeArray.ts
Vasya-M: Hey, I even forget about it, a great catch. 
So I've taken your idea of using a second map and used it for mapping lower case keys to case sensitive keys. so it will not consume so much memory, will not introduce additional computation and almost saves previous performance. 

but there is still be `O(n)` when we delete keys in case sensitive mode, cuz we can delete key that was used as alias in `KeysMap`

src/brsTypes/components/RoAssociativeArray.ts
Vasya-M: fixed

src/brsTypes/components/RoAssociativeArray.ts
```diff
@@ -17,13 +17,15 @@ export interface AAMember {
 export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIterable {
     readonly kind = ValueKind.Object;
     elements = new Map<string, BrsType>();
+    /** keyMap - maps lowercased key to original one which used in this.elements to store value
+     * Main benefit of it is fast case insensitive access
+     */
```
lkipke, line 22: nit: a little syntax
```suggestion
    /** Maps lowercased element name to original name used in this.elements.
     * Main benefit of it is fast, case-insensitive access.
     */
```

src/brsTypes/components/RoAssociativeArray.ts
```diff
@@ -142,7 +187,7 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
             returns: ValueKind.Void,
         },
         impl: (interpreter: Interpreter, key: BrsString, value: BrsType) => {
-            this.set(key, value);
+            this.set(key, value, true);
```
lkipke, line 190: can we add a comment to make it clear what this boolean refers to?
```suggestion
            this.set(key, value, /* isCaseSensitive */ true);
```

src/brsTypes/components/RoAssociativeArray.ts
```diff
@@ -125,7 +154,23 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
...
         impl: (interpreter: Interpreter, str: BrsString) => {
-            let deleted = this.elements.delete(str.value.toLowerCase());
+            let key = this.findElementKey(str.value, this.modeCaseSensitive);
+            let deleted = key ? this.elements.delete(key) : false;
+
+            let lKey = str.value.toLowerCase();
+            if (this.keyMap.get(lKey) === key) {
+                this.keyMap.delete(lKey);
+                // if we have {"DD": 0, "dD": 0},  keyMap["dd"] is pointed to "DD",
+                // and we delets it, then we should find another value for case insensitive accessing ("dD")
```
lkipke, line 164:
```suggestion
                // and we delete it, then we should find another value for case insensitive accessing ("dD")
```

src/brsTypes/components/RoAssociativeArray.ts
```diff
@@ -182,7 +228,9 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
                 return BrsInvalid.Instance;
             }
 
-            this.elements = new Map<string, BrsType>([...this.elements, ...obj.elements]);
+            obj.elements.forEach((value, key) => {
+                this.set(new BrsString(key), value, true);
+            });
```
lkipke, line 233: what is this logic change needed for?

test/e2e/resources/components/roAssociativeArray.brs
```diff
@@ -11,6 +11,7 @@ sub main()
     print "can delete elements: " aa.delete("baz")              ' => true
     print "can look up elements: " aa.lookup("foo") = "foo"     ' => true
     print "can look up elements (brackets): " aa["foo"] = "foo" ' => true
+    print "can case insensitive look up elements: " aa.lookupCI("foO") = "foo" ' => true
```
lkipke, line 14: Can we add another call to `lookupCI` after we call `setModeCaseSensitive` as well?

lkipke (approved):
Overall, looks great! Thanks for all the hard work here!

src/brsTypes/components/RoAssociativeArray.ts
Vasya-M: cuz with a new approach, we need add lowercase key to map too, and save original key, what `this.set` already do

alimnios72 (approved):
Other than the already given feedback this looks good to me

test/e2e/resources/components/roAssociativeArray.brs
Vasya-M: done

src/brsTypes/components/RoAssociativeArray.ts
```diff
@@ -125,7 +154,23 @@ export class RoAssociativeArray extends BrsComponent implements BrsValue, BrsIte
...
         impl: (interpreter: Interpreter, str: BrsString) => {
-            let deleted = this.elements.delete(str.value.toLowerCase());
+            let key = this.findElementKey(str.value, this.modeCaseSensitive);
+            let deleted = key ? this.elements.delete(key) : false;
+
+            let lKey = str.value.toLowerCase();
+            if (this.keyMap.get(lKey) === key) {
+                this.keyMap.delete(lKey);
+                // if we have {"DD": 0, "dD": 0},  keyMap["dd"] is pointed to "DD",
+                // and we delete it, then we should find another value for case insensitive accessing ("dD")
```
sjbarag, line 164: Thanks for calling this out!  This is an interesting case, and feels like `keyMap` should be `Map<string, Set<string>>`: a map of lowercase keys to all encountered cases of that key.  In your example with `{ "DD": 0, "dD": 1}`, `keyMap.get("dd")` would be the set `("DD", "dD")`.  When `someAA.delete("DD")` is called from BrightScript, this function removes the first element from that set (something like `let newSet = new Set([ ...keyMap.get(lKey) ].slice(1));`) in case-sensitive mode, and removes the entire entry (`this.keyMap.delete(lKey)`) in case-insensitive mode.

Lookups become pretty similar, using the element at index zero of this `Set<string>`.

sjbarag (changes requested):
We're really close!  I've got one last idea to avoid an `O(n)` operation for every `.delete()` call, but otherwise this is looking really good :)  Thanks for sticking with it, @Vasya-M!

src/brsTypes/components/RoAssociativeArray.ts
Vasya-M: done

sjbarag (approved):
Looks great!  Thanks for the changes @Vasya-M 😃

PR #641, opened 2021-04-09
fix(nightly): fetch two commits to detect commits-since-last-build

sjbarag:
The default behavior for GitHub's `actions/checkout@v2` only fetches the
latest commit [1], so our previous attempts to detect commits "since
last night's nightly build" always failed.  Fetch two commits to make it
possible to detect commits in `main` within the past 25 hours.

[1] https://github.com/actions/checkout/tree/25a956c84d5dd820d28caab9f86b8d183aeeff3d#checkout-v2

### DIFF
--- a/.github/workflows/nightly-build.yaml
+++ b/.github/workflows/nightly-build.yaml
@@ -10,12 +10,17 @@
     publish: {
       runs-on: "ubuntu-latest",
       steps: [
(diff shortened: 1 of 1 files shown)


sjbarag:
Merged without review because this only impacts the nightly build (which currently succeeds but doesn't do anything), and I want to make sure things work when this runs at 10:05am PT 😄

PR #643, opened 2021-04-12
feat(stdlib): Implement roAppInfo

### LINKED ISSUE #537: roAppInfo is not implemented.
https://developer.roku.com/en-gb/docs/references/brightscript/components/roappinfo.md

vbuchii:
fixes #537

### DIFF
--- a/src/brsTypes/components/RoAppInfo.ts
+++ b/src/brsTypes/components/RoAppInfo.ts
@@ -70,15 +70,11 @@ export class RoAppInfo extends BrsComponent implements BrsValue {
         impl: (interpreter: Interpreter) => {
             let manifest = this.manifest;
 
-            let subVersions = ["major_version", "minor_version", "build_version"];
-            let version = "";
-
-            for (let i = 0; i < subVersions.length; i++) {
-                let subVersion = manifest.get(subVersions[i]);
-
-                if (subVersion !== undefined) version += subVersion;
-                if (i < subVersions.length - 1) version += ".";
-            }
+            let version = ["major_version", "minor_version", "build_version"]
+                .map((key) => manifest.get(key))
+                .filter((key) => !!key)
+                .join(".");
+            version = version !== "" ? version : "..";
 
             return new BrsString(version);
         },
--- a/src/brsTypes/components/RoAppInfo.ts
+++ b/src/brsTypes/components/RoAppInfo.ts
@@ -8,7 +8,7 @@ import * as PP from "../../preprocessor";
 export class RoAppInfo extends BrsComponent implements BrsValue {
     readonly kind = ValueKind.Object;
 
-    private readonly manifest = PP.getManifestSync(process.cwd());
+    private static _manifest = PP.getManifestSync(process.cwd());
 
     constructor() {
         super("roAppInfo");
@@ -68,7 +68,7 @@ export class RoAppInfo extends BrsComponent implements BrsValue {
             returns: ValueKind.String,
         },
         impl: (interpreter: Interpreter) => {
-            let manifest = this.manifest;
+            let manifest = RoAppInfo._manifest;
 
             let version = ["major_version", "minor_version", "build_version"]
                 .map((key) => manifest.get(key))
@@ -86,7 +86,7 @@ export class RoAppInfo extends BrsComponent implements BrsValue {
             returns: ValueKind.String,
         },
         impl: (interpreter: Interpreter) => {
-            let title = this.manifest.get("title");
+            let title = RoAppInfo._manifest.get("title");
 
             return title !== undefined ? new BrsString(title.toString()) : new BrsString("");
         },
@@ -98,7 +98,7 @@ export class RoAppInfo extends BrsComponent implements BrsValue {
             returns: ValueKind.String,
         },
         impl: (interpreter: Interpreter) => {
-            let subtitle = this.manifest.get("subtitle");
+            let subtitle = RoAppInfo._manifest.get("subtitle");
 
             return subtitle !== undefined ? new BrsString(subtitle.toString()) : new BrsString("");
         },
@@ -128,9 +128,14 @@ export class RoAppInfo extends BrsComponent implements BrsValue {
             returns: ValueKind.String,
         },
         impl: (interpreter: Interpreter, key: BrsString) => {
-            let value = this.manifest.get(key.value);
+            let value = RoAppInfo._manifest.get(key.value);
 
             return value !== undefined ? new BrsString(value.toString()) : new BrsString("");
         },
     });
+
+    /** Sets the manifest for Brightscript functions. */
+    public static set manifest(manifest: PP.Manifest) {
+        RoAppInfo._manifest = manifest;
+    }
 }
--- a/src/brsTypes/components/RoAppInfo.ts
+++ b/src/brsTypes/components/RoAppInfo.ts
@@ -3,12 +3,9 @@ import { BrsComponent } from "./BrsComponent";
 import { BrsType } from "..";
 import { Callable, StdlibArgument } from "../Callable";
 import { Interpreter } from "../../interpreter";
-import * as PP from "../../preprocessor";
 
 export class RoAppInfo extends BrsComponent implements BrsValue {
     readonly kind = ValueKind.Object;
-
-    private static manifest = PP.getManifestSync(process.cwd());
 
     constructor() {
         super("roAppInfo");
@@ -72,7 +69,7 @@ export class RoAppInfo extends BrsComponent implements BrsValue {
             returns: ValueKind.String,
         },
         impl: (interpreter: Interpreter) => {
-            let manifest = RoAppInfo.manifest;
+            let manifest = interpreter.manifest;
 
             let version = ["major_version", "minor_version", "build_version"]
                 .map((key) => manifest.get(key))
@@ -94,7 +91,7 @@ export class RoAppInfo extends BrsComponent implements BrsValue {
             returns: ValueKind.String,
         },
         impl: (interpreter: Interpreter) => {
-            let title = RoAppInfo.manifest.get("title");
+            let title = interpreter.manifest.get("title");
 
             return title != null ? new BrsString(title.toString()) : new BrsString("");
         },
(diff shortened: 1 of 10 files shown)


src/brsTypes/components/RoAppInfo.ts
```diff
@@ -0,0 +1,140 @@
...
+
+            let subVersions = ["major_version", "minor_version", "build_version"];
+            let version = "";
+
+            for (let i = 0; i < subVersions.length; i++) {
+                let subVersion = manifest.get(subVersions[i]);
+
+                if (subVersion !== undefined) version += subVersion;
+                if (i < subVersions.length - 1) version += ".";
+            }
```
lkipke, line 81: I think we could do this in a more idiomatic way:
```suggestion
            let version = ["major_version", "minor_version", "build_version"]
                .map(key => manifest.get(key))
                .filter(key => !!key)
                .concat(".");
```

src/brsTypes/components/RoAppInfo.ts
```diff
@@ -0,0 +1,140 @@
...
+import { BrsComponent } from "./BrsComponent";
+import { BrsType } from "..";
+import { Callable, StdlibArgument } from "../Callable";
+import { Interpreter } from "../../interpreter";
+import * as PP from "../../preprocessor";
+
+export class RoAppInfo extends BrsComponent implements BrsValue {
+    readonly kind = ValueKind.Object;
+
+    private readonly manifest = PP.getManifestSync(process.cwd());
```
lkipke, line 11: Since the manifest doesn't change once the app is running, can we make this a `static` variable, so that we only ever read it once instead of each time we create a new `RoAppInfo`?

test/e2e/resources/components/roAppInfo.brs
```diff
@@ -0,0 +1,11 @@
+sub main()
```
lkipke, line 1: Can we write a test case where we use an actual `manifest` file?

lkipke (commented):
Looking good so far, a few comments

src/brsTypes/components/RoAppInfo.ts
vbuchii: Hi @lkipke 
Thank you for the suggestion. On Roku, the version is returned split by dots (e.g.: `"3.41.2"`) but with the suggested code, we would get` ""4,3,0,."`. Also, if there are no defined versions, on Roku 2 dots are returned `".."`, with that code it would be one: `"."`. So, I've added small changes to align with Roku.

src/brsTypes/components/RoAppInfo.ts
vbuchii: Good call! Will take a look tomorrow.

test/e2e/resources/components/roAppInfo.brs
vbuchii: Will do

src/brsTypes/components/RoAppInfo.ts
```diff
@@ -0,0 +1,141 @@
...
+            returns: ValueKind.String,
+        },
+        impl: (interpreter: Interpreter) => {
+            let manifest = RoAppInfo._manifest;
+
+            let version = ["major_version", "minor_version", "build_version"]
+                .map((key) => manifest.get(key))
+                .filter((key) => !!key)
+                .join(".");
+            version = version !== "" ? version : "..";
```
alimnios72, line 77: I feel like returning `0.0.0` instead of `..` (if no values are provided in manifest) it's a bit more meaningful, what do you think?

alimnios72 (commented):
I left a few comments below but overall this looks pretty good!

src/brsTypes/components/RoAppInfo.ts
```diff
@@ -0,0 +1,141 @@
...
+        signature: {
+            args: [],
+            returns: ValueKind.Boolean,
+        },
+        impl: (interpreter: Interpreter) => {
+            return BrsBoolean.True;
+        },
+    });
+
+    private getVersion = new Callable("getVersion", {
```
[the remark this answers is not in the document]
vbuchii: done

src/extensions/Process.ts
```diff
@@ -40,4 +42,20 @@ export const Process = new RoAssociativeArray([
             },
         }),
     },
+    {
+        name: new BrsString("setManifest"),
```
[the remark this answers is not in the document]
vbuchii: Hi @lkipke, @alimnios72 
Thank you for your feedback. I did try to use a similar approach but was unable to reach the needed result. Maybe it because of using the internal `cwd()` function. Then I found in the [codebase](https://github.com/sjbarag/brs/blob/35d45dfe276b06e137e681094076a3986bab175d/src/brsTypes/components/RoDeviceInfo.ts#L84) `set locale` function and decided to use that approach. 

I added changes to avoid using the additional function, please review 😊

src/brsTypes/components/RoAppInfo.ts
```diff
@@ -0,0 +1,141 @@
...
+
+    private getTitle = new Callable("getTitle", {
+        signature: {
+            args: [],
+            returns: ValueKind.String,
+        },
+        impl: (interpreter: Interpreter) => {
+            let title = RoAppInfo._manifest.get("title");
+
+            return title !== undefined ? new BrsString(title.toString()) : new BrsString("");
```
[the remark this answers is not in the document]
vbuchii: done, thanks

src/brsTypes/components/RoAppInfo.ts
sjbarag: Does that mean including no `build_version` would return `1.2.`?

test/e2e/BrsComponents.test.js
```diff
@@ -719,4 +722,22 @@ describe("end to end brightscript functions", () => {
             "",
         ]);
     });
+
+    test("components/roAppInfo.brs", async () => {
+        manifest = preprocessor.getManifestSync(
+            process.cwd() + "/test/e2e/resources/conditional-compilation"
+        );
+        RoAppInfo.manifest = manifest;
```
sjbarag, line 730: Watch out!  Since this value is never reset, any test that requires an accurate `RoAppInfo.manifest` after this will fail in a very unexpected way.  There's likely a way to solve this problem without explicitly reading an arbitrary manifest as part of this test (more details in a separate comment).

src/brsTypes/components/RoAppInfo.ts
```diff
@@ -0,0 +1,148 @@
...
+import { BrsComponent } from "./BrsComponent";
+import { BrsType } from "..";
+import { Callable, StdlibArgument } from "../Callable";
+import { Interpreter } from "../../interpreter";
+import * as PP from "../../preprocessor";
+
+export class RoAppInfo extends BrsComponent implements BrsValue {
+    readonly kind = ValueKind.Object;
+
+    private static manifest = PP.getManifestSync(process.cwd());
```
sjbarag, line 11: Passing `process.cwd()` will introduce some very subtle bugs when the root directory for `brs` isn't `process.cwd()` — it often isn't in the end-to-end tests in this project!

Since we parse the manifest as basically [the first step of this project](https://github.com/sjbarag/brs/blob/35d45dfe276b06e137e681094076a3986bab175d/src/index.ts#L56-L59) (before brightscript files are even loaded!) I'm curious if we could re-use the parsed manifest, possibly by storing it as a property on the `Interpreter` instance that [gets created early-on](https://github.com/sjbarag/brs/blob/35d45dfe276b06e137e681094076a3986bab175d/src/index.ts#L144-L148)?

src/brsTypes/components/RoAppInfo.ts
alimnios72: Oh wow, didn't know that. Let's keep things aligned with Roku then

sjbarag (changes requested):
We're definitely getting closer here!  I'd like to avoid re-parsing the manifest file if possible, but my bigger concern is around the use of `process.cwd()` in a static property of `RoAppInfo` — that won't work for our end-to-end tests (which you worked around by manually assigning to a private field), and will subtly break other use-cases as they arise.

Do you happen to know which `manifest` file is used when `createObject("roAppInfo")` is called from a [ComponentLibrary](https://developer.roku.com/en-gb/docs/references/scenegraph/control-nodes/componentlibrary.md)?  I imagine that will complicate things here a bit 😃

src/brsTypes/components/RoAppInfo.ts
vbuchii: Thanks @sjbarag 
I've introduced the `manifest` as property on the `Interpreter` instance as you suggested. So, we no longer re-parse manifest and don't use `process.cwd()`.

test/e2e/BrsComponents.test.js
vbuchii: Removed

vbuchii:
> We're definitely getting closer here! I'd like to avoid re-parsing the manifest file if possible, but my bigger concern is around the use of `process.cwd()` in a static property of `RoAppInfo` — that won't work for our end-to-end tests (which you worked around by manually assigning to a private field), and will subtly break other use-cases as they arise.
> 
> Do you happen to know which `manifest` file is used when `createObject("roAppInfo")` is called from a [ComponentLibrary](https://developer.roku.com/en-gb/docs/references/scenegraph/control-nodes/componentlibrary.md)? I imagine that will complicate things here a bit 😃

I was unable to find any related info about that in the Roku SDK. So, I've tested on a sample and seen that it is used the manifest file from root of the project. In other words, regardless of where the object is created, the main manifest is used.

src/interpreter/index.ts
```diff
@@ -94,6 +97,14 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
         return this._environment;
     }
 
+    get manifest() {
+        return this._manifest != null ? this._manifest : new Map<string, ManifestValue>();
```
sjbarag, line 101: Ooooh returning an empty map is a great idea!

sjbarag (approved):
Great work, @vbuchii !  Thanks for the improvements 😃

Issue #644, opened 2021-04-12
Reparent (ifSGNodeChildren function) does not add child node to the new parent


underwoo16:
When calling `child.reparent(newParentNode, false)`  the child's parent is updated but the newParentNode does not add child to its list of children. Example script below:

```brightscript
sub main()
    testParent = createObject("roSGNode", "Node")
    testParent.id = "test-parent"

    testChild = createObject("roSGNode", "Node")
    testChild.id = "test-child"

    print "children before reparent:"
    print testParent.getChildren(-1, 0)

    print "parent before reparent":
    print testChild.getParent()

    testChild.reparent(testParent, false)

    print "children after reparent:"
    print testParent.getChildren(-1, 0)

    print "parent after reparent":
    print testChild.getParent()
end sub
```

Which creates the following output. Note how the children array is empty before and after the call to reparent:
```
children before reparent:
<Component: roArray> =
[
]
parent before reparent
invalid
children after reparent:
<Component: roArray> =
[
]
parent after reparent
<Component: roSGNode:Node> =
{
    change: <Component: roAssociativeArray>
    focusable: false
    focusedchild: invalid
    id: test-parent
}
```

PR #645, opened 2021-04-12
fix(rsg): Reparent does not add child to new parent

underwoo16:
Fixes https://github.com/sjbarag/brs/issues/644

Fixed by calling `appendChildToParent` instead of `setParent` from `reparent`'s implementation.


sjbarag (approved):
Excellent catch, @underwoo16 !  This looks perfect — thanks for the quick fix and clear description! 🎉

Issue #623, opened 2021-03-05
Support `_brs_.testData`


lkipke:
# Summary

Oftentimes when mocking components/functions, it can be helpful to use a globally-scoped (or maybe it'd be scopeless? 🤔) variable that is accessible both from the mock and the test case/suite. Since the mock runs in the source code's `m` scope, this isn't possible with roca's `m.addContext`.

Let's add a property called `_brs_.testData` for this purpose.

PR #646, opened 2021-04-13
feat(extensions): Add _brs_.testData associative array

Vasya-M:
fix for issue #623 
added `testData` to `_brs_` object
exported new method `resetTestData` for cleaning `testData` before runing test suites(test files)
linked roca PR: https://github.com/hulu/roca/pull/88

### DIFF
--- a/src/extensions/index.ts
+++ b/src/extensions/index.ts
@@ -30,3 +30,7 @@ export const _brs_ = new RoAssociativeArray([
     { name: new BrsString("triggerKeyEvent"), value: triggerKeyEvent },
     { name: new BrsString("getStackTrace"), value: GetStackTrace },
 ]);
+
+export function _resetTestData() {
+    _brs_.set(new BrsString("testData"), new RoAssociativeArray([]));
+}
--- a/src/extensions/index.ts
+++ b/src/extensions/index.ts
@@ -26,6 +26,7 @@ export const _brs_ = new RoAssociativeArray([
     { name: new BrsString("runInScope"), value: RunInScope },
     { name: new BrsString("process"), value: Process },
     { name: new BrsString("global"), value: mGlobal },
+    { name: new BrsString("testData"), value: new RoAssociativeArray([]) },
     { name: new BrsString("triggerKeyEvent"), value: triggerKeyEvent },
     { name: new BrsString("getStackTrace"), value: GetStackTrace },
 ]);
--- a/src/extensions/index.ts
+++ b/src/extensions/index.ts
@@ -31,6 +31,7 @@ export const _brs_ = new RoAssociativeArray([
     { name: new BrsString("getStackTrace"), value: GetStackTrace },
 ]);
 
-export function _resetTestData() {
+/** resets _brs_.testData values to `{}` in brightscript representation */
+export function resetTestData() {
     _brs_.set(new BrsString("testData"), new RoAssociativeArray([]));
 }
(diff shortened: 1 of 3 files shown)


src/index.ts
```diff
@@ -16,6 +16,7 @@ import {
 } from "./componentprocessor";
 import { Parser } from "./parser";
 import { Interpreter, ExecutionOptions, defaultExecutionOptions } from "./interpreter";
+import { _resetTestData } from "./extensions";
```
lkipke, line 19: Instead of importing and re-exporting the function, we should be able to do this:
```suggestion
export { resetTestData } from "./extensions";
```

src/extensions/index.ts
```diff
@@ -26,6 +26,11 @@ export const _brs_ = new RoAssociativeArray([
     { name: new BrsString("runInScope"), value: RunInScope },
     { name: new BrsString("process"), value: Process },
     { name: new BrsString("global"), value: mGlobal },
+    { name: new BrsString("testData"), value: new RoAssociativeArray([]) },
     { name: new BrsString("triggerKeyEvent"), value: triggerKeyEvent },
     { name: new BrsString("getStackTrace"), value: GetStackTrace },
 ]);
+
+export function _resetTestData() {
```
lkipke, line 34: Can we add a JSDoc comment above this function to describe what it does?

lkipke (changes requested):
Looks great, just a couple of small nits 😄

lkipke (approved):
Makes sense to me! Can we also add a test case for this?

alimnios72 (approved):
+1 to test case but changes look good

test/extensions/testData.test.js
```diff
@@ -0,0 +1,29 @@
...
+        expect(testData.elements.size).toBe(0);
+        testData.set(new BrsString("foo"), BrsInvalid.Instance);
+        expect(testData.elements.size).toBe(1);
+
+        testData = new Interpreter().environment
+            .get(identifier("_brs_"))
+            .get(new BrsString("testData"));
+        expect(testData.elements.size).toBe(1);
+
+        brs.resetTestData(); // will clear testData for next new Interpreter instances
```
sjbarag, line 22: Do we want the `_brs_.testData` to persist across `interpreter` instances?  IIRC there's only one created as part of `roca` anyway.  We'll still want this type of behavior, but perhaps simpler solution would be for each `Interpreter` instance to get its own copy of the `_brs_` extensions 🤔

src/index.ts
```diff
@@ -16,6 +16,7 @@ import {
 } from "./componentprocessor";
 import { Parser } from "./parser";
 import { Interpreter, ExecutionOptions, defaultExecutionOptions } from "./interpreter";
+export { resetTestData } from "./extensions";
```
sjbarag, line 19: I'm not sure that this needs to be exposed from the root of the module.  If it does need to be exposed, I'd like for us to find (or add) a new property to put it on instead of it being a sibling of `brs.lexer`, `brs.parser`, `brs.types`, etc.  Those are all very low-level concepts, and `resetTestData` doesn't match those 😃

sjbarag (changes requested):
Thanks for tackling this, @Vasya-M!  It looks like `_brs_` persists across `Interpreter` instances, which may be the motivation behind exposing `resetTestData()` outside of this module.  I'm curious if we could ensure each `Interpreter` instance gets its own copy of `_brs_` and solve this problem without having to export a new function that consumers might forget to call!

src/index.ts
```diff
@@ -230,6 +231,7 @@ export async function createExecuteWithScope(
     interpreter.errors = [];
 
     return (filenames: string[], args: BrsTypes.BrsType[]) => {
+        resetTestData();
```
lkipke, line 234: I like this solution better than exporting it!

lkipke (approved):

sjbarag (approved):
Great idea!

PR #647, opened 2021-04-14
fix(interpreter): Support case insensitive names for functions in components

### LINKED ISSUE #502: Declaring Init() with capital "i" on node throws error when extending that node.
Let's suppose I have the following nodes:

**SomeNode.xml**
```
<component name="SomeNode" extends="Group">
    <script type="text/brightscript" uri="SomeNode.brs" />
</component>
```
**SomeNode.brs**
```
sub Init() 'NOTICE: Capital 'i' here (this is heavily used through SGDEX)
    ?"Initing SomeNode..."
end sub

function SomeNode_action(param=invalid)
    ?"SomeNode action  called."
    return param
end function
```

Now let's extend that **SomeNode**:

**ExtendedSomeNode.xml**
```
<component name="ExtendedSomeNode" extends="SomeNode">
    <script type="text/brightscript" uri="ExtendedSomeNode.brs" />
</component>
```
**ExtendedSomeNode.brs**
```
sub init()
    ?"Initing ExtendedSomeNode..."

    SomeNode_action()
end sub
```

Now on our **main.brs** let's do:
```
sub main(args=invalid)
  eSN = CreateObject("roSGNode", "ExtendedSomeNode")
  ?eSN
end sub
```

The following warning is shown:

> brs-bug-sample/components/SomeNode/SomeNode.brs(1,4-8): Attempting to declare function 'Init', but a property of that name already exists in this scope.

And also see that `SomeNode_action()` is not in the scope of **ExtendedSomeNode**:
> brs-bug-sample/components/ExtendedSomeNode/ExtendedSomeNode.brs(4,20-21): 'SomeNode_action' is not a function and cannot be called.

If I change `Init()`to `init()`on **SomeNode**, code works as expected.

alimnios72:
In RBI functions with the same name are always overriden (despite the casing) if extending components with the only exception of the `Init` method which is a special case. With this fix, it doesn't matter what convention you use to name your function as it's going to be converted to lowerCase to fix the override issue we were having.
In addition to the fix, I added a test case to functions to verify that casing doesn't matter when naming functions.

Fixes #502

### DIFF
--- a/src/parser/ComponentScopeResolver.ts
+++ b/src/parser/ComponentScopeResolver.ts
@@ -45,7 +45,7 @@ export class ComponentScopeResolver {
         let statementMemo = new Set(
             statements
                 .filter((_): _ is Stmt.Function => true)
-                .map((statement) => statement.name.text)
+                .map((statement) => statement.name.text.toLowerCase())
         );
         while (statementMap.length > 0) {
             let extendedFns = statementMap.shift() || [];
@@ -53,7 +53,7 @@ export class ComponentScopeResolver {
                 extendedFns
                     .filter((_): _ is Stmt.Function => true)
                     .filter((statement) => {
-                        let statementName = statement.name.text;
+                        let statementName = statement.name.text.toLowerCase();
                         let haveFnName = statementMemo.has(statementName);
                         if (!haveFnName) {
                             statementMemo.add(statementName);
--- a/test/e2e/resources/function/casing.brs
+++ b/test/e2e/resources/function/casing.brs
@@ -1,0 +1,6 @@
+function callMeInsensitive()
+    print "but I'm only interested in..."
+end function
+
+callmeinsensitive()
+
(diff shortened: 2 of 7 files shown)


test/e2e/resources/function/casing.brs
```diff
@@ -0,0 +1,6 @@
+function callMeInsensitive()
+    print "but I'm only interested in..."
```
sjbarag, line 2: dogs?  I'm going to guess you're only interested in dogs. 🐕

src/parser/ComponentScopeResolver.ts
```diff
@@ -45,15 +45,15 @@ export class ComponentScopeResolver {
...
             let extendedFns = statementMap.shift() || [];
             statements = statements.concat(
                 extendedFns
                     .filter((_): _ is Stmt.Function => true)
                     .filter((statement) => {
-                        let statementName = statement.name.text;
+                        let statementName = statement.name.text.toLowerCase();
                         let haveFnName = statementMemo.has(statementName);
                         if (!haveFnName) {
                             statementMemo.add(statementName);
```
sjbarag, line 59: That was easy!

sjbarag (approved):
This looks great!  Thanks for the clear description @alimnios72 🎉

PR #649, opened 2021-04-23
fix(extensions): reset mocks between executions

lkipke:
# Change summary

This resets mocks between executions that share the same base scope, to prevent them from polluting/interfering with each other. Got the idea from the implementation in #646!

Related: https://github.com/hulu/roca/issues/89


alimnios72 (approved):
Looks great!

sjbarag (approved):
Sorry for the delay here!

PR #654, opened 2021-05-27
feat(lex): Emit comments from the lexer

sjbarag:
Comment nodes are handy for non-executing tooling, and allow tools like `eslint` to automatically ignore rules based on specific directives. Emit `eslint`-compliant comment nodes from the lexer for consumption by external tools.


alimnios72 (approved):
Changes looks good to me

PR #658, opened 2021-06-03
chore(deps): Upgrade most dev dependencies

sjbarag:
The upgrade from husky 4 to husky 6 will require some manual intervention.  Upgrade all other dependencies to the latest version that still supports node@10.

### DIFF
--- a/src/parser/Parser.ts
+++ b/src/parser/Parser.ts
@@ -1084,11 +1084,8 @@ export class Parser {
         function printStatement(...additionalterminators: BlockTerminator[]): Stmt.Print {
             let printKeyword = advance();
 
-            let values: (
-                | Expr.Expression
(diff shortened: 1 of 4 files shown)


src/parser/Parser.ts
```diff
@@ -1084,11 +1084,8 @@ export class Parser {
         function printStatement(...additionalterminators: BlockTerminator[]): Stmt.Print {
             let printKeyword = advance();
 
-            let values: (
-                | Expr.Expression
-                | Stmt.PrintSeparator.Tab
-                | Stmt.PrintSeparator.Space
-            )[] = [];
+            let values: (Expr.Expression | Stmt.PrintSeparator.Tab | Stmt.PrintSeparator.Space)[] =
+                [];
```
sjbarag, line 1088: thanks prettier, i hate it

PR #659, opened 2021-06-09
fix(interp): allow functions to be typed as objects

lkipke:
# Changes

This fixes a bug where if you tried to do:
```brs
sub foo(arg as object) : end sub
sub bar() : end sub

foo(bar) ' runtime error
```
`brs` would throw an ugly error. RBI allows this type-casting, so we should too!

This error was actually getting thrown a bunch in the sister project, https://github.com/hulu/roca, since `brs@0.44.0`, because all the `describe` and `it` functions use the `as object` type for functions.

### DIFF
--- a/src/brsTypes/coercion.ts
+++ b/src/brsTypes/coercion.ts
@@ -25,6 +25,9 @@ export function tryCoerce(value: BrsType, target: ValueKind): BrsType | undefine
         } else if (value instanceof BrsComponent) {
             // types that are always objects should be returned unmodified
             return value;
+        } else if (value.kind === ValueKind.Callable) {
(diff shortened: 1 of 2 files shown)


alimnios72 (approved):
Great catch!

sjbarag (approved):
Looks great! I can't remember if functions can be boxed into `roFunction` so we may want to take a follow-up task to get that right. Excellent find - nice job digging through the new coercion code, @lkipke!

sjbarag:
This probably would have been caught if I released an RC last week instead of a full version. 🤔 perhaps we should make that part of the release flow?

PR #660, opened 2021-06-09
fix(stdlib): make RoSGNode#update case insensitive

lkipke:
# Change summary

In #629 there were a bunch of changes to `RoAssociativeArray` for case-sensitivity, namely preserving key case by default. Because `RoSGNode#update` uses an assoc array and `RoSGNode` stores all its fields lower case, we need to make it _not_ case-sensitive.


alimnios72 (approved):
Looks great

PR #663, opened 2021-06-21
fix(stdlib, interp): fix `RoSGNode#callFunc` argument error reporting

lkipke:
# Summary

This fixes a bug where we would throw a _Javascript_ error and display the resulting stack trace to users who used `RoSGNode#callFunc` with incorrect/missing arguments. I also consolidated our "incorrect arguments" error and removed some `finally`s that weren't behaving the way we expected them to.

### DIFF
--- a/src/brsTypes/Callable.ts
+++ b/src/brsTypes/Callable.ts
@@ -3,12 +3,7 @@ import * as Brs from ".";
 import * as Expr from "../parser/Expression";
 import { Scope } from "../interpreter/Environment";
 import { Location } from "../lexer";
-import { Int32 } from "./Int32";
-import { Float } from "./Float";
-import { Double } from "./Double";
-import { Int64 } from "./Int64";
 import { tryCoerce } from "./coercion";
-import { BrsError } from "../Error";
 import { generateArgumentMismatchError } from "../interpreter/ArgumentMismatch";
 
 /** An argument to a BrightScript `function` or `sub`. */
--- a/src/brsTypes/Callable.ts
+++ b/src/brsTypes/Callable.ts
@@ -8,6 +8,8 @@ import { Float } from "./Float";
 import { Double } from "./Double";
 import { Int64 } from "./Int64";
 import { tryCoerce } from "./coercion";
+import { BrsError } from "../Error";
+import { generateArgumentMismatchError } from "../interpreter/ArgumentMismatch";
 
 /** An argument to a BrightScript `function` or `sub`. */
 export interface Argument {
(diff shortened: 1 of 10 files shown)


src/interpreter/ArgumentMismatch.ts
```diff
@@ -0,0 +1,59 @@
+import { ValueKind, SignatureAndMismatches, MismatchReason, Callable, BrsType } from "../brsTypes";
```
lkipke, line 1: This is all just copy-pasta'd from `src/interpreter/index.ts`

src/interpreter/index.ts
```diff
@@ -237,15 +234,19 @@ export class Interpreter implements Expr.Visitor<BrsType>, Stmt.Visitor<BrsType>
...
+        let returnValue;
         try {
             this._environment = newEnv;
-            return func(this);
+            returnValue = func(this);
+            this._environment = originalEnvironment;
+            this._environment.setFocusedNode(newEnv.getFocusedNode());
         } catch (err) {
-            throw err;
-        } finally {
```
lkipke, line 245: it turns out, if you try to `throw` an error in `catch`, but you _also_ have a `finally`, then the `throw` is just ignored. so we were throwing errors but not propagating them upwards.

test/e2e/CallFunc.test.js
```diff
@@ -61,4 +60,32 @@ describe("callFunc", () => {
...
+
+        expect(
+            allArgs(outputStreams.stderr.write)
+                .filter((arg) => arg !== "\n")
+                .map((arg) => arg.trim())
+        ).toEqual([
+            expect.stringContaining(
+                `test/e2e/resources/components/call-func/scripts/throw/wrong-arg-type.brs(5,4-46): Provided arguments don't match stronglyTyped's signature.
+    function stronglyTyped(arg1 as String) as String:
+        * Argument 'arg1' must be of type String, but received Object.`
```
sjbarag, line 75: One of these days we'll pull in `ts-dedent` (https://www.npmjs.com/package/ts-dedent) and make these strings pretty 😜

sjbarag (approved):
Thanks for digging into this, @lkipke !

src/brsTypes/Callable.ts
```diff
@@ -8,6 +8,8 @@ import { Float } from "./Float";
 import { Double } from "./Double";
 import { Int64 } from "./Int64";
 import { tryCoerce } from "./coercion";
+import { BrsError } from "../Error";
```
alimnios72, line 11: was this import used?

alimnios72 (approved):
Looks good

PR #664, opened 2021-06-30
fix(ast): Rename `location` to `loc` in AST

sjbarag:
BREAKING CHANGE: There's nothing inherently wrong with having AST node locations in a property called `node`, but it does break compatibility with more common ASTs like the `estree` standard and makes integration with tools that understand `estree` significantly more difficult. Rename `location` to `loc` in the entire `brs` syntax tree to improve `estree` compatibility and enable more advanced `eslint` rules.


sjbarag:
This will require a few small changes in https://github.com/hulu/roca when this lands.  If we'd already hit 1.0 I'd say this requires 2.0.0, buuuuuuuuut we haven't.  Once I add `range` support to those nodes (luckily just an additive change) we might be able to consider a 1.0 release 🤔

sjbarag:
Either way, this probably warrants a `-rc` release? 🤔

alimnios72 (approved):
I was intimidated by the huge amount of files change but it was actually pretty easy to review.

sjbarag:
(Cleaning up Github stuff; Hi, anyone who sees this notification!)

PR #665, opened 2021-08-09
[NO-ISSUE] - Added boundingrect function to node

underwoo16:
Implemented boundingRect function for roSGNode with zero-defaulted values

https://developer.roku.com/docs/references/brightscript/interfaces/ifsgnodeboundingrect.md

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -373,6 +373,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
                 this.issubtype,
                 this.parentsubtype,
             ],
+            ifSGNodeBoundingRect: [this.boundingRect],
         });
     }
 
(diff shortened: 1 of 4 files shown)


lkipke (approved):
I'm fine with this, given that we have several other components with skeleton implementations. Can you also open a PR into `roca` to update the docs? Probably this one specifically: [implementation-gaps.md](https://github.com/hulu/roca/blob/main/docs/api/reference/implementation-gaps.md)

underwoo16:
> I'm fine with this, given that we have several other components with skeleton implementations. Can you also open a PR into `roca` to update the docs? Probably this one specifically: [implementation-gaps.md](https://github.com/hulu/roca/blob/main/docs/api/reference/implementation-gaps.md)

I think the comment that it is `mostly complete` is still accurate since I only implemented one out of ~7 methods from this interface. I have never seen the other methods used so I only implemented `boundingrect`. 

Would you prefer I implement all of the methods or file an issue to complete the missing methods?

sjbarag:
> > I'm fine with this, given that we have several other components with skeleton implementations. Can you also open a PR into `roca` to update the docs? Probably this one specifically: [implementation-gaps.md](https://github.com/hulu/roca/blob/main/docs/api/reference/implementation-gaps.md)
> 
> I think the comment that it is `mostly complete` is still accurate since I only implemented one out of ~7 methods from this interface. I have never seen the other methods used so I only implemented `boundingrect`.
> 
> Would you prefer I implement all of the methods or file an issue to complete the missing methods?

You're fine as-is for now 🤷 It's still mostly-complete, and we're getting closer every day.

sjbarag (approved):

PR #672, opened 2021-12-01
Implement clone

anyatokar:
Implemented [`.clone(isDeepCopy as Boolean)`](https://developer.roku.com/en-gb/docs/references/brightscript/interfaces/ifsgnodedict.md) for ifSGNodeDict interface. Both shallow and deep clones copy Fields but only the deep clone copies Children.

Added associated unit and e2e tests.

### DIFF
--- a/src/brsTypes/components/RoSGNode.ts
+++ b/src/brsTypes/components/RoSGNode.ts
@@ -373,7 +373,6 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
                 this.issubtype,
                 this.parentsubtype,
                 this.clone,
-                this.clonehelper,
             ],
             ifSGNodeBoundingRect: [this.boundingRect],
         });
@@ -1697,72 +1696,22 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         },
     });
 
-    private clonehelper = new Callable("clonehelper", {
-        signature: {
-            args: [
-                new StdlibArgument("toBeCloned", ValueKind.Object),
-                new StdlibArgument("isDeepCopy", ValueKind.Boolean),
-            ],
-            returns: ValueKind.Object,
-        },
-        impl: (interpreter: Interpreter, toBeCloned: RoSGNode, isDeepCopy: BrsBoolean) => {
-            let copy = createNodeByType(interpreter, new BrsString(toBeCloned.nodeSubtype));
-
-            let originalFields = toBeCloned.getFields();
-            let copiedFields = new RoAssociativeArray([]);
-
-            let addReplace = copiedFields.getMethod("addreplace");
-
-            for (let [key, value] of originalFields) {
-                if (addReplace) {
-                    addReplace.call(
-                        interpreter,
-                        new BrsString(key),
-                        getBrsValueFromFieldType(value.getType(), value.toString())
-                    );
-                }
-            }
-
-            if (!(copy instanceof BrsInvalid)) {
-                let update = copy.getMethod("update");
-
-                if (update) {
-                    update.call(interpreter, copiedFields, BrsBoolean.True);
-                }
-
-                if (isDeepCopy.toBoolean()) {
-                    let appendChild = copy.getMethod("appendchild");
-
-                    for (let child of toBeCloned.children) {
-                        let clone = child.getMethod("cloneHelper");
-
-                        if (clone && appendChild) {
-                            let childCopy = clone.call(interpreter, child, BrsBoolean.True);
-                            appendChild.call(interpreter, childCopy);
-                        }
-                    }
-                }
-            }
-
-            return copy;
-        },
-    });
-
     private clone = new Callable("clone", {
         signature: {
             args: [new StdlibArgument("isDeepCopy", ValueKind.Boolean)],
             returns: ValueKind.Object,
         },
         impl: (interpreter: Interpreter, isDeepCopy: BrsBoolean) => {
-            let cloneHelper = this.getMethod("clonehelper");
+            return cloneNode(interpreter, this, isDeepCopy.toBoolean());
+            // let cloneHelper = this.getMethod("clonehelper");
 
-            if (cloneHelper) {
-                let copy = cloneHelper.call(interpreter, this, isDeepCopy);
+            // if (cloneHelper) {
+            //     let copy = cloneHelper.call(interpreter, this, isDeepCopy);
 
-                return copy;
-            }
+            //     return copy;
+            // }
 
-            return BrsInvalid.Instance;
+            // return BrsInvalid.Instance;
         },
     });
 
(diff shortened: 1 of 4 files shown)


src/brsTypes/components/RoSGNode.ts
```diff
@@ -1914,3 +1925,55 @@ function addChildren(
...
+    // Copy the fields.
+    for (let [key, value] of originalFields) {
+        if (addReplace) {
+            addReplace.call(
+                interpreter,
+                new BrsString(key),
+                getBrsValueFromFieldType(value.getType(), value.toString())
+            );
+        }
+    }
```
strattonbrazil: All this can be moved into the `if copy` since it won't be used if copy isn't valid.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1914,3 +1925,55 @@ function addChildren(
...
+
+    if (!(copy instanceof BrsInvalid)) {
```
strattonbrazil: Can this instead by `(copy instanceof RoSGNode)`? Or did you prefer this way?

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1914,3 +1925,55 @@ function addChildren(
...
+        let update = copy.getMethod("update");
+
+        if (update) {
+            update.call(interpreter, copiedFields, BrsBoolean.True);
+        }
+
+        // A deep clone also copies children.
+        if (isDeepCopy) {
+            let appendChild = copy.getMethod("appendchild");
```
strattonbrazil, line 1961: I think we can use the `appendChildToParent` function directly to remove this line and replace the one below with:
```
copy.appendChildToParent(childCopy);
```

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1914,3 +1925,55 @@ function addChildren(
...
+            let getChildren = toBeCloned.getMethod("getchildren");
```
strattonbrazil, line 1962: This code would be a little cleaner if we exposed `getChildren` in typescript. I wonder if that's worth it. Then the callable would just wrap that response in an `RoArray`.

strattonbrazil (approved):
Looks good to me. A few suggestions that might clean things up a bit.

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1323,7 +1323,7 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
                 let childNodesElements = child_nodes.getElements();
                 if (childNodesElements.length !== 0) {
                     childNodesElements.forEach((childNode) => {
-                        if (childNode instanceof RoSGNode) {
+                        {
```
strattonbrazil: Can we remove these brackets?

underwoo16 (approved):
I agree with @strattonbrazil suggestions, otherwise this looks great 👍🏼

src/brsTypes/components/RoSGNode.ts
alimnios72: +1 to this

src/brsTypes/components/RoSGNode.ts
```diff
@@ -1695,6 +1696,16 @@ export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
         },
     });
 
+    private clone = new Callable("clone", {
```
alimnios72, line 1700: Please add docs to this method, it could be exactly the same docs found in the roku devs webpage

src/brsTypes/components/RoSGNode.ts
anyatokar: `if (childNode instanceof RoSGNode)` was deleted accidentally, I didn't intend to change the `append children` method 🤭 I am putting the entire line back in.

src/brsTypes/components/RoSGNode.ts
anyatokar: @alimnios72 thanks for the suggestion! Do you mean the description of the `clone` method from the BrightScript docs? If so, done!

src/brsTypes/components/RoSGNode.ts
anyatokar: @strattonbrazil I think I would have to expose the `appendChildtoParent` method to be able to use it here. Since I don't see a precedent for this in the `addChildren` function above which also uses the `appendChild` callable, I think I will leave it as is. Let me know if you feel strongly about this one though!

src/brsTypes/components/RoSGNode.ts
alimnios72: yup, that was it. Thanks

alimnios72 (approved):
Thanks for the changes

src/brsTypes/components/RoSGNode.ts
```diff
@@ -296,7 +296,7 @@ function isSubtypeCheck(currentNodeType: string, checkType: string): boolean {
 export class RoSGNode extends BrsComponent implements BrsValue, BrsIterable {
     readonly kind = ValueKind.Object;
     private fields = new Map<string, Field>();
-    private children: RoSGNode[] = [];
+    public children: RoSGNode[] = [];
```
strattonbrazil: I wouldn't want _children_ directly exposed as public. This could lead to bug down the line where the children array is altered, but we don't, for example, fire the callbacks for listeners watching the children. 

Instead I'd create a utility function like this:
```
private function sliceChildren(numChildren, index) : RoSGNode[] {
            // stolen from getChildren()
            let numChildrenValue = num_children.getValue();
            let indexValue = index.getValue();
            let childrenSize = this.children.length;
            if (numChildrenValue <= -1 && indexValue === 0) {
                //short hand to return all children
                return this.children.slice();
            } else if (numChildrenValue <= 0 || indexValue < 0 || indexValue >= childrenSize) {
                //these never return any children
                return [];
            } else {
                //only valid cases
                return this.children.slice(indexValue, indexValue + numChildrenValue);
            }

}
```

And then the `getChildren` callable would just look like this:
```
return RoArray(this.sliceChildren(...));
```

anyatokar: @strattonbrazil Thanks, I was not sure if I was on the right track there so that's helpful! 
So, exposing `children` the way I did for RoSGNode would be more dangerous than how [`children` are exposed](https://github.com/sjbarag/brs/blob/0505973a011453e41bca4bc538ad1364b73413fc/src/componentprocessor/index.ts#L56) in the [AddChildren](https://github.com/sjbarag/brs/blob/0505973a011453e41bca4bc538ad1364b73413fc/src/brsTypes/components/RoSGNode.ts#L1897) method above for the `ComponentNode`? I am wondering how the two situations are different for my understanding.

strattonbrazil: Sure, as I understand it the input to _addChildren_ references an object that has a children array and is basically just a block of data transporting between the interpreter and this function. It can be used and modified as its about to be descoped. 

In the case of _RoSGNode_, that _children_ reference is actually a member variable and by restricting access to it we can provide more safeguards to how its accessed. In the example I gave not letting an outside component insert, remove, or clear children without doing any necessary auxiliary tasks. 

For example:
```
    private children = ...
    public function appendChild(child) {
        children.append(child);

        // bookkeeping like to notify observers of this node's _change_ interface that its children have been altered
        ...
    }
```

If one were to expose children directly it introduces opportunities for parts of the code outside RoSGNode to avoid any necessary code for maintaining and messaging children.

anyatokar: Ah I see. Thank you!

anyatokar: This PR is not ready for merge, I had to prioritize another ticket and will get back to this this week.

anyatokar: I decided not to pursue this refactor due to time constraints :)

strattonbrazil (approved):

lvcabral:
@anyatokar this project is stalled and the Roku Community has forked it (see @sjbarag suggestion on https://github.com/sjbarag/brs/issues/681) do you mind re-pushing your PR on https://github.com/rokucommunity/brs ?

PR #686, opened 2023-12-01
Add create package script

TwitchBronBron:
Adds the `create-package` label support that will auto-generate packages that can be used to test PRs before they get released.
Also adds some vscode files to simplify testing a bit.


TwitchBronBron:
oops, wrong repo!

